PR #5, opened 2017-05-11
Restructure the project a little bit

brendandburns:
This is mainly because `swagger-codegen` overwrites the `pom.xml` in the `kubernetes` directory, so if I want to have  a custom `pom.xml` for hand-written code, it needs to be in a different module.

We should consider renaming the `kubernetes` directory to `generated` which is more descriptive.


mbohlool:
you can add `pom.xml` to `.swagger-codegen-ignore`?

mbohlool:
anyway, I prefer this new structure. What I am trying right now in python is to move some of these files (that added to codegen-ignore list and modified) to a base folder and created a symlink in the original location. this `base` and `util` folders should move to another $lang-base repo according to new client structure document.

PR #7, opened 2017-05-12
Add support for kubeconfig based client creation.

brendandburns:
@mbohlool

### DIFF
--- a/kubernetes/.swagger-codegen-ignore
+++ b/kubernetes/.swagger-codegen-ignore
@@ -1,6 +1,7 @@
 .gitignore
 git_push.sh
-# Remove this once https://github.com/swagger-api/swagger-codegen/pull/5629
-# is merged.
+# Remove this once swagger-codegen 2.2.3 is released and we update. 
+# We want https://github.com/swagger-api/swagger-codegen/pull/5629
+# in the release.
 src/main/java/io/kubernetes/client/ApiClient.java
 
--- a/kubernetes/.swagger-codegen-ignore
+++ b/kubernetes/.swagger-codegen-ignore
@@ -1,3 +1,6 @@
 .gitignore
 git_push.sh
+# Remove this once https://github.com/swagger-api/swagger-codegen/pull/5629
+# is merged.
+src/main/java/io/kubernetes/client/ApiClient.java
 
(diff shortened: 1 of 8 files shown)


kubernetes/.swagger-codegen-ignore
```diff
@@ -1,3 +1,6 @@
 .gitignore
 git_push.sh
+# Remove this once https://github.com/swagger-api/swagger-codegen/pull/5629
```
mbohlool: It needs to be released so we remove this. I think the comment should say "until 2.2.3 is released or something like that".

util/src/main/java/io/kubernetes/client/util/Config.java
```diff
@@ -32,4 +56,110 @@ public static ApiClient fromCluster() throws IOException {
...
+    }
+
+    public static ApiClient fromConfig(InputStream stream) {
+        return fromConfig(new InputStreamReader(stream));
+    }
+
+    public static ApiClient fromConfig(Reader input) {
+        // Note to the reader: I considered creating a Config object
+        // and parsing into that instead of using Maps, but honestly
+        // this seemed cleaner than a bunch of boilerplate classes
```
mbohlool: I think this method getting big because it is trying to do all loading by itself. What about breaking it down like what we did in python client? https://github.com/kubernetes-incubator/client-python/blob/master/kubernetes/base/config/kube_config.py#L108

kubernetes/.swagger-codegen-ignore
brendandburns: done.

util/src/main/java/io/kubernetes/client/util/Config.java
brendandburns: done, please take another look.

brendandburns:
@mbohlool comments addressed.

brendandburns:
@mbohlool friendly ping on this one...

Thanks!
--brendan

mbohlool:
Looks good. I am not sure about util/src/main/java/io/kubernetes/client/util/SSLUtils.java (looks a little complicated to me) but I assume it is code reviewed in the original repo. Please stash commits into one and rebase, and I will merge it.

brendandburns:
@mbohlool rebased.

Yeah, the `SSLUtil` is total cut and paste hackery. I don't really think its worth cleaning up, I'd rather treat it as a black box.

mbohlool:
SGTM

mbohlool:
Look like I merged this before stashing commits together. I prefer commits like "Address comments." to be stashed in other commits, let's do that for other PRs.

PR #15, opened 2017-05-13
Add a bunch of simplified helper methods for creating clients from different information.

### LINKED ISSUE #13: Support "simple" client creation
Methods for:
   * host only
   * host, user, password (basic auth), optional insecure
   * host, bearer token, optional insecure

brendandburns:
Fixes #13

### DIFF
--- a/util/src/main/java/io/kubernetes/client/util/Config.java
+++ b/util/src/main/java/io/kubernetes/client/util/Config.java
@@ -32,4 +32,35 @@ public class Config {
 
         return result;
     }
+
+    public static ApiClient fromUrl(String url) {
+        return fromUrl(url, true);
+    }
+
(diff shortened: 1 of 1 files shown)


util/src/main/java/io/kubernetes/client/util/Config.java
```diff
@@ -32,4 +32,33 @@ public static ApiClient fromCluster() throws IOException {
 
         return result;
     }
+
+    public static ApiClient fromUrl(String url) {
+        return fromHost(url, true);
```
mbohlool: was this suppose to be `fromUrl`?

brendandburns: Done.  (we need to add automatic CI/CD :)

PR #17, opened 2017-05-16
Automagic client creation

### LINKED ISSUE #12: Improve "automagic" client creation.
Order should be:
   * Check for `$KUBECONFIG` env var
   * Check for in cluster configuration (service account)
   * Check for `$HOME/.kube/config`
   * Check for kube proxy (localhost:8001)
   * Default to localhost:8080

brendandburns:
Fixes #12 

Builds on top of #7, that PR should be merged first.

### DIFF
--- a/util/pom.xml
+++ b/util/pom.xml
@@ -24,6 +24,19 @@
 			<artifactId>commons-codec</artifactId>
 			<version>1.10</version>
 		</dependency>
+		<!-- test dependencies -->
+		<dependency>
+			<groupId>junit</groupId>
+			<artifactId>junit</artifactId>
+			<version>4.12</version>
+			<scope>test</scope>
+		</dependency>
+		<!-- https://mvnrepository.com/artifact/com.github.stefanbirkner/system-rules -->
+		<dependency>
+			<groupId>com.github.stefanbirkner</groupId>
+			<artifactId>system-rules</artifactId>
+			<version>1.16.1</version>
+		</dependency>
 	</dependencies>
 	<build>
 		<plugins>
@@ -36,6 +49,22 @@
 					<target>1.7</target>
 				</configuration>
 			</plugin>
+			      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <version>2.12</version>
+        <configuration>
+          <systemProperties>
+            <property>
+              <name>loggerPath</name>
+              <value>conf/log4j.properties</value>
+            </property>
+          </systemProperties>
+          <argLine>-Xms512m -Xmx1500m</argLine>
+          <parallel>methods</parallel>
+          <forkMode>pertest</forkMode>
+        </configuration>
+      </plugin>
 		</plugins>
 	</build>
 </project>
--- a/util/src/main/java/io/kubernetes/client/util/Config.java
+++ b/util/src/main/java/io/kubernetes/client/util/Config.java
@@ -4,6 +4,7 @@ import io.kubernetes.client.ApiClient;
 
 import java.io.BufferedReader;
 import java.io.ByteArrayInputStream;
+import java.io.File;
 import java.io.FileReader;
 import java.io.FileInputStream;
 import java.io.FileNotFoundException;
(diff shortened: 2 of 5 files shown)


mbohlool:
Please rebase.

brendandburns:
@mbohlool rebased.

Thanks!

brendandburns:
@mbohlool friendly ping on this one.

Thanks!

mbohlool:
Look like this needs a rebase.

util/src/main/java/io/kubernetes/client/util/Config.java
```diff
@@ -137,4 +138,36 @@ public static ApiClient fromConfig(Reader input) {
...
+     * @returns The best APIClient given the previously described rules
+     */
+    public static ApiClient defaultClient() throws IOException {
+        String kubeConfig = System.getenv("KUBECONFIG");
+        if (kubeConfig != null) {
+            return fromConfig(new FileReader(kubeConfig));
+        }
+        String homeDir = System.getenv("HOME");
+        if (homeDir != null) {
+            File homeConfig = new File(new File(new File(homeDir), ".kube"), "config");
```
mbohlool: Can we extract constant for all of these strings? ".kube", "config" "HOME", "KUBECONFIG". I suggest we use something like "~/.kube.config" in constant and expand ~.

util/src/main/java/io/kubernetes/client/util/Config.java
```diff
@@ -137,4 +138,36 @@ public static ApiClient fromConfig(Reader input) {
...
+            if (homeConfig.exists()) {
+                return fromConfig(new FileReader(homeConfig));
+            }
+        }
+        File clusterCA = new File(SERVICEACCOUNT_CA_PATH);
+        if (clusterCA.exists()) {
+            return fromCluster();
+        }
+        ApiClient client = new ApiClient();
+        client.setBasePath("http://localhost:8080");
```
mbohlool: why this is not https? also better to have constant for this string too.

util/src/test/java/io/kuberentes/client/util/ConfigTest.java
```diff
@@ -0,0 +1,32 @@
...
+    @Rule
+    public final EnvironmentVariables environmentVariables
+        = new EnvironmentVariables();
+        
+    @Test
+    public void testDefaultClientNothingPresent() {
+        environmentVariables.set("HOME", "/non-existent");
+        try {
+            ApiClient client = Config.defaultClient();
+            assertEquals(client.getBasePath(), "http://localhost:8080");
```
mbohlool: can we test other cases too?

util/pom.xml
```diff
@@ -24,6 +24,19 @@
...
 		</dependency>
+		<!-- test dependencies -->
+		<dependency>
+			<groupId>junit</groupId>
+			<artifactId>junit</artifactId>
+			<version>4.12</version>
+			<scope>test</scope>
+		</dependency>
+		<!-- https://mvnrepository.com/artifact/com.github.stefanbirkner/system-rules -->
+		<dependency>
```
mbohlool: I am not sure if this dependency is used!?

brendandburns: The dependency provides the `EnvironmentVariables` class used here:

https://github.com/kubernetes-client/java/pull/17/files#diff-f4d44c285e36e2ee7e3edfe79246b2f7R22

brendandburns:
Comment addressed, please take another look.

mbohlool:
I still see unrelated commits in this PR.

brendandburns:
Rebased and squashed.  Thanks!

util/src/main/java/io/kubernetes/client/util/KubeConfig.java
```diff
@@ -61,22 +68,35 @@ public KubeConfig(ArrayList<Object> contexts,
...
-        currentContext = (Map<String, Object>) findObject(contexts, context).get("context");
+        Map<String, Object> ctx = findObject(contexts, context);
+        if (ctx == null) {
+            System.err.println("Failed to find context: " + context + " in " + contexts);
+            return false;
+        }
+        currentContext = (Map<String, Object>) ctx.get("context");
         if (currentContext == null) {
-            return;
+            System.err.println("Failed to find 'context' in " + ctx);
```
mbohlool: ditto

util/src/main/java/io/kubernetes/client/util/KubeConfig.java
```diff
@@ -61,22 +68,35 @@ public KubeConfig(ArrayList<Object> contexts,
...
     }
 
-    public void setContext(String context) {
+    public boolean setContext(String context) {
         currentCluster = null;
         currentUser = null;
-        currentContext = (Map<String, Object>) findObject(contexts, context).get("context");
+        Map<String, Object> ctx = findObject(contexts, context);
+        if (ctx == null) {
+            System.err.println("Failed to find context: " + context + " in " + contexts);
```
mbohlool: Should this be an exception?

brendandburns: I debated this... I decided that I preferered "return false" rather than "throw" because the context not existing isn't "exceptional" per-se but rather a potentially expected outcome...

(and I should probably get rid of the stderr logging)

If you feel strongly about throwing instead of returning false, I'm ok with that, this is a gray area.

(there is some interesting discussion on the subject here: https://codereview.stackexchange.com/questions/11724/is-it-better-practice-to-have-void-method-throw-an-exception-or-to-have-the-meth)

util/src/main/java/io/kubernetes/client/util/KubeConfig.java
brendandburns: see above.

brendandburns:
I debated the question you raised, ultimately I decided to `return false` instead of throw, but if you feel strongly I'm happy to switch it, let me know.

Thanks!

util/src/main/java/io/kubernetes/client/util/KubeConfig.java
mbohlool: The stderr thing distracted me. I agree that returning false is better here. Can you get rid of the stderr logging?

mbohlool:
As my only comment is the stderr things, to unblock other PRs, I am going to mark this as LGTM and merge it. You can remove those stderr logging in one of the other PRs if you see that fit.

PR #32, opened 2017-06-24
Support insecure ca root validation.

brendandburns:
@mbohlool this is needed for Bronze certification

Fixes https://github.com/kubernetes-client/java/issues/33

### DIFF
--- a/util/src/main/java/io/kubernetes/client/util/Config.java
+++ b/util/src/main/java/io/kubernetes/client/util/Config.java
@@ -114,16 +114,22 @@ public class Config {
             ex.printStackTrace();
         }
 
-        // It's silly to have to do it in this order, but each SSL setup
-        // consumes the CA cert, so if we do this before the client certs
-        // are injected the cert input stream is exhausted and things get
-        // grumpy'
-        String caCert = config.getCertificateAuthorityData();
-        String caCertFile = config.getCertificateAuthorityFile();
-        try {
-            client.setSslCaCert(SSLUtils.getInputStreamFromDataOrFile(caCert, caCertFile));
-        } catch (FileNotFoundException e) {
-            e.printStackTrace();
+        if (config.verifySSL()) {
+            // It's silly to have to do it in this order, but each SSL setup
+            // consumes the CA cert, so if we do this before the client certs
+            // are injected the cert input stream is exhausted and things get
+            // grumpy'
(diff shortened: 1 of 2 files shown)


util/src/main/java/io/kubernetes/client/util/Config.java
```diff
@@ -114,16 +114,22 @@ public static ApiClient fromConfig(Reader input) {
...
-        } catch (FileNotFoundException e) {
-            e.printStackTrace();
+        if (config.verifySSL()) {
+            // It's silly to have to do it in this order, but each SSL setup
+            // consumes the CA cert, so if we do this before the client certs
+            // are injected the cert input stream is exhausted and things get
+            // grumpy'
+            String caCert = config.getCertificateAuthorityData();
+            String caCertFile = config.getCertificateAuthorityFile();
+            if (caCert != null || caCertFile != null) {
```
mbohlool: should we fail if config.verifySSL() is true and both of these two are null?

util/src/main/java/io/kubernetes/client/util/Config.java
```diff
@@ -114,16 +114,22 @@ public static ApiClient fromConfig(Reader input) {
...
+                try {
+                    client.setSslCaCert(SSLUtils.getInputStreamFromDataOrFile(caCert, caCertFile));
+                } catch (FileNotFoundException e) {
+                    e.printStackTrace();
+                }
+            }
+        } else {
+            client.setVerifyingSsl(false);
```
mbohlool: should we also set this to true in the previous block. just to be bullet proof (e.g. somebody loading config with verifySSL false and then again with it set to true.

util/src/main/java/io/kubernetes/client/util/Config.java
brendandburns: No, because you could be using a legit well-known CA issued cert, in which case you don't have to supply your own CA root.

util/src/main/java/io/kubernetes/client/util/Config.java
brendandburns: This call is static and always creates both a new KubeConfig object and returns a new client, so I don't think that is a risk.

brendandburns:
Comments addressed, relevant issue filed and linked.

mbohlool:
LGTM

Issue #35, opened 2017-06-25
Support authenticator plugins


brendandburns:
Useful for rewriting tokens and the like.

brendandburns:
Closing this as it was fixed by #37

PR #37, opened 2017-06-26
Add an authenticator interface and implementation.

brendandburns:
Partial fix for #35


util/src/main/java/io/kubernetes/client/util/KubeConfig.java
```diff
@@ -155,14 +165,12 @@ public String getAccessToken() {
...
-                Date expiry = (Date) authConfig.get("expiry");
-                if (expiry != null) {
-                    if (expiry.compareTo(new Date()) <= 0) {
-                        // TODO: Generate new token here...
-                        throw new IllegalStateException("Token is expired!");
-                    }
+                String name = (String) authConfig.get("name");
+                Authenticator auth = authenticators.get(name);
+                if (auth != null) {
+                    // TODO: Persiter here.
```
lwander: s/Persiter/Persister

util/src/main/java/io/kubernetes/client/util/authenticators/GCPAuthenticator.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+    @Override
+    public String getName() {
+        return "gcp";
+    }
+
+    @Override
+    public String getToken(Map<String, Object> config, ConfigPersister persister) {
+        Date expiry = (Date) config.get("expiry");
+        if (expiry != null && expiry.compareTo(new Date()) <= 0) {
```
lwander: Don't all tokens expire? Can this be moved in to an abstract super class with an abstract `refreshToken` method?

util/src/main/java/io/kubernetes/client/util/authenticators/ConfigPersister.java
```diff
@@ -0,0 +1,26 @@
...
+import java.util.Map;
+
+public interface ConfigPersister {
+    /**
+     * Save a new version of the configuration
+     * @param name The name of the authenticator configuration (e.g. 'azure')
+     * @param config The new configuration to persist
+     * @throws IOException when an error occurs while persisting
+     */
+    void persistConfig(String name, Map<String, Object> config) throws IOException;
```
lwander: Why is the authenticator name required? It seems strange that you pass the `ConfigPersister` to `getToken` in the `Authenticator` which will always have to supply its own name to `persistConfig`.

util/src/main/java/io/kubernetes/client/util/KubeConfig.java
brendandburns: done.

util/src/main/java/io/kubernetes/client/util/authenticators/GCPAuthenticator.java
brendandburns: Refactored to do this.

util/src/main/java/io/kubernetes/client/util/authenticators/ConfigPersister.java
brendandburns: I was emulating what client-go does, but I got rid of this (and I like it better now, so thanks for the feedback!)

brendandburns:
@lwander rebased, let me know if you have any last comments, otherwise, I'll merge later today.

Thanks!

lwander:
Looks good!

PR #39, opened 2017-06-28
Update java client to spec 1.6.7 and added TPR support

### LINKED ISSUE #38: Support Third Party Resources
We've added TPR support in gen repo. This should be as simple as generating client again, though in the process, I've found a bug in java client generator and fixed it in gen repo.

mbohlool:
Fixes #38


brendandburns:
I'm going to blind LGTM this assuming that this is all auto-generated...

PR #40, opened 2017-06-28
Add Watch Support

### LINKED ISSUE #8: Support Watch operations

mbohlool:
This is my first try to implement watch. It is mostly implemented like the `Watch` in python client. We can think of changing templates (for both python and java) to create a `watch` call for each `list` operation in API classes.

Fixes #8

depends on https://github.com/kubernetes-client/gen/pull/16

built on top of #39

### DIFF
--- a/examples/src/main/java/io/kubernetes/client/examples/WatchExample.java
+++ b/examples/src/main/java/io/kubernetes/client/examples/WatchExample.java
@@ -1,0 +1,44 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package io.kubernetes.client.examples;
+
+import com.google.gson.reflect.TypeToken;
+import io.kubernetes.client.ApiClient;
+import io.kubernetes.client.Watch;
+import io.kubernetes.client.ApiException;
+import io.kubernetes.client.Configuration;
+import io.kubernetes.client.apis.CoreV1Api;
+import io.kubernetes.client.models.V1Namespace;
+import io.kubernetes.client.util.Config;
+
+import java.io.IOException;
+
+/**
+ * A simple example of how to use Watch API to watch changes in Namespace list.
+ */
+public class WatchExample {
+    public static void main(String[] args) throws IOException, ApiException{
+        ApiClient client = Config.defaultClient();
+        Configuration.setDefaultApiClient(client);
+
+        CoreV1Api api = new CoreV1Api();
+
+        Watch<V1Namespace> watch = client.watch(
+                api.listNamespaceCall(null, null, null, null, 5, Boolean.TRUE, null, null),
+                new TypeToken<Watch.Response<V1Namespace>>(){}.getType());
+
+        for (Watch.Response<V1Namespace> item : watch) {
+            System.out.printf("%s : %s%n", item.type, item.object.getMetadata().getName());
+        }
+    }
+}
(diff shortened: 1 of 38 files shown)


examples/src/main/java/io/kubernetes/client/examples/WatchExample.java
```diff
@@ -0,0 +1,44 @@
...
+
+        Watch<V1Namespace> watch = client.watch(
+                api.listNamespaceCall(null, null, null, null, 5, Boolean.TRUE, null, null),
+                new TypeToken<Watch.Response<V1Namespace>>(){}.getType());
+
+        for (Watch.Response<V1Namespace> item : watch) {
+            System.out.printf("%s : %s%n", item.type, item.object.getMetadata().getName());
+        }
+    }
+}
```
brendandburns: nit: newline at end of file?

(I actually don't care but I feel like you asked for this in a different PR and we should probably be consistent...)

kubernetes/src/main/java/io/kubernetes/client/ApiClient.java
```diff
@@ -949,6 +949,26 @@ public File prepareDownloadFile(Response response) throws IOException {
...
+                    try {
+                        respBody = response.body().string();
+                    } catch (IOException e) {
+                        throw new ApiException(response.message(), e, response.code(), response.headers().toMultimap());
+                    }
+                }
+                throw new ApiException(response.message(), response.code(), response.headers().toMultimap(), respBody);
+            }
+            return new Watch<>(this.json, response.body(), watchType);
+        } catch (IOException e) {
```
brendandburns: Why not have this throw IOException also? Seems better than catching and wrapping...

kubernetes/src/main/java/io/kubernetes/client/Watch.java
```diff
@@ -0,0 +1,93 @@
...
+
+/**
+ * Watch class implements watch mechansim of kubernetes. For every list API
+ * call with a watch parameter you should be able to pass its call to this class
+ * and watch changes to that list. For example CoreV1Api.listNamespace has watch
+ * parameter, so you can create a call using CoreV1Api.listNamespaceCall and
+ * set watch to True and watch the changes to namespaces.
+ */
+public class Watch<T> implements Iterable<Watch.Response<T>> {
+
```
brendandburns: Javadoc?

kubernetes/src/main/java/io/kubernetes/client/Watch.java
```diff
@@ -0,0 +1,93 @@
...
+    public Iterator<Response<T>> iterator() {
+        return new Iterator<Response<T>>() {
+            Response<T> nextItem = null;
+            boolean loadNext = true;
+            IOException lastException = null;
+            public boolean hasNext() {
+                if (loadNext) {
+                    try {
+                        nextItem = Watch.this.next();
```
brendandburns: Hrm, I don't think that this obeys the Java iterator contract.

There is no requirement that a user call `hasNext()` to advance the iterator afaik.  Why not just have all of this logic in the `next()` method?

mbohlool: I am calling `hasNext()` in `next()` method so there is no need for user to call this before `next()`. I faced two options here, to raise exception on `next()` or `hasNext()`. I chose this way so when I use for-each with this iterator and the watch closes because of timeout, it won't raise an exception (because for-each calls hasNext() before next()). Of course if user cares about what happen they can always call next() and get the reason (or just call next() on Watch class itself).

kubernetes/src/main/java/io/kubernetes/client/Watch.java
mbohlool: Done.

kubernetes/src/main/java/io/kubernetes/client/ApiClient.java
mbohlool: I just wanted to be consistent with the rest of the ApiClient code. Execute method, for example, do the same (as well as other methods).

examples/src/main/java/io/kubernetes/client/examples/WatchExample.java
mbohlool: Yep, I think we should be consistent. Done.

mbohlool:
Comments addressed or replied to. Please take another look.

kubernetes/src/main/java/io/kubernetes/client/Watch.java
mbohlool: I've looked at this at another angle and made is much much simpler now. Please look at the Watch class as a whole again. It is implements both Iterable and Iterator and converts all IOExceptions to RuntimeException to be able to implement Iterator.

brendandburns:
LGTM.

PR #43, opened 2017-06-29
Promote client capabilities to Silver and add relevant installation steps

mbohlool:
Look like our support level is Silver (we support watch). I've fixed that in this PR and added relevant installation instruction from auto generated README file. We need to update this procedures once we released the client.


brendandburns:
LGTM.

PR #44, opened 2017-06-30
Update POM to reflect License & Authors

lwander:
Let me know if these changes are accurate.

### DIFF
--- a/kubernetes/pom.xml
+++ b/kubernetes/pom.xml
@@ -19,18 +19,18 @@
 
   <licenses>
     <license>
-      <name>Unlicense</name>
-      <url>http://unlicense.org</url>
+      <name>The Apache Software License, Version 2.0</name>
+      <url>http://www.apache.org/licenses/LICENSE-2.0.txt</url>
       <distribution>repo</distribution>
     </license>
   </licenses>
 
   <developers>
     <developer>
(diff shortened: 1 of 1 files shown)


kubernetes/pom.xml
```diff
@@ -19,18 +19,18 @@
...
   </licenses>
 
   <developers>
     <developer>
-      <name>Swagger</name>
-      <email>apiteam@swagger.io</email>
-      <organization>Swagger</organization>
-      <organizationUrl>http://swagger.io</organizationUrl>
+      <name>The Kubernetes Authors</name>
+      <email>kubernetes-dev@googlegroups.com</email>
```
mbohlool: maybe kubernetes-sig-api-machinery@? @brendandburns What do you think?

brendandburns: Sure? I'm not sure it much matters, though I think eventually a sig-developers or sig-clients would be more appropriate than apimachinery (which seems dedicated to internals, not externals)

lwander: Is there an email address under the kubernetes.io domain that we could use perhaps?

brendandburns:
I'm going to merge for now, @mbohlool we should perhaps bring up the topic of "contact emails" at the next community meeting...

PR #47, opened 2017-07-10
Disable doclint when building with java8

lwander:
The generated javadoc comments in the `models` packages contain several
`<` characters which the java8 doclint plugin rejects. Note - this means if you
build with java8 you will _not_ have javadocs. Another solution would be
to fix the stems for the comments in that package, but I'm not sure how
to go about that.

### DIFF
--- a/kubernetes/pom.xml
+++ b/kubernetes/pom.xml
@@ -166,6 +166,15 @@
         </plugins>
       </build>
     </profile>
+    <profile>
+      <id>disable-java8-doclint</id>
+      <activation>
+        <jdk>[1.8,)</jdk>
+      </activation>
+      <properties>
+        <additionalparam>-Xdoclint:none</additionalparam>
+      </properties>
+    </profile>
   </profiles>
 
   <dependencies>


lwander:
@mbohlool @brendanburns

mbohlool:
Does this worth fixing templates in [swagger-codegen](https://github.com/swagger-api/swagger-codegen/tree/master/modules/swagger-codegen/src/main/resources/Java)?

lwander:
The comments look quite k8s specific, and it looks like swagger correctly escapes html characters in other places for the javadocs. I'll look into where the bad characters come from.

brendandburns:
LGTM, fixing where they come from makes sense, but let's get this building now, I had someone run into this issue...

lwander:
Here is the proper fix: https://github.com/swagger-api/swagger-codegen/pull/6040

mbohlool:
Can you keep an open issue for this to port it back and revert this PR when your proper fix merged?

lwander:
Sure thing

PR #53, opened 2017-07-17
Release parent project (to release util module)

lwander:
I realized we need to build & release the `util` submodule as well. I moved the release-specific bits of the `kubernetes/pom.xml` to the top-level `pom.xml`, and added a flag to the `examples` module to avoid releasing that.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2,13 +2,94 @@
 	xmlns="http://maven.apache.org/POM/4.0.0" 
 	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
 	<modelVersion>4.0.0</modelVersion>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -2,13 +2,94 @@
...
+      <distribution>repo</distribution>
+    </license>
+  </licenses>
+
+  <developers>
+    <developer>
+      <name>The Kubernetes Authors</name>
+      <email>kubernetes-dev@googlegroups.com</email>
+      <organization>Kubernetes</organization>
+      <organizationUrl>http://kubernetes.io</organizationUrl>
```
brendandburns: https?

brendandburns:
One small comment. Feel free to self-merge once it is addressed.

pom.xml
lwander: Thanks - good catch

lwander:
I seemed to have commit privileges before, can you merge on my behalf?

brendandburns:
Done.

PR #56, opened 2017-07-24
Fix header handling for resource create requests

lwander:
A specific content type needs to be specified for create requests; the kubernetes swagger.json lists `consumes [ "*/*" ] `  for most resource create requests, which can't be set as a content-type header directly.

Since all requests are serialized into JSON this seems appropriate.

@brendandburns @mbohlool PTAL

### DIFF
--- a/kubernetes/src/main/java/io/kubernetes/client/ApiClient.java
+++ b/kubernetes/src/main/java/io/kubernetes/client/ApiClient.java
@@ -768,10 +768,10 @@ public class ApiClient {
      *
      * @param contentTypes The Content-Type array to select from
      * @return The Content-Type header to use. If the given array is empty,
-     *   JSON will be used.
+     *    or matches any, JSON will be used.
      */
     public String selectHeaderContentType(String[] contentTypes) {
-        if (contentTypes.length == 0) {
+        if (contentTypes.length == 0 || contentTypes[0].equals("*/*")) {
             return "application/json";
         }
         for (String contentType : contentTypes) {
--- a/examples/src/main/java/io/kubernetes/client/examples/CreatePodExample.java
+++ b/examples/src/main/java/io/kubernetes/client/examples/CreatePodExample.java
@@ -1,0 +1,62 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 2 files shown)


kubernetes/src/main/java/io/kubernetes/client/ApiClient.java
```diff
@@ -768,10 +768,10 @@ public String selectHeaderAccept(String[] accepts) {
      *
      * @param contentTypes The Content-Type array to select from
      * @return The Content-Type header to use. If the given array is empty,
-     *   JSON will be used.
+     *    or matches any, JSON will be used.
      */
     public String selectHeaderContentType(String[] contentTypes) {
-        if (contentTypes.length == 0) {
```
brendandburns: I don't think we can do this, we'll need to make this change in the upstream swagger-codegen and I don't think we can land it there.

The right thing to do is to fix the Kubernetes swagger to not have this.

mbohlool: > ... and I don't think we can land it there.
We did this for python client. https://github.com/swagger-api/swagger-codegen/pull/4152
I think it is a reasonable thing to do because swagger clients expect json and when server accept *, it includes json.
> The right thing to do is to fix the Kubernetes swagger to not have this.
I also agree with this regardless of swagger-codegen fix, we need to fix this in our spec.

lwander: My read of the Kubernetes swagger was that it was declaring that it, as a server, accepts anything (`*/*`). If this is true, it's up to the client to decide what it wants to send, and requires either swagger-codegen to be smarter about assigning content types or us as consumers of the generated code to be explicit in what we are sending. But of course, if Kubernetes only accepts json & yaml for example then the right fix is in the Kubernetes swagger.

An alternative to modifying ApiClient in place is to subclass it and override this method to avoid having our changes overwritten by running codegen. Although - when I tested this against the autoupdate script it left my changes in place, so I'm not sure what's going on there.

brendandburns: Ok, let's due the PR in upstream swagger-codegen.  They've been great about quick turnaround in the past.  Once that is merged, I'm happy to merge this PR.

brendandburns: (and let me know if you need help with the PR, I've done a bunch of them myself)

lwander: Sounds good. 

And thanks - I've done one myself now too so this shouldn't be too bad.

lwander:
Closing for: https://github.com/kubernetes-client/java/issues/58

Issue #50, opened 2017-07-12
Reenable doclint for java8 builds


lwander:
Java8 builds of the client disabled doclint due to a bug in the swagger code gen that didn't HTML-escape certain parameter descriptions.

The fix will be to revert https://github.com/kubernetes-client/java/pull/47 once https://github.com/swagger-api/swagger-codegen/pull/6040 is merged.

lwander:
https://github.com/kubernetes-client/java/pull/61

Issue #58, opened 2017-07-25
Patch swagger-codegen to respect */* content-type header


lwander:

lwander:
https://github.com/swagger-api/swagger-codegen/pull/6175

PR #61, opened 2017-08-03
Update generated code from latest swagger-codegen

lwander:


lwander:
Closes https://github.com/kubernetes-client/java/issues/50 and https://github.com/kubernetes-client/java/issues/58

kubernetes/.swagger-codegen/VERSION
```diff
@@ -0,0 +1 @@
+2.3.0-SNAPSHOT
```
lwander: I am not sure how acceptable it is to be pointing at -SNAPSHOT here

mbohlool: Is this an auto-generated file? If it is not, I suggest you add hash instead of SNAPSHOT. we can add generating this file to the generator along with containerizing the generation.

mbohlool:
please separate generated and manual changes (if any) in different commits

kubernetes/.swagger-codegen/VERSION
lwander: It's auto-generated, based on whatever the version [here](https://github.com/kubernetes-client/gen/blob/master/openapi/java.xml#L15) is pointing at. I locally updated that to point at the latest snapshot build. Trouble is, 2.3.0 isn't released yet, and updating that file to point at 2.3.0-SNAPSHOT requires you to checkout & build swagger-codegen locally to run the code generator. As far as I know, you can't point at a specific commit hash anyhow.

lwander:
Will do

kubernetes/.swagger-codegen/VERSION
mbohlool: When we create containers for `gen` repo, they would checkout either a released version or specific commit of swagger-codegen and build using that. There, we can modify this file to record the commit it used to generate the code. It is a design I had in mind but if you have a better suggestion, let me know.

lwander: Specific commit or release version sounds best - how were you going about pointing at the specific commit? If that can be set externally to the pom.xml, pointing at -SNAPSHOT will use the codegen at that commit.

mbohlool: I wonder if we can set commit hash string instead of `-SNAPSHOT` in `pom.xml` file (e.g. 2.3.0-f9d44a838b26f537ea65fd36b5eeb6d07fb5d39f). It is allowed in semver. The script that suppose to build the container can use that specific commit or a released version otherwise.

lwander: To supply a version in the pom.xml that depends on the git commit we'd have to change how swagger-codegen is built. The version specified here is resolved by maven, not our build script, unfortunately. If our automation did this, it might get messy and interfere with users local development environments since to pull a -SNAPSHOT release your machines local maven repo is used.

It might make sense (for now in documentation only) require the latest @ HEAD version of swagger-codegen to be checked out & built locally each time the client generation is run (until 2.3.0 is released). This should technically agree with semver, given that the maintainers of swagger-codegen are diligent in segmenting changes to each release branch by feature/bug-fix/breaking-change/etc...

lwander:
Done

kubernetes/.swagger-codegen/VERSION
mbohlool: The solution was about a build container. It should not interfere with local development environment. Basically that is my suggestion for build-gen-contianer.sh for a specific language:

1. Read pom file of that language and extract swagger-codegen version from it.
2. If it is not a final release (has - in it), parse it and extract commit hash from it. (if it said SNAPSHOT then head would be the commit hash).
3. Build a container by checking out that specific commit (or a release version or head).
4. Copy and then modify language pom.xml into the container to replace the commit hash (if exists) with "-SNAPSHOT"
5. Setup rsync's and generate the client using this container.

brendandburns:
This LGTM, I'll wait for @mbohlool to give it one final look.

kubernetes/.swagger-codegen-ignore
```diff
@@ -1,8 +1,8 @@
 .gitignore
 git_push.sh
-# Remove this once swagger-codegen 2.2.3 is released and we update. 
+# Remove this once swagger-codegen 2.3.0 is released and we update.
```
mbohlool: Do we use the HEAD of swagger-codegen for this PR? In that case, why we still have this line?

kubernetes/.swagger-codegen/VERSION
mbohlool: (btw this is not blocker for this PR, will move the discussion to https://github.com/kubernetes-client/gen/issues/5)

lwander: SGTM

kubernetes/.swagger-codegen-ignore
lwander: HEAD of swagger-codegen is technically still behind 2.3.0 since it hasn't been released yet. I'm realizing the line below I left filenames commented-out that shouldn't be overwritten until that PR is merged.

kubernetes/src/main/java/io/kubernetes/client/ApiClient.java
```diff
@@ -90,6 +90,7 @@ public ApiClient() {
         // Setup authentications (key: authentication name, value: authentication).
         authentications = new HashMap<String, Authentication>();
         authentications.put("BearerToken", new ApiKeyAuth("header", "authorization"));
+        authentications.put("BasicAuth", new HttpBasicAuth());
```
lwander: BTW because `authentications` is private, and there is no way to modify it we (for now) need to make this edit in place. I'll make an edit in swagger codegen that allows us to specify custom authentications.

kubernetes/.swagger-codegen-ignore
mbohlool: What I meant was the comment is not useful anymore if we are using the latest swagger-codegen code to generate this. e.g. we don't need to wait for 2.3.0 to get this feature and we should not ignore any file here (as our change is already in HEAD).

I see all those files are commented out here anyway, so we don't ignore them in generation? Even ApiClient that we are manually editing bellow?

kubernetes/src/main/java/io/kubernetes/client/ApiClient.java
mbohlool: Follow up on my previous comment, if we are manually editing ApiClient.java, we should add it to the ignore list and remove it when the upstream PR is merged.

lwander: It's in the ignore list https://github.com/kubernetes-client/java/pull/61/files/13b2813174966eb5ca6ee1cff700531e5492273f#diff-ea99bc9c9c08dd8a42e296305e0e9c48R7

lwander: Oh sorry I had locally staged changes that weren't pushed, h/o

lwander: nvm they are pushed, you're commenting on an old commit.

kubernetes/.swagger-codegen-ignore
lwander: The change is not merged yet, which is why those files are in that ignore list. Once 2.3.0 is out (or the PR is merged) it will be safe to regenerate the files.

mbohlool:
LGTM.

PR #63, opened 2017-08-11
Release 0.1

lwander:
Sorry for the churn, there were a few things left to fix

Non [maven-release-plugin] commits were written by me to get the release to pass.

Sonatypes validation of our artifacts has passed, I just need to update the JIRA ticket to indicate we've done our first release and the artifacts should be available after that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,10 +25,16 @@ Add this dependency to your project's POM:
 ```xml
 <dependency>
     <groupId>io.kubernetes</groupId>
-    <artifactId>client-java</artifactId>
-    <version>1.0-SNAPSHOT</version>
+    <artifactId>client-java-util</artifactId>
+    <version>0.1</version>
     <scope>compile</scope>
 </dependency>
+```
+
+### Gradle users
+
+```groovy
+compile 'io.kubernetes:client-java-util:0.1'
 ```
 
 ### Others
@@ -43,7 +49,7 @@ mvn package
 
 Then manually install the following JARs:
 
-* target/client-java-1.0-SNAPSHOT.jar
+* target/client-java-0.2-SNAPSHOT.jar
 * target/lib/*.jar
 
 ## Example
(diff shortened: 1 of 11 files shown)


lwander:
Do we need to update the client capabilities badge?

mbohlool:
So there are commits that cancel out changes from the previous commit? I see a lot of change 1.0-SNAPSHOT to 0.1-SNAPSHOT in one commit, then 0.1-SNAPSHOT to 0.1 in another commit then 0.1 to 0.2-SNAPSHOT in another one. Why is that? We probably need to create a release branch if we need those intermediate commits for the release process.

lwander:
The release branch has already been created: https://github.com/kubernetes-client/java/tree/release-0.1 (and is pointing at the next snapshot release).

The fact that our snapshot dependencies were sitting at 1.0 was incorrect. For the maven-release-plugin to publish the artifacts at version 0.1 they need to be checked it at version 0.1-SNAPSHOT - the plugin then automatically checked in a commit at version 0.1, released that, and revved the version to 0.2-SNAPSHOT.

Further development against master is under the assumption that we will either release 0.2 or 1.0 next. Patch changes are made against the branch I linked to above.

mbohlool:
I see, it didn't realize those two PRs were created by the maven-release-plugin. This looks good to me.

lwander:
NP - thanks! The 0.1 release is complete :)

mbohlool:
Nice. Thanks. Can you announce it on kubernetes dev mailing list?

lwander:
My email should have gone out about an hour ago - kubernetes-dev @ googlegroups?

mbohlool:
I saw it. The artifact ID is client-java-util? That is strange. I expected
client-java.

lwander:
Let's fix that with the next release

lwander:
https://github.com/kubernetes-client/java/issues/64

PR #67, opened 2017-08-31
Initial add of proto-buffer files

brendandburns:
@mbohlool @lwander 

Uses https://github.com/kubernetes-client/gen/pull/23


lwander:
Some of these seem to follow /src/main/java/k8s/io, and others /src/main/java/io/kubernetes (the rest of the repo follow the later). Should we make them consistent?

brendandburns:
@lwander fixed (and fixed the generation script too...)

Please take another look.

lwander:
Thanks! I don't think I can review much beyond that... so LGTM

What's the plan for integrating this with the client library?

brendandburns:
@lwander good question! I need this library to start experimenting. If you'd prefer, I can wait to merge until I have the complete story, I'll branch off of this PR and keep exploring.

lwander:
I always like to merge sooner rather than later - feel free to merge whenever you feel ready!

PR #71, opened 2017-09-11
Add create, update and delete protocol buffer methods.

brendandburns:
Finishes off basic protocol buffer support.  It's not perfect, but it's usable.

@mbohlool @lwander

### DIFF
--- a/examples/src/main/java/io/kubernetes/client/examples/ProtoExample.java
+++ b/examples/src/main/java/io/kubernetes/client/examples/ProtoExample.java
@@ -1,0 +1,47 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package io.kubernetes.client.examples;
+
+import io.kubernetes.client.ApiClient;
+import io.kubernetes.client.ApiException;
+import io.kubernetes.client.Configuration;
+import io.kubernetes.client.ProtoClient;
+import io.kubernetes.client.proto.V1.Pod;
+import io.kubernetes.client.proto.V1.PodList;
+import io.kubernetes.client.util.Config;
+
+import java.io.IOException;
+
+
+/**
+ * A simple example of how to use the Java API
+ * 
+ * Easiest way to run this:
+ *   mvn exec:java -Dexec.mainClass="io.kubernetes.client.examples.ProtoExample"
+ * 
+ * From inside $REPO_DIR/examples
+ */
+public class ProtoExample {
+    public static void main(String[] args) throws IOException, ApiException, InterruptedException {
+        ApiClient client = Config.defaultClient();
+        Configuration.setDefaultApiClient(client);
+
+        ProtoClient pc = new ProtoClient(client);
+        PodList list = pc.list(PodList.newBuilder(), "/api/v1/namespaces/default/pods");
+
+        if (list.getItemsCount() > 0) {
+            Pod p = list.getItems(0);
+            System.out.println(p.toString());
+        }
+    }
+}
--- a/examples/src/main/java/io/kubernetes/client/examples/ProtoExample.java
+++ b/examples/src/main/java/io/kubernetes/client/examples/ProtoExample.java
@@ -16,6 +16,11 @@ import io.kubernetes.client.ApiClient;
 import io.kubernetes.client.ApiException;
 import io.kubernetes.client.Configuration;
 import io.kubernetes.client.ProtoClient;
+import io.kubernetes.client.ProtoClient.ObjectOrStatus;
+import io.kubernetes.client.proto.Meta.ObjectMeta;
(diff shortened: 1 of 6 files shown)


util/src/main/java/io/kubernetes/client/ProtoClient.java
```diff
@@ -69,43 +103,98 @@ public void setApiClient(ApiClient apiClient) {
...
+        return request(builder, path, "DELETE", null, null, null);
+    }
+
+    /**
      * Generic protocol buffer based HTTP request.
      * Not intended for general consumption, but public for advance use cases.
      * @param builder The appropriate Builder for the object receveived from the request.
      * @param method The HTTP method (e.g. GET) for this request.
      * @param path The URL path to call (e.g. /api/v1/namespaces/default/pods/pod-name)
      * @return A Message of type T
```
mbohlool: add documentation for new fields body, apiVersion, kind?

util/src/main/java/io/kubernetes/client/ProtoClient.java
```diff
@@ -69,43 +103,98 @@ public void setApiClient(ApiClient apiClient) {
...
 
     /**
+     * Create a Kubernetes API object using protocol buffer encoding. Performs a POST
+     * @param obj The object to create
+     * @param path The URL path to call
+     * @param apiVersion The api version to use
+     * @param kind The kind of the object
+     * @return The response received.
+     */
+    public <T extends Message> ObjectOrStatus<T> create(T obj, String path, String apiVersion, String kind)
```
mbohlool: Until we find a generic way to do it, I suggest we extract apiVersion and kind from T using reflection. I know it is not ideal, but I prefer it vs adding these two parameters to the surface of our proto APIs. ditto for other calls.

mbohlool:
First round of review done. Thanks.

util/src/main/java/io/kubernetes/client/ProtoClient.java
```diff
@@ -69,43 +103,98 @@ public void setApiClient(ApiClient apiClient) {
...
+    public <T extends Message> ObjectOrStatus<T> update(T obj, String path, String apiVersion, String kind)
+            throws ApiException, IOException {
+        return request(obj.newBuilderForType(), path, "PUT", obj, apiVersion, kind);
+    }
+
+    /**
+     * Delete a kubernetes API object using protocol buffer encoding.
+     * @param builder The builder for the response
+     * @param path The path to call in the API server
+     * @return The response received
```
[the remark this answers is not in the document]
brendandburns: done. (changed the return type)

util/src/main/java/io/kubernetes/client/ProtoClient.java
brendandburns: done.

util/src/main/java/io/kubernetes/client/ProtoClient.java
brendandburns: Sadly, the fields aren't present in the proto.  See the long discussion here:

https://groups.google.com/forum/#!topic/kubernetes-sig-api-machinery/deP_LDaAdVs

Otherwise I would have used reflection.

brendandburns:
@mbohlool comments (mostly) addressed, ptal.

Thanks!
--brendan

mbohlool:
@brendandburns I went through that thread about protos and replied there. I think grpc proxy would be nice addition to our generated clients and it could solve the problem of all clients with one generated proxy.

brendandburns:
Cool, I'll read your design doc.

In the meantime, can we merge this PR?

Thanks
--brendan

mbohlool:
Sure.

util/src/main/java/io/kubernetes/client/ProtoClient.java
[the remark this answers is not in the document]
mbohlool: Should the document for return value be the status response? ditto for other methods.

util/src/main/java/io/kubernetes/client/ProtoClient.java
```diff
@@ -69,43 +103,101 @@ public void setApiClient(ApiClient apiClient) {
...
+
+    /**
+     * Update a Kubernetes API object using protocol buffer encoding. Performs a PUT
+     * @param obj The object to create
+     * @param path The URL path to call
+     * @param apiVersion The api version to use
+     * @param kind The kind of the object
+     * @return The response received.
+     */
+    public <T extends Message> ObjectOrStatus<T> update(T obj, String path, String apiVersion, String kind)
```
mbohlool: is it possible for non-delete calls to return status?

brendandburns: Yes, update in particular sometimes returns a 201 (I observed this behavior)

brendandburns:
Tried to address the comments, please take another look.

examples/src/main/java/io/kubernetes/client/examples/ProtoExample.java
```diff
@@ -16,6 +16,11 @@
 import io.kubernetes.client.ApiException;
```
shivamsinghrajput325, line 16: import io.kubernetes.client.ApiException;

<u><strong><font color=”blue” size=”8px”>H<strong><font color=”red” size=”8px”>E<strong><font color=”gold” size=”8px”>L<strong><font color=”blue” size=”8px”>L<strong><font color=”green” size=”8px”>O <strong><font color=”blue” size=”8px”>G <strong><font color=”red” size=”8px”>O <strong><font color=”gold” size=”8px”>O <strong><font color=”blue” size=”8px”>G <strong><font color=”green” size=”8px”>L <strong><font color=”red” size=”8px”>E

<test>

PR #76, opened 2017-09-22
Generate a jar with dependencies and build a Dockerfile for it.

brendandburns:
@lwander @mbohlool 

Useful for testing in clusters.

### DIFF
--- a/examples/Dockerfile
+++ b/examples/Dockerfile
@@ -1,0 +1,7 @@
+FROM openjdk:8-alpine
+
+COPY target/client-java-examples-0.2-SNAPSHOT-jar-with-dependencies.jar /examples.jar
+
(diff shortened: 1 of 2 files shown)


mbohlool:
lgtm

rvowles:
Problems:

- Your "jar with dependencies" also contains all of the test dependencies, so junit & hamcrest is there
- Everything is flattened into one namespace, so you are getting unpredictable behaviour when some files may be the same (e.g. manifests, configuration files, service loaders, etc).

It would have been better to use spring boot loader (which has nothing to do with spring boot thankfully). It is just as easy to configure and avoids all of these problems:

https://docs.spring.io/spring-boot/docs/current/maven-plugin/usage.html

PR #81, opened 2017-09-29
Allow kubeconfig to be loaded without current-context

lwander:
Before this change, any attempt to load a kubeconfig w/o current-context field would NPE. Since `kubectl` doesn't have this behavior, and we expose the ability to set a custom context on a kubeconfig, we shouldn't crash here.


brendandburns:
LGTM

PR #84, opened 2017-09-30
Update client to kubernetes 1.8

mbohlool:
Add update client script and update client to kubernetes 1.8.


lwander:
Looks like the URL path stays the same for CoreV1Pods. When a resource is promoted from a beta track to v1, does the API spec keep the beta resource around?

mbohlool:
I will look into this. Probably I need to change the pre-processing script to keep the names backward compatible. That would benefit all clients. Will update this PR and ping you when it is ready. Thanks.

mbohlool:
@lwander I fixed the client generator in https://github.com/kubernetes-client/gen/pull/24 and updated the client here using the new code to see if it fixed the problem. Please take another look.

lwander:
Looks good - shame the extra parameters are required for all API calls

lwander:
I'm ready to merge when you are

mbohlool:
The other PR is merged. Please merge this one. Thanks.

PR #87, opened 2017-10-08
Add an initial logs utility class.

brendandburns:
Adds capability to get logs as a stream, not just as a string.

### DIFF
--- a/util/src/main/java/io/kubernetes/client/Logs.java
+++ b/util/src/main/java/io/kubernetes/client/Logs.java
@@ -1,0 +1,94 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package io.kubernetes.client;
+
+import io.kubernetes.client.Configuration;
+import io.kubernetes.client.apis.CoreV1Api;
+import io.kubernetes.client.models.V1Pod;
+import io.kubernetes.client.util.WebSockets;
+import io.kubernetes.client.util.WebSocketStreamHandler;
+
+import java.io.ByteArrayInputStream;
+import java.io.InputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.io.PipedInputStream;
+import java.io.PipedOutputStream;
+import java.io.Reader;
+
+import org.apache.commons.lang.StringUtils;
+
+import com.squareup.okhttp.Call;
+import com.squareup.okhttp.Response;
+
+/**
+ * Utility class offering streaming access to Pod logs.
+ */
+public class Logs {
+    private ApiClient apiClient;
+    private CoreV1Api coreClient;
+
+    /**
+     * Simple Logs API constructor, uses default configuration
+     */
+    public Logs() {
+        this(Configuration.getDefaultApiClient());
+    }
+
+    /**
+     * Logs API Constructor
+     * @param apiClient The api client to use.
+     */
+    public Logs(ApiClient apiClient) {
+        this.apiClient = apiClient;
+        this.coreClient = new CoreV1Api(apiClient);
+    }
+
+    /**
+     * Get the API client for these Logs operations.
+     * @return The API client that will be used.
+     */
+    public ApiClient getApiClient() {
+        return apiClient;
+    }
+
+    /**
+     * Set the API client for subsequent Logs operations.
+     * @param apiClient The new API client to use.
+     */
+    public void setApiClient(ApiClient apiClient) {
+        this.apiClient = apiClient;
+    }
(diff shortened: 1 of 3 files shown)


util/src/main/java/io/kubernetes/client/Logs.java
```diff
@@ -0,0 +1,94 @@
...
+
+    // Important note. You must close this stream or else you can leak connections.
+    public InputStream streamNamespacedPodLog(String namespace, String name, String container,
+                                              Integer sinceSeconds, Integer tailLines, boolean timestamps) throws ApiException, IOException {
+        Call call = coreClient.readNamespacedPodLogCall(name, namespace, container, true, null, "false", false, sinceSeconds, tailLines, timestamps, null, null);
+        Response response = call.execute();
+        if (!response.isSuccessful()) {
+            throw new ApiException("Logs request failed: " + response.code());
+        }
+        return response.body().byteStream();
```
lwander: Do we always know what this content-type will be? (for decoding)

util/src/main/java/io/kubernetes/client/Logs.java
```diff
@@ -0,0 +1,94 @@
...
+
+    // Important note. You must close this stream or else you can leak connections.
+    public InputStream streamNamespacedPodLog(String namespace, String name, String container) throws ApiException, IOException {
+        return streamNamespacedPodLog(namespace, name, container, null, null, false);
+    }
+
+
+    // Important note. You must close this stream or else you can leak connections.
+    public InputStream streamNamespacedPodLog(String namespace, String name, String container,
+                                              Integer sinceSeconds, Integer tailLines, boolean timestamps) throws ApiException, IOException {
```
lwander: Is `sinceSeconds` since epoch, or when the container started?

brendandburns: Added JavaDoc (stolen from Swagger docs)

util/src/main/java/io/kubernetes/client/Logs.java
brendandburns: Added JavaDoc (it's `text/plain`)

util/src/main/java/io/kubernetes/client/Logs.java
```diff
@@ -0,0 +1,117 @@
...
+import org.apache.commons.lang.StringUtils;
+
+import com.squareup.okhttp.Call;
+import com.squareup.okhttp.Response;
+
+/**
+ * Utility class offering streaming access to Pod logs.
+ */
+public class Logs {
+    private ApiClient apiClient;
```
mbohlool: suggestion: I would make apiClient `public final` and remove setter and getter for it.

util/src/main/java/io/kubernetes/client/Logs.java
```diff
@@ -0,0 +1,117 @@
...
+        this.apiClient = apiClient;
+    }
+
+    /**
+     * Stream logs from a running container.
+     * Important note. You must close this stream or else you can leak connections.
+     * @param pod The Pod to get logs for, uses the first container in the pod for logs
+     * @return A UTF-8 encoded InputStream
+     */
+    public InputStream streamNamespacedPodLog(V1Pod pod) throws ApiException, IOException {
```
mbohlool: should we just call this `streamPodLog`?

util/src/main/java/io/kubernetes/client/Logs.java
```diff
@@ -0,0 +1,117 @@
...
+
+import org.apache.commons.lang.StringUtils;
+
+import com.squareup.okhttp.Call;
+import com.squareup.okhttp.Response;
+
+/**
+ * Utility class offering streaming access to Pod logs.
+ */
+public class Logs {
```
mbohlool: Would this be confused with a Logger? When I saw the topic for this PR, I though this is a special logger. I suggest we call it something like `PodLogs`?

util/src/main/java/io/kubernetes/client/Logs.java
lwander: Why `public`?

mbohlool: That means you don't need a getter for it. It is a common practice for these type of variables but now that I think more about it, maybe having getter is not a bad idea. But I still suggest we don't have a setter for this and make it `private final`.

lwander: Yeah was just curious since the rest of the generated code uses getters in favor of `public final`. Agreed that a setter probably shouldn't exist.

util/src/main/java/io/kubernetes/client/Logs.java
lwander: I think this is done for consistency, all the generated api methods seem to follow `<verb>Namespaced<resource>` for this kind of stuff

util/src/main/java/io/kubernetes/client/Logs.java
brendandburns: Renamed.

util/src/main/java/io/kubernetes/client/Logs.java
brendandburns: Yeah, this was for consistency...

util/src/main/java/io/kubernetes/client/Logs.java
brendandburns: Removed the setter.

brendandburns:
Comments addressed, please re-check.

Thanks!
--brendan

PR #89, opened 2017-10-08
Add a URL prefix if one is missing.

### LINKED ISSUE #88: Config loading doesn't handle servers w/o http(s)
When loading a config that looks like:

```yaml
apiVersion: v1
clusters:
- cluster:
    server: "192.168.1.135:8001"
  name: "nuc"
```

Connection requests fail, because there is no protocol for the URL.

However `kubectl` works properly. We should add a prefix if it is missing.

brendandburns:
Fixes #88

### DIFF
--- a/util/src/main/java/io/kubernetes/client/util/Config.java
+++ b/util/src/main/java/io/kubernetes/client/util/Config.java
@@ -104,7 +104,15 @@ public class Config {
 
     public static ApiClient fromConfig(KubeConfig config) {
         ApiClient client = new ApiClient();
-        client.setBasePath(config.getServer());
+        String server = config.getServer();
+        if (!server.startsWith("http://") && !server.startsWith("https://")) {
+            if (server.indexOf(":443") != -1) {
+                server = "https://" + server;
+            } else {
+                server = "http://" + server;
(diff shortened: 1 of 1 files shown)


util/src/main/java/io/kubernetes/client/util/Config.java
```diff
@@ -104,7 +104,15 @@ public static ApiClient fromConfig(Reader input) {
 
     public static ApiClient fromConfig(KubeConfig config) {
         ApiClient client = new ApiClient();
-        client.setBasePath(config.getServer());
+        String server = config.getServer();
+        if (!server.startsWith("http://") && !server.startsWith("https://")) {
+            if (server.indexOf(":443") != -1) {
```
lwander: Should we rely on `config.verifySSL()` instead/as well?

brendandburns: verifySSL is default true, so I don't think that's the right thing to test here...

lwander: Ah good point.

mbohlool:
LGTM

PR #91, opened 2017-10-09
Add support for PortForward.

### LINKED ISSUE #11: Support port-forward to Pod

brendandburns:
Fixes #11 

Uses WebSocket for Port Forward as added by https://github.com/kubernetes/kubernetes/pull/33684

@mbohlool @lwander

### DIFF
--- a/util/src/main/java/io/kubernetes/client/PortForward.java
+++ b/util/src/main/java/io/kubernetes/client/PortForward.java
@@ -1,0 +1,193 @@
+package io.kubernetes.client;
+
+import io.kubernetes.client.Configuration;
+import io.kubernetes.client.models.V1Pod;
+import io.kubernetes.client.util.WebSockets;
+import io.kubernetes.client.util.WebSocketStreamHandler;
+
+import java.io.InputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+
+/**
+ * Utility class for setting up port-forwarding connections.
+ * Uses the WebSockets API, not the SPDY API (which the Go client uses)
+ * 
+ * The protocol is undocumented as far as I can tell, but the PR that added
+ * it is here:
+ * https://github.com/kubernetes/kubernetes/pull/33684
+ * 
+ * And the protocol is:
+ * 
+ * ws://server/api/v1/namespaces/<namespace>/pods/<pod>/portforward?ports=80&ports=8080
+ * 
+ * I/O for first port (80) is on Channel 0
+ * Err for first port (80) is on Channel 1
+ * I/O for second port (8080) is on Channel 2
+ * Err for second port (8080) is on Channel 3
+ * <and so on for remaining ports>
+ * 
+ * The first two bytes of each output stream is the port that is being forwarded
+ * in little-endian format.
+ */
+public class PortForward {
+    private ApiClient apiClient;
+
+    /**
+     * Simple PortForward API constructor, uses default configuration
+     */
+    public PortForward() {
+        this(Configuration.getDefaultApiClient());
+    }
+
+    /**
+     * PortForward API Constructor
+     * @param apiClient The api client to use.
+     */
+    public PortForward(ApiClient apiClient) {
+        this.apiClient = apiClient;
+    }
+
+    /**
+     * Get the API client for these PortForward operations.
+     * @return The API client that will be used.
+     */
+    public ApiClient getApiClient() {
+        return apiClient;
+    }
+
+    /**
+     * Set the API client for subsequent PortForward operations.
+     * @param apiClient The new API client to use.
+     */
+    public void setApiClient(ApiClient apiClient) {
+        this.apiClient = apiClient;
+    }
+
+    private String makePath(String namespace, String name) {
+          return "/api/v1/namespaces/" +
+            namespace +
+            "/pods/" +
+            name +
+            "/portforward";
+    }
+
+    /**
+     * PortForward to a container
+     * 
+     * @param pod The pod where the port forward is run.
+     * @param ports The ports to forward
+     * @return The result of the Port Forward request.
+     */
+    public PortForwardResult forward(V1Pod pod, List<Integer> ports) throws ApiException, IOException {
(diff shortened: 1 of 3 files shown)


util/src/main/java/io/kubernetes/client/util/WebSockets.java
```diff
@@ -81,13 +83,28 @@
...
         headers.put(STREAM_PROTOCOL_HEADER, allProtocols);
         headers.put(HttpHeaders.CONNECTION, HttpHeaders.UPGRADE);
         headers.put(HttpHeaders.UPGRADE, SPDY_3_1);
 
-        Request request = client.buildRequest(path, method, new ArrayList<Pair>(), new ArrayList<Pair>(), null, headers, new HashMap<String, Object>(), new String[0], null);
+        Request request = client.buildRequest(path, method, queryParams, new ArrayList<Pair>(), null, headers, new HashMap<String, Object>(), new String[0], null);
+        streamRequest(request, client, listener);
+    }
+
+    /** If we ever upgrade to okhttp 3...
```
lwander: I think this might get interpreted as a javadoc with that extra `*`

util/src/main/java/io/kubernetes/client/PortForward.java
```diff
@@ -0,0 +1,180 @@
...
+        }
+
+        /**
+         * Get the output stream for the specified port number (e.g. 80)
+         * @param port The port number to get the stream for.
+         */
+        public OutputStream getOutboundStream(int port) {
+            int portIndex = findPortIndex(port);
+            if (portIndex == -1) {
+                throw new IllegalArgumentException("No such port!");
```
lwander: Would it be neater to have `findPortIndex(..)` throw this exception instead?

util/src/main/java/io/kubernetes/client/PortForward.java
```diff
@@ -0,0 +1,180 @@
...
+     * @param namespace The namespace of the Pod
+     * @param name The name of the Pod
+     * @param ports The ports to forward
+     */
+    public PortForwardResult forward(String namespace, String name, List<Integer> ports) throws ApiException, IOException {
+        String path = makePath(namespace, name);
+        WebSocketStreamHandler handler = new WebSocketStreamHandler();
+        PortForwardResult result = new PortForwardResult(handler, ports);
+        List<Pair> queryParams = new ArrayList<>();
+        queryParams.add(new Pair("ports", "80"));
```
lwander: Is it always '80' or just once for the example you gave above?

util/src/main/java/io/kubernetes/client/PortForward.java
```diff
@@ -0,0 +1,180 @@
...
+ * https://github.com/kubernetes/kubernetes/pull/33684
+ * 
+ * And the protocol is:
+ * 
+ * ws://server/api/v1/namespaces/<namespace>/pods/<pod>/portforward?ports=80&ports=8080
+ * 
+ * I/O for first port (80) is on Channel 0
+ * Err for first port (80) is on Channel 1
+ * I/O for second port (8080) is on Channel 2
+ * Err for second port (8080) is on Channel 2
```
lwander: Channel 2 or 3?

util/src/main/java/io/kubernetes/client/PortForward.java
```diff
@@ -0,0 +1,180 @@
...
+    public PortForwardResult forward(V1Pod pod, List<Integer> ports) throws ApiException, IOException {
+        return forward(pod.getMetadata().getNamespace(), pod.getMetadata().getNamespace(), ports);
+    }
+
+    /**
+     * PortForward to a container.
+     * 
+     * @param namespace The namespace of the Pod
+     * @param name The name of the Pod
+     * @param ports The ports to forward
```
lwander: Please add `@return`, I don't think our linter is catching it but the release will fail w/o it.

util/src/main/java/io/kubernetes/client/PortForward.java
```diff
@@ -0,0 +1,180 @@
...
+            "/pods/" +
+            name +
+            "/portforward";
+    }
+
+    /**
+     * PortForward to a container
+     * 
+     * @param pod The pod where the port forward is run.
+     * @param ports The ports to forward
```
lwander: `@return`

util/src/main/java/io/kubernetes/client/PortForward.java
```diff
@@ -0,0 +1,180 @@
...
+     * PortForward API Constructor
+     * @param apiClient The api client to use.
+     */
+    public PortForward(ApiClient apiClient) {
+        this.apiClient = apiClient;
+    }
+
+    /**
+     * Get the API client for these PortForward operations.
+     * @returns The API client that will be used.
```
lwander: s/returns/return

util/src/main/java/io/kubernetes/client/util/WebSockets.java
brendandburns: done.

util/src/main/java/io/kubernetes/client/PortForward.java
brendandburns: Good catch, fixed.

util/src/main/java/io/kubernetes/client/PortForward.java
brendandburns: fixed.

util/src/main/java/io/kubernetes/client/PortForward.java
brendandburns: fixed.

util/src/main/java/io/kubernetes/client/PortForward.java
brendandburns: fixed.

util/src/main/java/io/kubernetes/client/PortForward.java
brendandburns: ugh! good catch, fixed.

util/src/main/java/io/kubernetes/client/PortForward.java
brendandburns: switched to return null instead of throw.

brendandburns:
@lwander comments addressed, please take another look.

Thanks!

PR #95, opened 2017-10-14
add close to watch

### LINKED ISSUE #94: Leak from OkHttpClient
on performing repeated watches on custom objects I eventually see in logs

```2017-10-14 15:26:12.773  WARN 5 --- [ ConnectionPool] com.squareup.okhttp.OkHttpClient         : A connection to https://10.0.0.1/ was leaked. Did you forget to close a response body?```

I don't see any way to close the response body from the Watch wrappers or gain access to it to close manually.

I'm doing:
```java
CustomObjectsApi api = new CustomObjectsApi();
		Watch<Object> watch = Watch.createWatch(
                client,
                api.listNamespacedCustomObjectCall("xxx.yyy", "v1alpha1", "default", "mykind", null, null, rs, true, null, null),
                new TypeToken<Watch.Response<Object>>(){}.getType());
```

cliveseldon:
fixes #94

### DIFF
--- a/util/src/main/java/io/kubernetes/client/util/Watch.java
+++ b/util/src/main/java/io/kubernetes/client/util/Watch.java
@@ -120,4 +120,8 @@ public class Watch<T> implements Iterable<Watch.Response<T>>,
     public void remove() {
         throw new UnsupportedOperationException("remove");
     }
+    
+    public void close() throws IOException {
+    	this.response.close();
+    }
 }
--- a/util/src/main/java/io/kubernetes/client/util/Watch.java
+++ b/util/src/main/java/io/kubernetes/client/util/Watch.java
@@ -31,7 +31,8 @@ import java.util.Iterator;
  * set watch to True and watch the changes to namespaces.
  */
 public class Watch<T> implements Iterable<Watch.Response<T>>,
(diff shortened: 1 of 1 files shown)


util/src/main/java/io/kubernetes/client/util/Watch.java
```diff
@@ -120,4 +120,8 @@ public boolean hasNext() {
     public void remove() {
         throw new UnsupportedOperationException("remove");
     }
+    
+    public void close() throws IOException {
```
lwander: I'm wondering if this is meant to go in `remove()` - @brendandburns?

brendandburns: No, `remove()` is part of implementing the Java `Iterable` interface, (and it's legal for it to throw if you don't support it...)

We should mark `remove()` with `@Override` to make this clear.

brendandburns:
Generally LGTM.  You need to sign the CLA.

One comment: can you make the class implement `java.lang.Closeable` since we're adding that method.

Thanks!
--brendan

cliveseldon:
I signed a CLA

brendandburns:
LGTM, thanks!

PR #104, opened 2017-10-26
Use setApiKey over setUsername/Password/AccessToken as they are not used

lewisheadden:
`ApiClient#setUsername`, `ApiClient#setPassword`, and `ApiClient#setAccessToken` all correspond to different underlying authentication mechanisms for Swagger. In order for these to be applied the mechanism must be enabled in two places: it must be registered in the client and it must be specified for the API request in Swagger JSON. 

Currently, this client only has `BearerToken` and `HttpBasicAuth` registered. As such all calls to `ApiClient#setAccessToken` will fail complaining that `OAuth` is not registered. 

Further, although `HttpBasicAuth` is registered, it is not specified for any API requests in the Swagger JSON. As such `setUsername`/`setPassword` calls will work but will never be passed when making API requests.

To get around this, this PR uses `setApiKey` for all authentication types and add an appropriate prefix. The value supplied to `setApiKey` is simply sent as the `Authorization` header and as such is usable for OAuth, Basic Auth, and Bearer Token.

### DIFF
--- a/util/src/main/java/io/kubernetes/client/util/Config.java
+++ b/util/src/main/java/io/kubernetes/client/util/Config.java
@@ -13,6 +13,7 @@ limitations under the License.
 package io.kubernetes.client.util;
 
 import io.kubernetes.client.ApiClient;
+import okio.ByteString;
 import org.apache.log4j.Logger;
 
 import javax.net.ssl.KeyManager;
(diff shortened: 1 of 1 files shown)


lewisheadden:
Working on getting approval for the CLA from my employer.

brendandburns:
This LGTM, once CLA is present.

lewisheadden:
CLA signed, @brendandburns

brendandburns:
LGTM, thanks @lewisheadden

PR #106, opened 2017-10-28
Fix up some potential leaks, by calling close()

brendandburns:
Apropos of https://github.com/kubernetes-client/java/pull/95

I audited the rest of our code...

### DIFF
--- a/examples/src/main/java/io/kubernetes/client/examples/AttachExample.java
+++ b/examples/src/main/java/io/kubernetes/client/examples/AttachExample.java
@@ -42,7 +42,7 @@ public class AttachExample {
         Configuration.setDefaultApiClient(client);
 
         Attach attach = new Attach();
-        final Attach.AttachResult result = attach.attach("default", "nginx-2371676037-czqx3", true);
+        final Attach.AttachResult result = attach.attach("default", "nginx-4217019353-k5sn9", true);
 
         new Thread(new Runnable() {
             public void run() {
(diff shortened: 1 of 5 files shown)


examples/src/main/java/io/kubernetes/client/examples/AttachExample.java
```diff
@@ -42,7 +42,7 @@ public static void main(String[] args) throws IOException, ApiException, Interru
         Configuration.setDefaultApiClient(client);
 
         Attach attach = new Attach();
-        final Attach.AttachResult result = attach.attach("default", "nginx-2371676037-czqx3", true);
+        final Attach.AttachResult result = attach.attach("default", "nginx-4217019353-k5sn9", true);
```
lwander: How useful is this pod name? Looks like it was generated by a deployment

brendandburns: Well, this is example code :)

I could put `"nginx-example-pod"`

I suppose I could dynamically query and try to attach to the first pod I find, but I'm not sure how perfect an example needs to be.  (I could also take the pod as a command line argument I suppose, but it seems like something for a different PR)

lwander: Fair enough :) maybe we can revisit if someone files an issue but you're right, let's not complicate it

brendandburns:
@lwander any other comments? @mbohlool any thoughts?

If not, I can haz lgtmz?

mbohlool:
Will look and merge today.

PR #108, opened 2017-11-01
Add IntOrString model and GSON adapter

lewisheadden:
Add an `IntOrString` model along with its type adapter. This type allows us to receive a type from the API and then re-serialize it without changing its type. Previously all values received from the API were converted to strings. This meant that if returned to the API, the API would respond with a 400 Bad Request because the type was not always as expected (e.g. `maxUnavailable` should either be an `int` or a `String` that ends with a `%` - if previously `1` had been specified it would be rejected as it was not `1%`).

There will be a corresponding PR in [the gen repo](https://github.com/kubernetes-client/gen) for this.


brendandburns:
LGTM.

Issue #117, opened 2017-11-14
ExtensionsV1beta1Api.createNamespacedIngress is throwing error for servicePort


jeesmon:
java-client: v1.0.0-beta1
k8s: v1.8.2

In V1beta1IngressBackend model, servicePort is String. But calling ExtensionsV1beta1Api.createNamespacedIngress with "servicePort": "443" throws 

```
{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Ingress.extensions \"ingress\" is invalid: [spec.backend.servicePort: Invalid value: \"443\": must contain at least one letter or number (a-z, 0-9), spec.rules[0].http.backend.servicePort: Invalid value: \"443\": must contain at least one letter or number (a-z, 0-9)]","reason":"Invalid","details":{"name":"ingress","group":"extensions","kind":"Ingress","causes":[{"reason":"FieldValueInvalid","message":"Invalid value: \"443\": must contain at least one letter or number (a-z, 0-9)","field":"spec.backend.servicePort"},{"reason":"FieldValueInvalid","message":"Invalid value: \"443\": must contain at least one letter or number (a-z, 0-9)","field":"spec.rules[0].http.backend.servicePort"}]},"code":422}
```

Serialized version of V1beta1Ingress captured from debug

```
{
    "apiVersion": "extensions/v1beta1",
    "kind": "Ingress",
    "metadata": {
        "annotations": {
            "ingress.kubernetes.io/ssl-passthrough": "true"
        },
        "name": "ingress",
        "namespace": "test"
    },
    "spec": {
        "backend": {
            "serviceName": "svc-1",
            "servicePort": "443"
        },
        "rules": [
            {
                "host": "myhost.com",
                "http": {
                    "paths": [
                        {
                            "backend": {
                                "serviceName": "svc-1",
                                "servicePort": "443"
                            },
                            "path": "/path"
                        }
                    ]
                }
            }
        ],
        "tls": [
            {
                "hosts": [
                    "myhost.com"
                ],
                "secretName": "ingress-secret"
            }
        ]
    }
}
```

```
kubectl apply -f ingress.json 
The Ingress "ingress" is invalid: 
* spec.backend.servicePort: Invalid value: "443": must contain at least one letter or number (a-z, 0-9)
* spec.rules[0].http.backend.servicePort: Invalid value: "443": must contain at least one letter or number (a-z, 0-9)
```

[...]

PR #118, opened 2017-11-15
Update kubernetes generated code.

### LINKED ISSUE #115: NetworkPolicyPort fails at numerical ports
Any attempt at saving a V1NetworkPolicyPort fails with:
```Invalid value: \"80\": must contain at least one letter or number (a-z, 0-9)```

It looks like this value should be an IntOrString type rather than a string so that the rendered json would be of int type when required and string type otherwise.

brendandburns:
Regenerate code to pick up the `IntOrString` improvements.

Fixes #115 #117 

@mbohlool ptal

@lwander I think this warrants rev.ing a new package.


lwander:
SGTM, 1.0.0-beta2?

brendandburns:
Works for me...

--brendan

kadwanev:
when can a beta2 be pushed to maven?

brendandburns:
I'm working on getting my permissions right to push to Maven...



________________________________________

PR #129, opened 2017-11-22
Add Quantity type

lewisheadden:
Adds an implementation of [Kubernetes Quantity handling](https://github.com/kubernetes/apimachinery/blob/master/pkg/api/resource/quantity.go) in the Java client.

This allows users to interact with `Quantity`s as their raw value without having to parse units (e.g. `1024` instead of `1Ki`, `1000` instead of `1e3`, `0.001` instead of `1m`). This is helpful for example when you submit a value like `1024Mi` and the Kubernetes API will return `1Gi`. You cannot reliably manipulate the string without parsing it.

This `Quantity` implementation gives you an object containing its `BigDecimal` value and the format of the suffix (e.g. `DECIMAL_SI`, `BINARY_SI`, or `DECIMAL_EXPONENT`) when parsed or the desired format to be written. Using this normalized form you can then manipulate the `BigDecimal` representation and use the `Quantity` implementation to re-serialize it in the appropriate notation.

A PR will also be needed to `gen` to make the Swagger pre-processor not turn `Quantity` into a `String` but map it to this class.

Quantities are complex, to say the least, in Kubernetes but thanks to some guidance from @lavalamp and a lot of time in the Go Debugger, I think this behaves pretty accurately.


kubernetes/src/main/java/io/kubernetes/client/custom/BaseExponent.java
```diff
@@ -0,0 +1,56 @@
+package io.kubernetes.client.custom;
+
+public class BaseExponent {
```
brendandburns: JavaDoc?

kubernetes/src/main/java/io/kubernetes/client/custom/BaseExponent.java
```diff
@@ -0,0 +1,56 @@
...
+
+    private final int base;
+
+    private final int exponent;
+
```
brendandburns: Line spacing is weird here. preference for:

```java
private final int base;
private final int exponent;
...

public BaseExponent(...) {
...
```

kubernetes/src/main/java/io/kubernetes/client/custom/BaseExponent.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        BaseExponent that = (BaseExponent) o;
+
+        if (base != that.base) return false;
```
brendandburns: nit: Preference for:

```java
return base == that.base && exponent == that.exponent && format == that.format;
```

kubernetes/src/main/java/io/kubernetes/client/custom/BaseExponent.java
```diff
@@ -0,0 +1,56 @@
...
+        if (exponent != that.exponent) return false;
+        return format == that.format;
+    }
+
+    @Override
+    public int hashCode() {
+        int result = base;
```
brendandburns: This is funky, and I'm not entirely sure that it is correct.  (or at least I don't understand why it is correct...

Why not:
```java
this.toString().hashCode();
```

kubernetes/src/main/java/io/kubernetes/client/custom/Pair.java
```diff
@@ -0,0 +1,20 @@
+package io.kubernetes.client.custom;
+
+public class Pair<L, R> {
```
brendandburns: We already depend on apache-commons, rev the version to 3.0 and use the Pair implementation from there?

kubernetes/src/main/java/io/kubernetes/client/custom/Quantity.java
```diff
@@ -0,0 +1,71 @@
...
+            return base;
+        }
+    }
+
+    public Quantity(final BigDecimal number, final Format format) {
+        this.number = number;
+        this.format = format;
+    }
+
+    public BigDecimal getBigDecimal() {
```
brendandburns: nit: `getNumber` ?

brendandburns:
Thanks for the PR! some small nits, but looks good to me.

kubernetes/src/main/java/io/kubernetes/client/custom/BaseExponent.java
lewisheadden: I must confess that this was entirely auto-generated by IntelliJ, however, I don't see any error in it.

Given our use case, I think we can happily use `this.toString().hashCode()` as the cost is not prohibitive to us and it will be more future proof in the case of adding or removing fields.

kubernetes/src/main/java/io/kubernetes/client/custom/BaseExponent.java
lewisheadden: Ironically, in its own auto-generated code, IntelliJ suggests this quick fix.

kubernetes/src/main/java/io/kubernetes/client/custom/BaseExponent.java
lewisheadden: Happy to add JavaDoc - to clarify, is this an across the board ask or just for `BaseExponent`?

brendandburns:
LGTM, JavaDoc can wait (but yes, for all classes would be great, but I realized we haven't really done a good job of that in general, so don't want to hold this PR on that...)

util/pom.xml
```diff
@@ -1,109 +1,109 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
```
isaacdong: pom.xml is reformatted. hard to locate changes.

Issue #127, opened 2017-11-22
ExtensionsV1beta1Api.patchNamespacedDeployment throws 500 Internal server error because of wrong content type in header


kondapally1989:
reason: content-type in header is being set to **application/json-patch+json** instead of **application/strategic-merge-patch+json**
 this is causing 500 internal server error

 when i test the patch call to  **/apis/extensions/v1beta1/namespaces/default/deployments/kubeexample?pretty=false**  with content type as    **application/json-patch+json** (in postman rest client) , getting following error 

```
{
    "kind": "Status",
    "apiVersion": "v1",
    "metadata": {},
    "status": "Failure",
    "message": "json: cannot unmarshal object into Go value of type jsonpatch.Patch",
    "code": 500
}
```

I believe content type should be **application/strategic-merge-patch+json** (with it as content type deployment gets updated)

in ExtensionsV1beta1Api.patchNamespacedDeploymentCall

localVarContentTypes is being set to **application/json-patch+json** instead of **application/strategic-merge-patch+json** because of ApiClient.selectHeaderContentType(String [] arr)  always select first element of array({"application/json-patch+json", "application/merge-patch+json", "application/strategic-merge-patch+json"}) as it matches with regex.

Rafaesp:
Same error when calling `new NetworkingV1Api().patchNamespacedNetworkPolicy()`

mjschmidt:
Any word on this issue?

kondapally1989:
@mjschmidt 
with strategic merge patches, it throws internal server error(because of SWAGGER/OPENAPI/codegen related issues). but with Json Patches it will work.
working example  of JsonPatch is here  https://github.com/kubernetes-client/java/blob/master/examples/src/main/java/io/kubernetes/client/examples/PatchExample.java

costimuraru:
How can we use a JsonPatch to patch a Node?

PR #133, opened 2017-11-29
fix for header's  Content-Type:"strategic merge patch"  for patch operation's.

kondapally1989:
fix for issue https://github.com/kubernetes-client/java/issues/127
tried to handle it according to doc https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#patch-operations and also based on similar fix in python client (https://github.com/kubernetes-incubator/client-python/issues/187)
but i believe this is not complete fix, as Content-Type: application/merge-patch+json is never set for merge operation of JSONs.
is it possible to explicitly tell whether it is **application/merge-patch+json** just for patch calls?

### DIFF
--- a/kubernetes/src/main/java/io/kubernetes/client/ApiClient.java
+++ b/kubernetes/src/main/java/io/kubernetes/client/ApiClient.java
@@ -920,13 +920,19 @@ public class ApiClient {
      * @throws ApiException If fail to serialize the request body object
      */
     public Request buildRequest(String path, String method, List<Pair> queryParams, List<Pair> collectionQueryParams, Object body, Map<String, String> headerParams, Map<String, Object> formParams, String[] authNames, ProgressRequestBody.ProgressRequestListener progressRequestListener) throws ApiException {
-        updateParamsForAuth(authNames, queryParams, headerParams);
+        String contentType = (String) headerParams.get("Content-Type");
+        // Content-Type: JSON Patch( for json object) and Strategic Merge Patch (for list of json object)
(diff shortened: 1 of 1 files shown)


brendandburns:
@mbohlool I think that maybe this should be fixed in the OpenAPI/Code Generation?

kondapally1989:
closing the  pull request as this has to be fixed else where (OpenAPI/Code Generation)

mbohlool:
I am not sure about a fix in OpenAPI or code generations. This is a special case we handle it similarly in python client (by checking the type of provided object and change the content-type accordingly). If you guys have a suggestion on how to fix this in code generation, I would like to discuss it, otherwise we should check in something similar to this PR.

kondapally1989:
In OpenAPI 2.0, filter/selection based on content type in request header is not present ([link](https://github.com/OAI/OpenAPI-Specification/issues/146) to issue raised in open api git) may be another usecase but  [issue](https://github.com/OAI/OpenAPI-Specification/issues/182) also.
But in OpenAPI 3.0 this feature/support has been added ([link](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#requestBodyObject) to requestbody).

but swagger([3.0.0.0](https://github.com/swagger-api/swagger-codegen#compatibility)) which supports OpenAPI 3.0  is still in development phase. 

one blunt solution to generate individual functions with correct content type is by splitting the paths of rest end point. some thing like /strategicMergePatchDeploymentWithList, /patchDeployment and  /mergePatchDeployment in specification.

PR #134, opened 2017-11-30
Update proto files with newly generated API objects.

brendandburns:
@lwander @cliveseldon


brendandburns:
Friendly post kubecon ping on this one :)

PR #139, opened 2017-12-07
Project to control its own maven version

westse:
- See https://github.com/takari/maven-wrapper

I have found this useful to both control the maven version, as well as to remove the requirement that maven is even pre-installed on the developer's machine (`./mvnw` downloads it for you if necessary).

Does this practice fit well here too?


westse:
Closing while I try to get my github user email correct

PR #141, opened 2017-12-08
Project to control its own maven version.

westse:
- See https://github.com/takari/maven-wrapper

I have found this useful to both control the maven version, as well as to remove the requirement that maven is even pre-installed on the developer's machine (./mvnw downloads it for you if necessary).

Does this practice fit well here too?


brendandburns:
@lwander is our Maven packaging expert... I defer to him.

lwander:
:sweat: not an expert at all... 

@westse is this similar to gradle's `gradlew` wrapper script? It seems useful

westse:
@lwander , yes in fact it was copied from `gradlew` wrapper because they though it was a good idea for more reproducible builds and developer convenience.

lwander:
Sweet, this seems great. Is there any way to avoid checking in maven-wrapper.jar? Should the script be fetching that?

westse:
At this point the jar is expected to be checked in (gradlew does this as well). That said, it looks like they are working on support for not having the Jar checked in: https://github.com/takari/maven-wrapper/pull/70

lwander:
Cool, works for me

PR #142, opened 2017-12-08
Fix javadocs in examples submodule

davidxia:


examples/src/main/java/io/kubernetes/client/examples/Example.java
```diff
@@ -26,10 +26,10 @@
 /**
  * A simple example of how to use the Java API
  * 
- * Easiest way to run this:
+ * <p>Easiest way to run this:
```
lwander: Do these tags need to be terminated?

brendandburns: In theory yes, but all browsers handle unterminated <p> correctly for legacy reasons

brendandburns: In theory yes, but all browsers handle unterminated <p> correctly for legacy reasons

davidxia: The HTML specs say closing tags can be omitted for p tags. https://stackoverflow.com/a/33787857/553994 I got used to doing this in javadocs but can add them here if you’d like.

brendandburns:
@lwander unless you care about the `<p>` tags, this LGTM.  I'll merge in a couple days unless I hear from you.

lwander:
I don't mind the missing <p> tags - happy to merge now

davidxia:
Thanks
-- 
David Xia
davidxia.com

PR #157, opened 2017-12-20
Add better docs to the WebSocket example.

### LINKED ISSUE #151: WebSocketsExample: Missing example path and authentication error
The WebSocketExample accepts the path as an argument. That is ok. But it would be great if you can include some example path which can be used (for examle as comment or in the README file).

I tried it for example with `/api/v1/pods?watch=true` (but also other adresses). But I always get "403 Forbidden" error although with all other example the authentication worked fine:

```
java.net.ProtocolException: Expected HTTP 101 response but was '403 Forbidden'
	at com.squareup.okhttp.ws.WebSocketCall.createWebSocket(WebSocketCall.java:123)
	at com.squareup.okhttp.ws.WebSocketCall.access$000(WebSocketCall.java:40)
	at com.squareup.okhttp.ws.WebSocketCall$1.onResponse(WebSocketCall.java:98)
	at com.squareup.okhttp.Call$AsyncCall.execute(Call.java:177)
	at com.squareup.okhttp.internal.NamedRunnable.run(NamedRunnable.java:33)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```

Does the WebSocket example need some special authentication? I can see this problem with both master and 1.0.0-beta1. I'm runnng it against Minikube v0.24.1 (Kubernetes 1.8.0).

brendandburns:
Fixes #151 

@mbohlool @lwander

### DIFF
--- a/examples/src/main/java/io/kubernetes/client/examples/WebSocketsExample.java
+++ b/examples/src/main/java/io/kubernetes/client/examples/WebSocketsExample.java
@@ -25,6 +25,18 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.Reader;
 
+/**
+ * This is a pretty low level, most people won't need to use WebSockets
+ * directly.
+ *
+ * If you do need to run it, you can run:
+ *   mvn exec:java \
+ *     -Dexec.mainClass=io.kubernetes.client.examples.WebSocketsExample \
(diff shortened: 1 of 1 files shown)


mbohlool:
LGTM

PR #162, opened 2017-12-27
Fix code that didn't follow the stream protocol.

### LINKED ISSUE #158: PortForward is broken for multiple ports...
The port forward code doesn't appear to skip through the channels accurately, I think we get away with this because it "just works" for a single port.

We should expand to multiple ports and see if it still works.

brendandburns:
Fixes #158


brendandburns:
@lwander @mbohlool friendly ping for a quick review :)

Issue #164, opened 2018-01-04
Watch response of ERROR causes ApiException


bhabermaas:
When performing a watch on a list if the server returns ERROR (kind: status) there is a deserialization error because the status object differs from the Watch.Response<T>. This is particularly prevalent using custom resources.  The Watch response handler blindly uses watchType when deserializing which causes the json deserializer to throw an exception when the response object does not match the watch type.

rjeberhard:
We think was is happening here is that the k8s api server is returning a status in the stream of objects returned by the watch as per here: https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#response-status-kind

However, io.kubernetes.client.util.Watch.Response.next() simply deserializes using the target type rather than checking the kind of object and supporting a potential status.

PR #165, opened 2018-01-11
When watch can't parse an object, try to parse it to a Status.

brendandburns:
Fixes https://github.com/kubernetes-client/java/issues/164

@bhabermas
@jberhard
@mbohlool

### DIFF
--- a/util/src/main/java/io/kubernetes/client/util/Watch.java
+++ b/util/src/main/java/io/kubernetes/client/util/Watch.java
@@ -12,13 +12,16 @@ limitations under the License.
 */
 package io.kubernetes.client.util;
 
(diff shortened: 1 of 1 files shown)


mbohlool:
Catching exception works for now but it would be nice if we had a way to deserialize metadata first to see what is the actual object and then deserialize it in the right object. That would be much cleaner than trying to deserialize it and count on exception.

brendandburns:
Yeah. I thought about that. The trouble is that you'd need to deserialize it into a Map and then go splunking and then re-parse into an object, the performance of that seemed worse than catching an exception...

--brendan



________________________________________

Issue #168, opened 2018-01-17
Easy in-memory instantiation of config


brendandburns:
Referenced in #119 and #167 

Quote from #119:

Something that would allow me to do something like this is what I'm after:
https://github.com/spotify/styx/blob/c0654ac66a74692ffbc6974b824f3d763a6e6bc4/styx-scheduler-service/src/main/java/com/spotify/styx/StyxScheduler.java#L693-L699

kondapally1989:
@brendanburns 

 Will the below ConfigBuilder fit the purpose

```java
package io.kubernetes.client.util;

import java.io.FileNotFoundException;
import java.io.FileReader;
import java.io.IOException;
import java.io.InputStream;
import java.io.InputStreamReader;
import java.io.Reader;
import java.nio.charset.Charset;

import javax.net.ssl.KeyManager;

import okio.ByteString;
import io.kubernetes.client.ApiClient;
import io.kubernetes.client.util.Config;
import io.kubernetes.client.util.KubeConfig;
import io.kubernetes.client.util.SSLUtils;

public  class ConfigBuilder {

	private boolean trustCerts = false;
	private boolean clusterMode = false;
	private boolean defaultKubeConfigMode = false;
	private boolean defaultClientMode = false;
	private boolean verifyingSsl = false;
	private String basePath = "http://localhost";
	private String certificateAuthorityFile = null;
	private String certificateAuthorityData = null;
	private String apiKey = null;
	private String userName = null;
	private String password = null;
	private KeyManager[] keyMgrs = null;
	private String accessToken = null;
	private String apiKeyPrefix = null;
	private InputStream sslCaCert = null;
	private KubeConfig kubeConfig = null;


	public ConfigBuilder() {
	}

	public String getUserName() {
		return userName;
	}

	public ConfigBuilder setUserName(String userName) {
		this.userName = userName;
		return this;
	}

	public String getPassword() {
		return password;
	}

	public ConfigBuilder setPassword(String password) {
		this.password = password;
		return this;
	}

	public String getApiKey() {
		return apiKey;
	}

	public ConfigBuilder setApiKey(String apiKey) {
		this.apiKey = apiKey;
		return this;
	}

	public boolean isTrustCerts() {
		return trustCerts;
	}

	public ConfigBuilder setTrustCerts(boolean trustCerts) {
		this.trustCerts = trustCerts;
		return this;
	}

	public String getbasePath() {
		return basePath;
	}

[...]

PR #171, opened 2018-01-22
adding custom ConfigBuilder

kondapally1989:
custom ConfigBuilder. issue:  https://github.com/kubernetes-client/java/issues/168

### DIFF
--- a/util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
+++ b/util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
@@ -33,11 +33,8 @@ import io.kubernetes.client.ApiClient;
 import io.kubernetes.client.util.Config;
 import io.kubernetes.client.util.KubeConfig;
 
-public class ConfigBuilder {
+public class ClientBuilder {
 
-	private boolean clusterMode = false;
-	private boolean defaultKubeConfigMode = false;
-	private boolean defaultClientMode = false;
 	private boolean verifyingSsl = false;
 	private String basePath = null;
 	private File certificateAuthorityFile = null;
@@ -49,6 +46,7 @@ public class ConfigBuilder {
 	private String accessToken = null;
 	private String apiKeyPrefix = null;
 	private KubeConfig kubeConfig = null;
+	private ApiClient client = null;
 
 	private static final Logger log = Logger.getLogger(Config.class);
 
@@ -56,7 +54,7 @@ public class ConfigBuilder {
 		return userName;
 	}
 
-	public ConfigBuilder setUserName(String userName) {
+	public ClientBuilder setUserName(String userName) {
 		this.userName = userName;
 		return this;
 	}
@@ -65,7 +63,7 @@ public class ConfigBuilder {
 		return password;
 	}
 
-	public ConfigBuilder setPassword(String password) {
+	public ClientBuilder setPassword(String password) {
 		this.password = password;
 		return this;
 	}
@@ -74,7 +72,7 @@ public class ConfigBuilder {
 		return apiKey;
 	}
 
-	public ConfigBuilder setApiKey(String apiKey) {
+	public ClientBuilder setApiKey(String apiKey) {
 		this.apiKey = apiKey;
 		return this;
 	}
@@ -83,7 +81,7 @@ public class ConfigBuilder {
 		return basePath;
 	}
 
-	public ConfigBuilder setBasePath(String basePath) {
+	public ClientBuilder setBasePath(String basePath) {
 		this.basePath = basePath;
 		return this;
 	}
@@ -92,7 +90,7 @@ public class ConfigBuilder {
 		return certificateAuthorityFile;
 	}
 
-	public ConfigBuilder setCertificateAuthority(File certificateAuthorityFile) {
+	public ClientBuilder setCertificateAuthority(File certificateAuthorityFile) {
 		this.certificateAuthorityFile = certificateAuthorityFile;
 		this.verifyingSsl = true;
 		return this;
@@ -102,38 +100,41 @@ public class ConfigBuilder {
 		return certificateAuthorityData;
 	}
 
-	public ConfigBuilder setCertificateAuthority(String certificateAuthorityData) {
+	public ClientBuilder setCertificateAuthority(String certificateAuthorityData) {
 		this.certificateAuthorityData = certificateAuthorityData;
 		this.verifyingSsl = true;
 		return this;
 	}
 
-	public ConfigBuilder setClusterMode() {
-		this.clusterMode = true;
+	public ClientBuilder setClusterMode() throws IOException {
+		this.client = Config.fromCluster();
 		return this;
 	}
 
-	public ConfigBuilder setKubeConfig(KubeConfig config) {
+	public ClientBuilder setKubeConfig(KubeConfig config) {
 		this.kubeConfig = config;
+		if( this.kubeConfig !=null) {
+			this.client = Config.fromConfig(this.kubeConfig);
+		}
 		return this;
 	}
 
-	public ConfigBuilder setDefaultKubeConfigMode() {
-		this.defaultKubeConfigMode = true;
+	public ClientBuilder setDefaultKubeConfigMode() throws FileNotFoundException {
+		this.client = Config.fromConfig(KubeConfig.loadDefaultKubeConfig());
 		return this;
 	}
 
-	public ConfigBuilder setKubeConfig(String fileName) throws FileNotFoundException {
-		this.kubeConfig = KubeConfig.loadKubeConfig(new FileReader(fileName));
+	public ClientBuilder setKubeConfig(File kubeFile) throws FileNotFoundException {
+		this.kubeConfig = KubeConfig.loadKubeConfig(new FileReader(kubeFile));
 		return this;
 	}
 
-	public ConfigBuilder setKubeConfig(Reader input) {
+	public ClientBuilder setKubeConfig(Reader input) {
 		this.kubeConfig = KubeConfig.loadKubeConfig(input);
 		return this;
 	}
 
-	public ConfigBuilder setKubeConfig(InputStream stream) {
+	public ClientBuilder setKubeConfig(InputStream stream) {
 		this.kubeConfig = KubeConfig.loadKubeConfig(new InputStreamReader(stream));
 		return this;
 	}
@@ -142,7 +143,7 @@ public class ConfigBuilder {
 		return keyMgrs;
 	}
 
-	public ConfigBuilder setKeyMgrs(KeyManager[] keyMgrs) {
+	public ClientBuilder setKeyMgrs(KeyManager[] keyMgrs) {
 		this.keyMgrs = keyMgrs;
 		return this;
 	}
@@ -151,17 +152,14 @@ public class ConfigBuilder {
 		return verifyingSsl;
 	}
 
-	public ConfigBuilder setVerifyingSsl(boolean verifyingSsl) {
+	public ClientBuilder setVerifyingSsl(boolean verifyingSsl) {
 		this.verifyingSsl = verifyingSsl;
 		return this;
 	}
 
-	public boolean isDefaultClientMode() {
-		return defaultClientMode;
-	}
 
-	public ConfigBuilder setDefaultClientMode() {
-		this.defaultClientMode = true;
+	public ClientBuilder setDefaultClientMode() throws IOException {
+		client = Config.defaultClient();
 		return this;
 	}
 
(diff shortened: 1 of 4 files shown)


util/src/main/java/io/kubernetes/client/util/ConfigBuilder.java
```diff
@@ -0,0 +1,285 @@
...
+	public boolean isTrustCerts() {
+		return trustCerts;
+	}
+
+	public ConfigBuilder setTrustCerts(boolean trustCerts) {
+		this.trustCerts = trustCerts;
+		return this;
+	}
+
+	public String getbasePath() {
```
brendandburns: `setBasePath()`

util/src/main/java/io/kubernetes/client/util/ConfigBuilder.java
```diff
@@ -0,0 +1,285 @@
...
+		return basePath;
+	}
+
+	public ConfigBuilder setbasePath(String basePath) {
```
brendandburns: `setBasePath(...)`

util/src/main/java/io/kubernetes/client/util/ConfigBuilder.java
```diff
@@ -0,0 +1,285 @@
...
+		return apiKeyPrefix;
+	}
+
+	public ConfigBuilder setApiKeyPrefix(String apiKeyPrefix) {
+		this.apiKeyPrefix = apiKeyPrefix;
+		return this;
+	}
+
+	public ApiClient build()  {
+
```
brendandburns: Delete newline

util/src/main/java/io/kubernetes/client/util/ConfigBuilder.java
```diff
@@ -0,0 +1,285 @@
...
+		ApiClient client = new ApiClient();
+
+		if( kubeConfig !=null) {
+			client = Config.fromConfig(kubeConfig);
+			return client;
```
brendandburns: I think this logic is a little odd.

I would expect that I could do:

```
client = new ConfigBuilder().setKubeConfig().setBasePath().build()
```

And it would mostly load the KubeConfig, but then override the basepath with the custom value I set.

Instead, it just returns the existing KubeConfig and ignores any other `setFoo` directives.

I think that this is going to confuse people, and doesn't match the desired use case.

util/src/main/java/io/kubernetes/client/util/ConfigBuilder.java
```diff
@@ -0,0 +1,285 @@
...
+		}
+
+		if(defaultKubeConfigMode == true) {
+			try {
+				client = Config.fromConfig(KubeConfig.loadDefaultKubeConfig());
+			} catch (FileNotFoundException e) {
+				e.printStackTrace();
+			}
+			return client;
```
brendandburns: as above.

util/src/main/java/io/kubernetes/client/util/ConfigBuilder.java
```diff
@@ -0,0 +1,285 @@
...
+		}
+
+		if(clusterMode == true) {
+			try {
+				client = Config.fromCluster();
+			} catch (IOException e) {
+				e.printStackTrace();
+			}
+			return client;
```
brendandburns: as above.

util/src/main/java/io/kubernetes/client/util/ConfigBuilder.java
```diff
@@ -0,0 +1,285 @@
...
+		}
+
+		if(defaultClientMode ==true ) {
+			try {
+				client = Config.defaultClient();
+			} catch (IOException e) {
+				e.printStackTrace();
+			}
+			return client;
+
```
brendandburns: delete newline (here and everywhere it's extraneous)

util/src/main/java/io/kubernetes/client/util/ConfigBuilder.java
```diff
@@ -0,0 +1,285 @@
...
+		}
+
+		if (basePath != null ) {
+			if(basePath.endsWith("/")) {
+				basePath = basePath.substring(0, basePath.length() - 1);
+			}
+			client.setBasePath(basePath)
+			.setVerifyingSsl(verifyingSsl);
```
brendandburns: why are you setting `verifyingSsl` here?

util/src/main/java/io/kubernetes/client/util/ConfigBuilder.java
```diff
@@ -0,0 +1,285 @@
...
+		}
+
+		else {
+			try {
```
brendandburns: This is weird. You are `try{ ... } catch` and then throwing the exception yourself.  Just log to stderr? or throw an exception?

util/src/main/java/io/kubernetes/client/util/ConfigBuilder.java
```diff
@@ -0,0 +1,285 @@
...
+
+		if(sslCaCert != null) {
+			client.setSslCaCert(sslCaCert);
+		}
+		
+		if(verifyingSsl){
+			if((certificateAuthorityData != null) || (certificateAuthorityFile != null)){
+				try {
+					client.setSslCaCert(SSLUtils.getInputStreamFromDataOrFile(certificateAuthorityData, certificateAuthorityFile));
+				} catch (FileNotFoundException e) {
```
brendandburns: I think you should load the cert when someone calls `setCertificateAuthority(...)`

And not throw/catch here.

util/src/main/java/io/kubernetes/client/util/ConfigBuilder.java
```diff
@@ -0,0 +1,285 @@
...
+	public String getApiKey() {
+		return apiKey;
+	}
+
+	public ConfigBuilder setApiKey(String apiKey) {
+		this.apiKey = apiKey;
+		return this;
+	}
+
+	public boolean isTrustCerts() {
```
brendandburns: This doesn't seem to do anything. Delete?

util/src/main/java/io/kubernetes/client/util/ConfigBuilder.java
kondapally1989: by default(if user doesn't set anything ), want to set verifyingSsl to false. 
but i forget to add the actual setter..(will rectify it)
if( verifyingSsl != null ){
client..setVerifyingSsl(verifyingSsl);
}

util/src/main/java/io/kubernetes/client/util/ConfigBuilder.java
```diff
@@ -0,0 +1,272 @@
...
+		this.kubeConfig = config;
+		return this;
+	}
+
+	public ConfigBuilder setDefaultKubeConfigMode() {
+		this.defaultKubeConfigMode = true;
+		return this;
+	}
+
+	public ConfigBuilder setKubeConfig(String fileName) throws FileNotFoundException {
```
brendandburns: It's a little weird to use `File` for CertificateAuthority and `String` here, I think consistency one way or the other is probably better.

util/src/main/java/io/kubernetes/client/util/ConfigBuilder.java
```diff
@@ -0,0 +1,272 @@
...
+
+	public ApiClient build() {
+		ApiClient client = new ApiClient();
+
+		if( kubeConfig !=null) {
+			client = Config.fromConfig(kubeConfig);
+		}
+
+		if(defaultKubeConfigMode == true) {
+			try {
```
brendandburns: Let's get rid of the try/catch and just have `build()` throw, I think that's better.

brendandburns: Better: move `KubeConfig.loadDefaultKubeConfig()` up into the `setDefaultMode()` function.

util/src/main/java/io/kubernetes/client/util/ConfigBuilder.java
```diff
@@ -0,0 +1,272 @@
...
+				client = Config.fromConfig(KubeConfig.loadDefaultKubeConfig());
+			} catch (FileNotFoundException e) {
+				log.error("Unable to find the file", e);
+			}
+		}
+
+		if(clusterMode == true) {
+			try {
+				client = Config.fromCluster();
```
brendandburns: Same here.

util/src/main/java/io/kubernetes/client/util/ConfigBuilder.java
```diff
@@ -0,0 +1,272 @@
...
+			} catch (IOException e) {
+				log.error("Exception ->", e);
+			}
+		}
+
+		if(defaultClientMode ==true ) {
+			try {
+				client = Config.defaultClient();
```
brendandburns: here too?

util/src/main/java/io/kubernetes/client/util/ConfigBuilder.java
```diff
@@ -0,0 +1,272 @@
...
+			}
+		}
+
+		if (basePath != null ) {
+			if(basePath.endsWith("/")) {
+				basePath = basePath.substring(0, basePath.length() - 1);
+			}
+			client.setBasePath(basePath);
+		} else {
+			if((clusterMode == false) && (defaultClientMode == false) && (defaultKubeConfigMode == false)) {
```
brendandburns: I wouldn't throw here, I would just do:

```
if (client.getBasePath().length() == 0) {
   client.setBasePath("http://localhost:8080")
}
```

util/src/main/java/io/kubernetes/client/util/ConfigBuilder.java
```diff
@@ -0,0 +1,272 @@
...
+				client.setApiKeyPrefix("Bearer");
+			}
+			client.setApiKey(apiKey);
+		}
+
+		client.setVerifyingSsl(verifyingSsl);
+
+		if(certificateAuthorityFile != null) {
+			try {
+				client.setSslCaCert(new FileInputStream(certificateAuthorityFile));
```
brendandburns: Let's move this exception up to the `setCertificateAuthority` call...

brendandburns:
A few more comments, thanks for your patience, this is getting close!

util/src/main/java/io/kubernetes/client/util/ConfigBuilder.java
kondapally1989: @brendandburns 
is the following what you intend

```java
public ConfigBuilder setDefaultMode() {
		this.defaultMode = true;
		return this;
	}
build()  throws Exception{
....
if(defaultMode == true) {
	client = Config.fromConfig(KubeConfig.loadDefaultKubeConfig());
}
...
}
```

brendandburns: @kondapally1989 no, what I was thinking was this:

```java
public ConfigBuilder setDefaultMode() throws IOException {
      this.client = config.fromConfig(KubeConfig.loadDefaultKubeConfig());
}
```

kondapally1989: @brendandburns 
 if setDefaultMode()  points to Config.fromConfig(KubeConfig.loadDefaultKubeConfig())

and setDefaultClientMode points to Config.defaultClient();

isn't the naming convention confusing. instead of setDefaultMode() cant we have setDefaultKubeConfigMode()

lewisheadden:
An oddity about `Config` which has bothered me is that it returns `ApiClient` not configurations and I was wondering whether this should really be called `ClientBuilder` or similar given that it returns an `ApiClient`?

lewisheadden:
Also any thoughts on doing something like https://github.com/lewisheadden/java/commit/ac7c1b564a8bf9bf9922f4228d591e4478e37af4? It seems like it's hard to know which set of fields should be set for which auth method and this might simplify it.

kondapally1989:
ClientBuilder sounds apt.
and
 https://github.com/lewisheadden/java/commit/ac7c1b564a8bf9bf9922f4228d591e4478e37af4
@brendanburns  ?

brendandburns:
That approach seems reasonable...

kondapally1989:
ok i will change it to ClientBuilder.
@lewisheadden is it okay if push the changes you suggested or will you push after this is completed ?

lewisheadden:
@kondapally1989 I can open a PR afterwards if that's easier for you. The example I pushed needs cleaned up.

kondapally1989:
@lewisheadden  just changed the classname and few changes, you can create PR for the changes you suggested after this PR is done.

brendandburns:
LGTM, thanks for the patience.  There are a couple more cleanups, but they can wait for a follow-on PR.

util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
```diff
@@ -0,0 +1,242 @@
...
+	public String getApiKeyPrefix() {
+		return apiKeyPrefix;
+	}
+
+	public ClientBuilder setApiKeyPrefix(String apiKeyPrefix) {
+		this.apiKeyPrefix = apiKeyPrefix;
+		return this;
+	}
+
+	public ApiClient build() throws FileNotFoundException {
```
brendandburns: I'd like to see this get to a point where it doesn't throw `FileNotFoundException` I think that means moving the cert loading to the setter function.

Issue #174, opened 2018-02-03
Document or adopt standard formatting for Java code


lewisheadden:
I've noticed a few different styles of formatted code in the repository for non-generated code. Is there a standard formatter file used for Java code within the Kubernetes community?

If there is a default, can we add it to the documentation for this repo? If not, can we use something like [Google Java Format](https://github.com/google/google-java-format)?

brendandburns:
I'm fine with Google Java Format, that's more or less what I write in anyway...

PR #175, opened 2018-02-03
Consolidate logic from Config into ClientBuilder

lewisheadden:
The introduction of `ClientBuilder` was a great step but now we have logic scattered between `Config` and `ClientBuilder`. This PR introduces some abstractions and new methods in `ClientBuilder` to allow all `Config` functionality to be delegated to `ClientBuilder`.

In short this now allows
```
final ApiClient client = ClientBuilder.defaultClient();
```
as well as
```
final ApiClient client = ClientBuilder.standard()
  .setBasePath("https://localhost")
  .build();
```
which does the same as `ClientBuilder.defaultClient()` but overrides the `basePath`.

It also adds `Authentication` instead of lots of different fields to set.
```
final ApiClient client = ClientBuilder.standard()
  .setAuthentication(new UsernamePasswordAuthentication("user", "pass"))
  .build();
```

A more detailed overview is below:
- Adds `ClientBuilder.defaultClient()` which returns an `ApiClient` built from the default order of `KUBECONFIG`, `HOME/.kube/config`, Cluster config, and then fallback host.
- Removes authentication concerns from `ClientBuilder` and introduces `Authentication` as an interface which can be provided to the `ClientBuilder` and used to configure an `ApiClient`.
- Adds various implementations of `Authentication` for Username/Password, Token, Client Cert, and Kubeconfig.
- Replaces all logic in `Config` with calls to `ClientBuilder`.
- Sets the default value for `verifyingSsl` to `true` as per [docs](https://github.com/kubernetes-client/java/blob/master/kubernetes/src/main/java/io/kubernetes/client/ApiClient.java#L170).
- Simplifies flow in `SSLUtils` by moving all file or data decisions to a helper function in `KubeConfig`.
- Added Javadoc to `ClientBuilder`.

Changes in tests:
- Simplifies tests by removing need for temporary files and using `src/test/resources` instead.
- Expands test coverage for `ClientBuilder`.
- Added test coverage for `Authentication`s.

Apologies for the big PR. I'm also aware of a few formatting changes (I think it makes it more consistent with the rest of the code base) but I also opened #174 to try and resolve that.


util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
```diff
@@ -12,231 +12,217 @@
...
+    }
+    final File clusterCa = new File(SERVICEACCOUNT_CA_PATH);
+    if (clusterCa.exists()) {
+      return cluster();
+    }
+    return new ClientBuilder();
+  }
+
+  private static FileReader findConfigFromEnv() {
+    try {
```
brendandburns: Given that the only caller that calls `findConfigFromEnv` throws IOException, let's just mark this method as throwing `FileNotFoundException` and throw it through to the caller?

util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
```diff
@@ -12,231 +12,217 @@
...
+      }
+      return new FileReader(kubeConfig);
+    } catch (FileNotFoundException e) {
+      log.info("Could not find file specified in $KUBECONFIG");
+      return null;
+    }
+  }
+
+  private static FileReader findConfigInHomeDir() {
+    try {
```
brendandburns: Same thing wrt removing this try/catch block and just throwing.

util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
```diff
@@ -12,231 +12,217 @@
...
+   *  The {@link KubeConfig} to configure the builder from.
+   * @return <tt>ClientBuilder</tt> configured from the provided <tt>KubeConfig</tt>
+   * @throws IOException
+   *  if the files specified in the provided <tt>KubeConfig</tt> are not readable
+   */
+  public static ClientBuilder kubeconfig(KubeConfig config) throws IOException {
+    final ClientBuilder builder = new ClientBuilder();
+
+    String server = config.getServer();
+    if (!server.startsWith("http://") && !server.startsWith("https://")) {
```
brendandburns: Maybe overkill, but just test for the presence of `"://"` ?

util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
```diff
@@ -12,231 +12,217 @@
...
+      if (server.contains(":443")) {
+        server = "https://" + server;
+      } else {
+        server = "http://" + server;
+      }
+    }
+
+    if(config.verifySSL()) {
+      final byte[] caBytes = KubeConfig.getDataOrFile(config.getCertificateAuthorityData(),
```
brendandburns: This isn't required for `verifySSL` imagine if someone has a cluster with a valid globally issued CA, they may not need to load in a custom CA...

util/src/test/java/io/kubernetes/client/util/credentials/AccessTokenAuthenticationTest.java
```diff
@@ -0,0 +1,7 @@
+package io.kubernetes.client.util.credentials;
+
+import static org.junit.Assert.*;
+
+public class AccessTokenAuthenticationTest {
+
```
brendandburns: Either implement a test here or delete this file?

brendandburns:
Some minor comments, but basically LGTM.

util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
lewisheadden: Having the try/catch allows fall through in `ClientBuilder.standard()` if we can't find `~/.kube/config` or `$KUBECONFIG` otherwise we will fail and not try other methods. 

See https://github.com/lewisheadden/java/blob/CredentialProvider/util/src/main/java/io/kubernetes/client/util/ClientBuilder.java#L76-L90

util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
lewisheadden: Agreed. I wanted to preserve the original logic but will change to `.contains("://")`.

lewisheadden:
@brendandburns Addressed most of the comments and left some thoughts on others. Also added tests for Authentications that I thought I'd pushed on Friday. Note the Client Cert Auth dummy cert/key are valid until `Feb  4 16:38:56 2028 GMT`.

util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
lewisheadden: Added check to see if `caBytes` is null before making this call. Force `setVerifyingSsl` to true on the next line.

util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
brendandburns: Can we use `File.exists()` to test for existence prior to trying to open the file?

util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
brendandburns: This logic feels convoluted to me.

Why not:
```java
byte[] caBytes = ...
if (caBytes != null) {
   builder.setCertificateAuthority(caBytes);
}
builder.setVerifyingSsl(true);
```

That feels cleaner to me.

Yes, it's unlikely that someone would set verifySSL to false and still pass a ca authority, but I don't think there's any harm in allowing it...

brendandburns:
A few more small comments.  Thanks!

util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
lewisheadden: You are right. I will remove the `if` statement related to `config.verifySSL` and also make `setCertificateAuthority` no longer set `verifyingSsl` to `true` as that is the default value.

This should provide the following behaviour
- Having `certificate-authority` or `certificate-authority-file` set will make the client verify with provided CA cert.
- Having `insecure-skip-tls-verify` set to `false` will set `verifyingSsl` to `false` regardless of providing a CA cert above.

lewisheadden:
@brendandburns I think I've addressed all the issues. A few minor comments on `setCertificateAuthority`/`verifyingSsl`

brendandburns:
LGTM, thanks!

PR #182, opened 2018-02-09
Use junit system-rules dependency with test scope

elek:
com.github.stefanbirkner:system-rules is licensed under Common Public License 1.0, which is "weak copyleft" licence according to the[ ASF policies](https://www.apache.org/legal/resolved.html).

As it is used only in test files I propose to change the scope of the dependency to test. With test scope it won't be included in the final artifacts and will be easier to use this client project from Apache projects.


elek:
@lewisheadden I did it just now. Hopefully the cla bot will notice it.

brendandburns:
LGTM, thanks.

PR #184, opened 2018-02-14
Fixed listPodForAllNamespaces to follow along with the actual Example class

timothyjosefik:
[Example class here](https://github.com/kubernetes-client/java/blob/master/examples/src/main/java/io/kubernetes/client/examples/Example.java)


brendandburns:
LGTM, thanks!

PR #211, opened 2018-03-16
Notify on exec process end and read exit code

rjeberhard:
I've implemented reading the exec process' exit code and trigger a notifyAll() when the web socket is closed so that waitFor() and waitFor(long, TimeUnit) work correctly.

The exit code is written to stream "3" in a format like this:

command terminated with non-zero exit code: Error executing in Docker Container: 1

This is a little surprising, so the code here searches for the last colon, trims the result and tries to parse an integer.  Otherwise, the exit code is assumed to be 0.

### DIFF
--- a/util/src/main/java/io/kubernetes/client/Exec.java
+++ b/util/src/main/java/io/kubernetes/client/Exec.java
@@ -19,7 +19,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.util.concurrent.TimeUnit;
-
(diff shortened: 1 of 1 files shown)


rjeberhard:
I've clicked through the CLA.  Oracle already has an agreement that covers all employee contributions.  How do I re-trigger the CLA check so that the pull request shows clean checks?

rjeberhard:
I'm going to close this for now.  I'm still seeing a problem when exec'ing a command that doesn't need stdin.

PR #213, opened 2018-03-18
Exec improvements to read non-zero exit code and implement waitFor(long, TimeUnit)

rjeberhard:
Three improvements:

1) Correct exitValue() to match Process.exitValue() documentation and implement waitFor(long, TimeUnit).

2) When the web socket is being closed parse the message in channel 3, if any, since this will contain the non-zero exit value of the exec.

3) Delay closing web socket streams until destroy() to resolve timing issue around reading from and closing these streams.

@brendanburns My fault getting the request for beta3 release and this change confused in order.  I initially requested beta3 thinking that exec would work fully and then realizing I could contribute the improvements.  If possible, and if this change is approved, please hold beta3 so that this can be included.  Thanks!

Note: I still can't resolve the the fact that the API server returns 400 Bad Request any time stdin=false; however, it doesn't seem to hurt anything if I consistently pass stdin=true even if the command I'm exec'ing doesn't need any input.

### DIFF
--- a/util/src/main/java/io/kubernetes/client/Exec.java
+++ b/util/src/main/java/io/kubernetes/client/Exec.java
@@ -12,13 +12,19 @@ limitations under the License.
 */
 package io.kubernetes.client;
 
+import com.google.gson.reflect.TypeToken;
 import io.kubernetes.client.models.V1Pod;
+import io.kubernetes.client.models.V1Status;
+import io.kubernetes.client.models.V1StatusCause;
+import io.kubernetes.client.models.V1StatusDetails;
 import io.kubernetes.client.util.WebSocketStreamHandler;
 import io.kubernetes.client.util.WebSockets;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.lang.reflect.Type;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 import org.apache.commons.lang3.StringUtils;
@@ -184,7 +190,7 @@ public class Exec {
     return exec;
   }
 
-  static int parseExitCode(InputStream inputStream) {
+  static int parseExitCode(ApiClient client, InputStream inputStream) {
     try {
       int available = inputStream.available();
 
@@ -193,24 +199,38 @@ public class Exec {
 
       byte[] b = new byte[available];
       inputStream.read(b);
-      String result = new String(b, "UTF-8");
-      int idx = result.lastIndexOf(':');
-      if (idx > 0) {
-        try {
-          return Integer.parseInt(result.substring(idx + 1).trim());
-        } catch (NumberFormatException nfe) {
-          log.error("Error parsing exit code from status channel response", nfe);
+      String body = new String(b, "UTF-8");
+
+      Type returnType = new TypeToken<V1Status>() {}.getType();
+      V1Status status = client.getJSON().deserialize(body, returnType);
+      if ("Success".equals(status.getStatus())) return 0;
+
+      if ("NonZeroExitCode".equals(status.getReason())) {
+        V1StatusDetails details = status.getDetails();
+        if (details != null) {
+          List<V1StatusCause> causes = details.getCauses();
+          if (causes != null) {
+            for (V1StatusCause cause : causes) {
+              if ("ExitCode".equals(cause.getReason())) {
+                try {
+                  return Integer.parseInt(cause.getMessage());
+                } catch (NumberFormatException nfe) {
+                  log.error("Error parsing exit code from status channel response", nfe);
+                }
+              }
+            }
+          }
         }
       }
-    } catch (IOException io) {
-      log.error("Error parsing exit code from status channel response", io);
+    } catch (Throwable t) {
+      log.error("Error parsing exit code from status channel response", t);
     }
 
     // Unable to parse the exit code from the content
     return -1;
   }
 
-  private static class ExecProcess extends Process {
+  private class ExecProcess extends Process {
     private final WebSocketStreamHandler streamHandler;
     private volatile int statusCode;
     private final Map<Integer, InputStream> input = new HashMap<>();
@@ -222,7 +242,7 @@ public class Exec {
             @Override
             public void close() {
               if (statusCode == -1) {
-                int exitCode = parseExitCode(ExecProcess.this.getInputStream(3));
+                int exitCode = parseExitCode(apiClient, ExecProcess.this.getInputStream(3));
 
                 // notify of process completion
                 synchronized (ExecProcess.this) {
--- a/util/src/main/java/io/kubernetes/client/Exec.java
+++ b/util/src/main/java/io/kubernetes/client/Exec.java
@@ -19,7 +19,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.util.concurrent.TimeUnit;
-
 import org.apache.commons.lang3.StringUtils;
 
 public class Exec {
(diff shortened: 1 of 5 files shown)


util/src/main/java/io/kubernetes/client/Exec.java
```diff
@@ -170,20 +171,59 @@ public Process exec(
...
+    public ExecProcess() throws IOException {
       this.statusCode = -1;
+      this.streamHandler =
+          new WebSocketStreamHandler() {
+            @Override
+            public void close() {
+              if (statusCode == -1) {
+                InputStream inputStream = getInputStream(3);
+                int exitCode = 0;
+                try {
```
brendandburns: This code seems fragile. I'd love to see it extracted into a separate method and unit tested.

util/src/main/java/io/kubernetes/client/Exec.java
```diff
@@ -170,20 +171,59 @@ public Process exec(
...
+                  if (available > 0) {
+                    byte[] b = new byte[available];
+                    inputStream.read(b);
+                    String result = new String(b, "UTF-8");
+                    int idx = result.lastIndexOf(':');
+                    if (idx > 0) {
+                      try {
+                        exitCode = Integer.parseInt(result.substring(idx + 1).trim());
+                      } catch (NumberFormatException nfe) {
+                        // no-op
```
brendandburns: We should at least log warning here.

util/src/main/java/io/kubernetes/client/Exec.java
```diff
@@ -170,20 +171,59 @@ public Process exec(
...
+                      }
+                    }
+                  }
+                } catch (IOException e) {
+                }
```
brendandburns: Log here too.

brendandburns:
This seems reasonable, but a few comments to make it a slightly better change.

Happy to hold the release for this (at least for a little while)

Thanks for the PR!

rjeberhard:
Thanks, @brendandburns for the review comments.  I'll update the PR today.  Do you think it's worth creating an issue against Kubernetes itself?  I was really surprised by how the API server returns the exit code.  The content on channel 3 is the following:

command terminated with non-zero exit code: Error executing in Docker Container: 1

I was expecting some type of ExecStatus object.

rjeberhard:
I went ahead and created https://github.com/kubernetes/kubernetes/issues/61346

rjeberhard:
@brendandburns I've updated per your code review comments (thanks!).  Also, I found two other race conditions:

1) Caller could try to write to stdin before the websocket was open; write should wait for open()
2) WebSocketStreamHandler.close() was closing the PipedInputStreams before they could be drained (the ugly Thread.sleep() in the sample).  Switch to just closing the PipedOutputStreams and saving the close() of the piped input streams for Proc.destroy().

util/src/main/java/io/kubernetes/client/Exec.java
```diff
@@ -170,20 +177,65 @@ public Process exec(
...
-    WebSocketStreamHandler handler = new WebSocketStreamHandler();
-    ExecProcess exec = new ExecProcess(handler);
+    ExecProcess exec = new ExecProcess();
+    WebSocketStreamHandler handler = exec.getHandler();
     WebSockets.stream(path, "GET", apiClient, handler);
 
     return exec;
   }
 
+  static int parseExitCode(InputStream inputStream) {
```
brendandburns: This returns zero on a number of edge cases which seems wrong.  I suggested some refactors...

util/src/main/java/io/kubernetes/client/Exec.java
```diff
@@ -170,20 +177,65 @@ public Process exec(
...
+    int exitCode = 0;
```
brendandburns: delete this.

util/src/main/java/io/kubernetes/client/Exec.java
```diff
@@ -170,20 +177,65 @@ public Process exec(
...
+    try {
+      int available = inputStream.available();
+      if (available > 0) {
+        byte[] b = new byte[available];
+        inputStream.read(b);
+        String result = new String(b, "UTF-8");
+        int idx = result.lastIndexOf(':');
+        if (idx > 0) {
+          try {
+            exitCode = Integer.parseInt(result.substring(idx + 1).trim());
```
brendandburns: just `return Integer.parseInt(...)` here.

util/src/main/java/io/kubernetes/client/Exec.java
```diff
@@ -170,20 +177,65 @@ public Process exec(
...
+          } catch (NumberFormatException nfe) {
+            log.error("Error parsing exit code from status channel response", nfe);
+          }
+        }
+      }
+    } catch (IOException io) {
+      log.error("Error parsing exit code from status channel response", io);
+    }
+
+    return exitCode;
```
brendandburns: return -1 here.

util/src/main/java/io/kubernetes/client/Exec.java
```diff
@@ -170,20 +177,65 @@ public Process exec(
...
 
-    public ExecProcess(WebSocketStreamHandler handler) throws IOException {
-      this.streamHandler = handler;
+    public ExecProcess() throws IOException {
       this.statusCode = -1;
+      this.streamHandler =
+          new WebSocketStreamHandler() {
+            @Override
+            public void close() {
+              if (statusCode == -1) {
```
brendandburns: I feel like we may want to loop here in case the network stream is blocked up or otherwise unavailable.  Obviously probably need to time out eventually...

brendandburns:
@rjeberhard sorry added a few more comments :)

rjeberhard:
No problem!  I can tell we're going to be friends.  I just wish it wasn't my code that needed so much commenting.

util/src/main/java/io/kubernetes/client/Exec.java
rjeberhard: I checked the Web Socket specification here: https://datatracker.ietf.org/doc/rfc6455/?include_text=1.  The server has sent a Close message.  This is why there is the callback to WebSocketListener.onClose(), which is then calling this close() method.

Therefore, we know that there will be no more calls to onMessage().

rjeberhard:
My review of the spec. and my testing here suggests that there will not be further data-carrying messages on the web socket after the close().  Therefore, we don't need the suggested looping.

Note: web socket channel 3 is presently used to send back only non-zero exit codes.  If the exit code is 0, then this channel has no content.  In the enh. request I filed against k8s, I didn't mention this.

About the only extra check we could do is to validate the web socket close code in WebSockets$Listener.onClose(int, String).  This code will be 1000 on a normal termination of the web socket session.  I'd prefer to do any work there as a separate PR.

rjeberhard:
Based on the conversation on https://github.com/kubernetes/kubernetes/issues/61346, I could adjust this solution slightly.

Presently, I'm waiting for the web socket conversation to end, then checking to see if channel 3 has pending content.  If it does, parsing that to get the exit code.  Otherwise, (if channel 3 has no content), then knowing that this indicates the exit code is 0.

An alternate implementation would be, any message to channel 3 is parsed as the exit code.  When the web socket channel closes, then the exit code is 0 unless an earlier message had arrived for channel 3.

rjeberhard:
One more update coming for my idea, as described above.  Will update here when it's okay to commit, depending on your review.

rjeberhard:
@brendandburns okay, I think this is ready to go.

util/src/main/java/io/kubernetes/client/Exec.java
```diff
@@ -170,20 +186,90 @@ public Process exec(
...
+  static int parseExitCode(ApiClient client, InputStream inputStream) {
+    try {
+      Type returnType = new TypeToken<V1Status>() {}.getType();
+      String body;
+      try (final Reader reader = new InputStreamReader(inputStream)) {
+        body = CharStreams.toString(reader);
+      }
+
+      V1Status status = client.getJSON().deserialize(body, returnType);
+      if ("Success".equals(status.getStatus())) return 0;
```
brendandburns: nit: would be great to have these be constants somewhere in a utility (it's really too bad the swagger doesn't capture these kind of constants)

util/src/main/java/io/kubernetes/client/util/WebSocketStreamHandler.java
```diff
@@ -38,76 +39,97 @@
...
     } catch (IOException ex) {
       log.error("Error writing message", ex);
     }
   }
 
+  protected void handleMessage(int stream, InputStream inStream) {
+    try {
+      OutputStream out = getSocketInputOutputStream(stream);
+      ByteStreams.copy(inStream, out);
+    } catch (IOException ex) {
```
brendandburns: Given that both callers of `handleMessage` still catch `IOException` I would just have `handleMessage` throw IOException.

brendandburns:
Couple of small cleanups, but LGTM. I will merge and you can send cleanups as an additional PR...

PR #218, opened 2018-03-22
Regenerate client.

brendandburns:
@lewisheadden


lewisheadden:
Thanks @brendandburns! What was the issue in the end?

brendandburns:
It turns out that there are two active benches in the code generator.

The patch I wanted was in the master branch which doesn't support the additional properties feature, which was in the other branch (3.0.0)

So instead I used the 3.0.0 branch and directly added the fix I wanted from the master branch...

Ugly but it gets us where we want to be....

--brendan

kubernetes/.swagger-codegen/VERSION
```diff
@@ -1 +1 @@
-2.4.0-SNAPSHOT
\ No newline at end of file
+2.3.0-SNAPSHOT
```
lwander: Is this correct given that you used the 3.0.0  branch?

brendandburns: Yeah, that's what it generated. I have no idea why :)

PR #227, opened 2018-04-02
Add default true to stdout and stderr on Exec.

### LINKED ISSUE #225: Exec not working for k8s 1.9, 1.10 servers
The recent work on Exec and ExecExample is working great for k8s 1.7.x and 1.8.x servers; however, it's not working against 1.9 and 1.10.

The behavior with 1.9 and 1.10 is that no messages come in on channel 1, but instead there is an immediate empty message on channel 3 (the signaling channel).

I'm happy to help debug and resolve this, but I'd be grateful for insights into what changed in the API server's use of web sockets.

brendandburns:
Fixes #225, though I would argue this is a breaking change for Kubernetes 1.9 and 1.10.

I'm going to try to repro and file a bug on upstream...

Sorry for the big diff. I also stripped `\r\n` from the file, I can revert for a cleaner diff if that is preferred.

The key diff is:

```java
+            + "&stdout=true"
+            + "&stderr=true"
```

in `Exec.java`

@rjeberhard

### DIFF
--- a/util/src/main/java/io/kubernetes/client/Exec.java
+++ b/util/src/main/java/io/kubernetes/client/Exec.java
@@ -88,6 +88,8 @@ public class Exec {
             + "/exec?"
             + "stdin="
             + stdin
+            + "&stdout=true"
+            + "&stderr=true"
             + "&tty="
             + tty
             + (container != null ? "&container=" + container : "")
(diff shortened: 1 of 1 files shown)


brendandburns:
@rjeberhard I think I might have originally written this file on Windows, I think that they've been there all along...

Merging.

Issue #233, opened 2018-04-08
Equals symbols in base64 encoded data are escaped to HTML Safe characters resulting in errors when making HTTP calls.


archcosmo:
I've noticed this when creating a secret with .dockerconfigjson set to an array of bytes that is base64 encoded to a string that contains an equals character. The result is '\u003d' is inserted in to the request body which then results in an error as Kubernetes cannot parse the invalid base64 characters.

It seems to occur in the JSON class which uses Gson to encode the data, but the Gson htmlSafe if true by default, so the characters are escaped.

This demonstrates the issue:

```
package comp3200.am3g15.forestgrowthsrc.k8s.client;

import com.google.gson.Gson;
import com.google.gson.GsonBuilder;
import com.google.gson.TypeAdapter;
import com.google.gson.stream.JsonReader;
import com.google.gson.stream.JsonWriter;
import io.kubernetes.client.JSON;
import io.kubernetes.client.models.V1Secret;
import okio.ByteString;

import java.io.IOException;
import java.util.HashMap;
import java.util.Map;

public class Base64Test {

    public static void main(String[] args) {
        V1Secret secret = new V1Secret();
        Map<String, byte[]> secretData = new HashMap<>();
        secretData.put(".dockerconfigjson", new String("{\"auths\":{\"server\":{\"username\":\"username\",\"password\":\"password\",\"email\":\"email\",\"auth\":\"authhh\"}}}").getBytes());
        secret.setData(secretData);

        //What occurs when CoreV1Api.createNamespacedSecret() is called
        JSON json = new JSON();
        String erroneous = json.serialize(secret);

        //The Equivalent without the JSON class
        Gson gson = new GsonBuilder().registerTypeAdapter(byte[].class, new ByteArrayAdapter()).create();
        String stillErroneous = gson.toJson(secret);

        //With HTML Safety Turned Off
        gson = new GsonBuilder().disableHtmlEscaping().registerTypeAdapter(byte[].class, new ByteArrayAdapter()).create();
        String correct = gson.toJson(secret);

        System.out.printf("Bad: %s\nBad: %s\nCorrect: %s", erroneous, stillErroneous, correct);
    }

    //Pulled from the JSON class
    public static class ByteArrayAdapter extends TypeAdapter<byte[]> {
        public ByteArrayAdapter() {
        }

        public void write(JsonWriter out, byte[] value) throws IOException {
            if (value == null) {
                out.nullValue();
            } else {
[...]

PR #240, opened 2018-04-18
Fix issue #233

archcosmo:
This should fix issue #233.

### DIFF
--- a/kubernetes/src/main/java/io/kubernetes/client/JSON.java
+++ b/kubernetes/src/main/java/io/kubernetes/client/JSON.java
@@ -126,11 +126,14 @@ public class JSON {
 
         @Override
         public void write(JsonWriter out, byte[] value) throws IOException {
+            boolean oldHtmlSafe = out.isHtmlSafe();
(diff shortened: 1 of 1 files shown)


brendandburns:
LGTM, though you need to sign the CLA.

Can you add a unit test to validate this behavior so that we don't regress in the future?

Also, if you feel like it, it might be worth making a similar change in the swagger-codegen repository since that's where this code was derived from...

Thanks!

brendandburns:
oh, wait, nevermind about the CLA, it just turned green...

But please do add a unit test.

Bas83:
Just curious, why preserve the htmlSafe boolean? Is there ever a reason to create htmlSafe output?

brendandburns:
I think that is correct in terms of "do no harm" or "least surprise" you want to preserve whatever the settings were prior to you mucking with things...

archcosmo:
Yes, I wasn't sure where else the writer was being used and didn't want to affect anything else.
I've written a jUnit test, does it fit with the convention?

Issue #221, opened 2018-03-27
Error while creating Kubernetes Job


mmmaks:
When creating Kubernetes Job, with yaml : 
    `  kind: Job`
    `apiVersion: batch/v1`
Error : 
> java.io.IOException: Unknown apiVersionKind: batch/v1/Job known kinds are:

and then a list of known kinds.
But [here](https://v1-9.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.9/#job-v1-batch) shows `apiVersion: batch/v1` for job.

Also, when I try to create Job with yaml : 
`  kind: Job`
   `apiVersion: v1`

Error : 
`io.kubernetes.client.ApiException: Bad Request
	at io.kubernetes.client.ApiClient.handleResponse(ApiClient.java:882)
	at io.kubernetes.client.ApiClient.execute(ApiClient.java:798)
	at io.kubernetes.client.apis.BatchV1Api.createNamespacedJobWithHttpInfo(BatchV1Api.java:160)
	at io.kubernetes.client.apis.BatchV1Api.createNamespacedJob(BatchV1Api.java:144)`

brendandburns:
Yeah, somehow we're losing the API group here...

@mbohlool do you know how we are doing the API group in the swagger?

Why does `Apps` and `Extensions` get generated into the object name but not `Batch`?

brendandburns:
I think the tl;dr; is that we may need to do more work on API discovery here.  Probably need to download the swagger spec from the server.  *sigh*

For now you will want to call:

```java
Yaml.addModelMap("batch/v1", V1Job.class)
```

Somewhere in your code before you try to load that object from YAML that should manually specify the mapping we should have discovered auto-magically.

mmmaks:
ok, Thanks

iming0319:
@brendandburns 
What I meet is "Unknown apiVersionKind: apps/v1beta1/Deployment known kinds are:...."
And I got a know kind "apps/v1/beta1Deployment".

so, I think this issue is related to the method initModelMap(), in Yaml.java
We shoud edit it 
form:
if (name.startsWith("AppsV1")) {
      return "apps/v1";
}
if (name.startsWith("AppsV1beta1")) {
    return "apps/v1beta1";
}
to:
if (name.startsWith("AppsV1beta1")) {
    return "apps/v1beta1";
}
if (name.startsWith("AppsV1")) {
    return "apps/v1";
}

brendandburns:
[...]

PR #246, opened 2018-04-20
Changes to the Yaml object auto-detection

archcosmo:
There are some issues with the way the Yaml class auto-detects the Kubernetes object during the `load` function.

Commit `9e0b3b8` fixes a bug where the Apps Api is checked in the wrong order.
Commit `dd93b9d` auto-detects in a different way to include more of the available Api groups.
Commit `5a0439c` attempts to temporarily address the issue #221 where some classes have not been prefixed with their Api groups, until the class names have been generated properly.

See `dd93b9d`'s commit message for a list of Api prefixes that are missing from class names.

Keep what you want

### DIFF
--- a/util/src/main/java/io/kubernetes/client/util/Yaml.java
+++ b/util/src/main/java/io/kubernetes/client/util/Yaml.java
@@ -18,7 +18,13 @@ import java.io.FileReader;
 import java.io.IOException;
 import java.io.Reader;
 import java.io.StringReader;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import org.apache.commons.lang3.tuple.MutablePair;
+import org.apache.commons.lang3.tuple.Pair;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


util/src/main/java/io/kubernetes/client/util/Yaml.java
```diff
@@ -18,61 +18,93 @@
 import java.io.IOException;
 import java.io.Reader;
 import java.io.StringReader;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
brendandburns: Style nit: import specific classes don't use wildcards.

util/src/main/java/io/kubernetes/client/util/Yaml.java
```diff
@@ -18,61 +18,93 @@
...
+    apiVersions.add("V2beta1");
+    apiVersions.add("V2alpha1");
+    apiVersions.add("V1beta2");
+    apiVersions.add("V1beta1");
+    apiVersions.add("V1alpha1");
+    apiVersions.add("V1");
+  }
+
+  private static String[] getApiGroup(String name) {
+    String[] parts = new String[2];
```
brendandburns: use the `Pair` class from apache commons lib instead?

brendandburns:
Thanks for the PR. Some minor comments, but LGTM.

archcosmo:
Ah my bad, my IDE was grouping imports. Updated as advised.

brendandburns:
LGTM, thanks!

Issue #161, opened 2017-12-27
ExtensionsV1beta1Api deleteNamespacedReplicaSet throw 500 Error


PGDream:
at call ExtensionsV1beta1Api.deleteNamespacedReplicaSet throw Internal Server Error , code 500;
curl -X DELETE "http://1xxxx/apis/extensions/v1beta1/namespaces/default/deployments/api-test-nginx" , Deployment can normal deleted.
code: 
       <pre>
        ExtensionsV1beta1Api api = new ExtensionsV1beta1Api();
        V1DeleteOptions options = new V1DeleteOptions();
        options.setApiVersion("extensions/v1beta1");
        options.setKind("Deployment");
        api.deleteNamespacedDeployment("api-test-nginx", "default", options, null, null, null, null);
      </pre>

exception:
      io.kubernetes.client.ApiException: Internal Server Error

	at io.kubernetes.client.ApiClient.handleResponse(ApiClient.java:882)
	at io.kubernetes.client.ApiClient.execute(ApiClient.java:798)
	at io.kubernetes.client.apis.ExtensionsV1beta1Api.deleteNamespacedReplicaSetWithHttpInfo(ExtensionsV1beta1Api.java:3129)
	at io.kubernetes.client.apis.ExtensionsV1beta1Api.deleteNamespacedReplicaSet(ExtensionsV1beta1Api.java:3109)
	at TestClient.deleteReplicationSet(TestClient.java:49)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:47)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:44)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:271)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:70)
[...]

PR #255, opened 2018-04-28
modify the type of status field of V1Staus from String to Object

SataQiu:
Prevent exceptions from deleting resources, such as namespace.
Issue: ExtensionsV1beta1Api deleteNamespacedReplicaSet throw 500 Error #161


kondapally1989:
These Model classes are auto generated via swagger-client gen and should not be edited manually.

Issue #257, opened 2018-04-30
Exec not supporting special characters in commands


kondapally1989:
Exec is not able to execute commands followed after specials characters.
example: 
               ls ; pwd 
               ls && pwd
for above commands only first ls commands get executed

Reason: in Exec.makePath param values, are not encoded.
             it can be observed even in kubectl exec command param values are encoded.
              POST https://192.168.99.100:8443/api/v1/namespaces/default/pods/hello-node-6d4df5494c-ts8vj/exec?command=bash&command=-c&command=**ls%3Bpwd**&container=hello-node&container=hello-node&stderr=true&stdout=true

kondapally1989:
closing the issue. fixed in https://github.com/kubernetes-client/java/pull/258

PR #258, opened 2018-04-30
fix for exec command encoding issue

kondapally1989:
fixes issue https://github.com/kubernetes-client/java/issues/257 by performing URLEncoder.encode on list of commands

### DIFF
--- a/util/src/main/java/io/kubernetes/client/Exec.java
+++ b/util/src/main/java/io/kubernetes/client/Exec.java
@@ -27,7 +27,6 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.OutputStream;
 import java.io.Reader;
-import java.io.UnsupportedEncodingException;
 import java.lang.reflect.Type;
 import java.net.URLEncoder;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


util/src/main/java/io/kubernetes/client/Exec.java
```diff
@@ -74,12 +76,11 @@ public void setApiClient(ApiClient apiClient) {
...
 
   private String makePath(
-      String namespace,
-      String name,
-      String[] command,
-      String container,
-      boolean stdin,
-      boolean tty) {
+      String namespace, String name, String[] command, String container, boolean stdin, boolean tty)
+      throws UnsupportedEncodingException {
```
brendandburns: Hrm, I feel like populating this forward is probably not right. The encoder is _extremely_ unlikely to throw this since UTF-8 is supported everywhere (and it better be, or HTTP won't work right...)

So I would just try/catch and then throw RuntimeException or some other sort of exception that doesn't have to be caught.

util/src/main/java/io/kubernetes/client/Exec.java
```diff
@@ -80,6 +81,13 @@ private String makePath(
       String container,
       boolean stdin,
       boolean tty) {
+    for (int i = 0; i < command.length; i++) {
+      try {
+        command[i] = URLEncoder.encode(command[i], "UTF-8");
+      } catch (Exception ex) {
```
brendandburns: Don't catch `Exception` catch `UnknownFormatException`

brendandburns: oops I meant `UnsupportedEncodingException`

brendandburns:
LGTM, thanks!

PR #270, opened 2018-05-20
 Migrated projects to OkHttp 3 library

wlsc:
Please review the following PR.
I've migrated the OkHttp 2.x library to a new 3.x version.

Sidenote: WebSockets are now a part of OkHttp 3 core library.


brendandburns:
@wlsc was this done via the code generator? We use https://github.com/swagger-api/swagger-codegen
to generate the API code, and thus we need the changes to come in via that automation.

If this was done via regenerating the code, great, but if not, you'll need to add okhttp3 support to swagger-codegen first and then regenerate the code.

Please let me know if you have questions or want pointers.

Thanks!

wlsc:
@brendanburns unfortunately, I have modified it manually. It would be great if you could share any command line options on how to generate the API based on your experience.

brendandburns:
@wlsc The basic instructions are here:

https://github.com/kubernetes-client/java#update-client

Where `${GEN_REPO_BASE}` is checked out from:

https://github.com/kubernetes-client/gen

You will need to update the `swagger-codegen` project here:

https://github.com/swagger-api/swagger-codegen/tree/master/modules/swagger-codegen/src/main/resources/Java/libraries/okhttp-gson

They're generally pretty receptive to code changes, but I've never tried to do one as big as this would be.

Let me know if you need help. I'm going to close this PR for now since we can't move forward as is.

Thanks!

PR #283, opened 2018-05-30
Support equals() for IntOrString

russgold:
Implemented equals() and hash() for the IntOrString class. Unit test included

### DIFF
--- a/kubernetes/src/main/java/io/kubernetes/client/custom/IntOrString.java
+++ b/kubernetes/src/main/java/io/kubernetes/client/custom/IntOrString.java
@@ -1,12 +1,13 @@
 package io.kubernetes.client.custom;
 
+import java.io.IOException;
+import java.util.Objects;
+
 import com.google.gson.TypeAdapter;
 import com.google.gson.annotations.JsonAdapter;
 import com.google.gson.stream.JsonReader;
 import com.google.gson.stream.JsonToken;
 import com.google.gson.stream.JsonWriter;
-
-import java.io.IOException;
 
 @JsonAdapter(IntOrString.IntOrStringAdapter.class)
 public class IntOrString {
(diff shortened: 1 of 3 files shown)


kubernetes/src/main/java/io/kubernetes/client/custom/IntOrString.java
```diff
@@ -47,7 +48,22 @@ public Integer getIntValue() {
...
     	return (isInt ? String.valueOf(intValue) : strValue);
-    }	    
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        return this == o || (o instanceof IntOrString && equals((IntOrString) o));
+    }
+
+    private boolean equals(IntOrString o) {
```
brendandburns: minor nit, but is there value in making this a separate function?

Why not:

```java
if (this == o) {
   return true;
}
if (! o instanceof IntOrString) {
   return false;
}
IntOrString other = (IntOrString) o;
...
}
```

brendandburns:
Thanks for the PR, one minor nit...

kubernetes/src/main/java/io/kubernetes/client/custom/IntOrString.java
russgold: I use this approach because I find it easier to read. The approach you show uses seven lines of code to accomplish what the submitted change does in one. Extra functions can act as a very simple way of documenting functionality (see Robert Martin's "Clean Code"), and get optimized away at runtime.

But my goal is just getting this fixed so that my project doesn't need a workaround. if using my approach would hold up the pull request, I am perfectly happy to see it changed.

brendandburns: I don't feel strongly enough about this to have you correct it (though overally I think my suggestion is shorter once you consider the lines to define the new function..)

I'm merging as-is.

PR #298, opened 2018-06-23
Correctly load base64-encoded strings from Secret Yaml files.

johndaniels:
Loading secrets from YAML files was failing for me. Let me know if this change is the right way to fix that, thanks!

### DIFF
--- a/util/src/main/java/io/kubernetes/client/util/Yaml.java
+++ b/util/src/main/java/io/kubernetes/client/util/Yaml.java
@@ -19,11 +19,8 @@ import java.io.FileReader;
 import java.io.IOException;
 import java.io.Reader;
 import java.io.StringReader;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
+import okio.ByteString;
 import org.apache.commons.lang3.tuple.MutablePair;
 import org.apache.commons.lang3.tuple.Pair;
 import org.slf4j.Logger;
@@ -269,6 +266,9 @@ public class Yaml {
       if (node.getType() == IntOrString.class) {
         return constructIntOrString((ScalarNode) node);
       }
+      if (node.getType() == byte[].class) {
+        return constructByteArray((ScalarNode) node);
+      }
       return super.constructObject(node);
     }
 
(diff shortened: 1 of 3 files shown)


johndaniels:
Should be all good on the contributor agreement.

util/src/main/java/io/kubernetes/client/util/Yaml.java
```diff
@@ -19,11 +19,8 @@
 import java.io.IOException;
 import java.io.Reader;
 import java.io.StringReader;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
brendandburns: As a matter of style, we don't use wild-card imports.

util/src/test/java/io/kubernetes/client/util/YamlTest.java
```diff
@@ -15,14 +15,12 @@
 import static org.junit.Assert.*;
 
 import com.google.common.io.Resources;
-import io.kubernetes.client.models.AppsV1beta1Deployment;
-import io.kubernetes.client.models.V1ObjectMeta;
-import io.kubernetes.client.models.V1Service;
-import io.kubernetes.client.models.V1ServicePort;
+import io.kubernetes.client.models.*;
```
brendandburns: as above wrt to wildcard imports.

brendandburns:
I think this is ok, but I think that we need the encoding side of this also?  e.g. if I place bytes into a `V1Secret` object it should encode them too?

Basically, can you add a class that tries to round-trip a `V1Secret` if that test passes then I think we're good to go.

johndaniels:
Still aware of this. I started on changes but didn't have time to finish them. It looks like round-tripping with YAML wasn't tested at all so it's a little bit more involved to get that working but I think it will be good.

brendandburns:
@johndaniels thanks! sorry to send you off on a bigger change, but it will definitely help improve the code!

johndaniels:
Ok, fixed the issue with the wildcard imports and got everything successfully round-tripping back to Yaml. Let me know if anything else needs changing.

brendandburns:
Thanks, changes LGTM, merging.

PR #314, opened 2018-07-17
Avoid use SystemClassLoader on Yaml.

takashima0411:
Now, Yaml class uses SystemClassLoader for reflection.

If this library is  loaded by an another class loader.
SystemClassLoader doesn't know Yaml class.
Therefore`cp.getTopLevelClasses("io.kubernetes.client.models");` returns an empty list.


brendandburns:
LGTM, thanks!

PR #316, opened 2018-07-17
Format Compatibility Table in README

joelhandwell:
ref: https://www.tablesgenerator.com/markdown_tables
Click rich diff button to see how it looks like:
image


brendandburns:
Thanks!

Issue #333, opened 2018-08-02
[ERROR] Failed to execute goal on project client-java-api: Could not resolve dependencies for project io.kubernetes:client-java-api:bundle:3.0.0-SNAPSHOT: Could not find artifact com.sun:tools:jar:1.7 at specified path /usr/java/jdk-9.0.1/../lib/tools.jar -> [Help 1]


mtsr:
Another Error I had when building is this weird dependency in io.sundr.builder-annotations:

```
[ERROR] Failed to execute goal on project client-java-api: Could not resolve dependencies for project io.kubernetes:client-java-api:bundle:3.0.0-SNAPSHOT: Could not find artifact com.sun:tools:jar:1.7 at specified path /usr/java/jdk-9.0.1/../lib/tools.jar -> [Help 1]
```

The dependency in question is
```
            <dependency>
                <groupId>com.sun</groupId>
                <artifactId>tools</artifactId>
                <version>${sun.tools.version}</version>
                <scope>system</scope>
                <systemPath>${java.home}/../lib/tools.jar</systemPath>
            </dependency>
```
which is clearly invalid on some systems, such as mine. The quick solution is to exclude this transitive dependency in kubernetes/pom.xml, which makes the build succeed and leaves everything in a working state as far as I can tell. This can be done by adding the following to the io.sundr.builder-annotations dependency:

```
      <exclusions>
          <exclusion>
              <artifactId>tools</artifactId>
              <groupId>com.sun</groupId>
          </exclusion>
      </exclusions>
```

PR #334, opened 2018-08-02
Exlude com.sun.tools from io.sundr.builder-annotations transitive dep…

mtsr:
…endencies

Solves https://github.com/kubernetes-client/java/issues/333

### DIFF
--- a/kubernetes/pom.xml
+++ b/kubernetes/pom.xml
@@ -134,6 +134,12 @@
       <artifactId>builder-annotations</artifactId>
       <version>${sundrio.version}</version>
       <scope>compile</scope>
+      <exclusions>
+        <exclusion>
+          <artifactId>tools</artifactId>
+          <groupId>com.sun</groupId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>io.swagger</groupId>


mtsr:
Signed CLA for corporate email address...

mtsr:
Signed CLA for corporate email address...
/check-cla

mtsr:
Hmm, again
/check-cla

jglick:
This does not suffice. In a project using this library I cannot build on Windows:

```
[ERROR] 'dependencies.dependency.systemPath' for com.sun:tools:jar must specify an absolute path but is /usr/lib/jvm/java-8-jdk/jre/../lib/tools.jar @ io.sundr:sundr-codegen:[unknown-version], c:\…workspace…\.repository\io\sundr\sundr-codegen\0.9.2\sundr-codegen-0.9.2.pom, line 54, column 19
```

(https://github.com/jenkinsci/google-kubernetes-engine-plugin/pull/4#issuecomment-461818176 was a similar case I encountered a while back.)

Picking up https://github.com/sundrio/sundrio/pull/118 does not help. The problem is not that `com.sun:tools` is in the dependency tree but that the POM itself is considered invalid.

Workaround is to

```xml
<exclusion>
    <groupId>io.sundr</groupId>
    <artifactId>sundr-codegen</artifactId>
</exclusion>
```

when using this lib.

PR #336, opened 2018-08-07
adding sample yaml to test examples.jar from with in a pod

kondapally1989:
- updated docker file with latest examples snpashot jar 
- changed openjdk:8-alpine to openjdk:8-jre as  alpine some times causing tls handshake error when pulling image(inside minikube docker)
- added createPod.sh which  creates a pod and runs Example.java(list pods for all namespaces) on starting of pod

### DIFF
--- a/examples/Dockerfile
+++ b/examples/Dockerfile
@@ -1,6 +1,6 @@
-FROM openjdk:8-alpine
+FROM openjdk:8-jre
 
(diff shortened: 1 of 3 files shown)


brendandburns:
LGTM.  (fwiw, it's @brendandburns, note the 'd' :)

PR #342, opened 2018-08-20
set content-type to merge-patch+json while building nodePatchCall

hellowenqi:
We should set content-type to merge-patch+json while building nodePatchCall.

### DIFF
--- a/kubernetes/src/main/java/io/kubernetes/client/apis/CoreV1Api.java
+++ b/kubernetes/src/main/java/io/kubernetes/client/apis/CoreV1Api.java
@@ -22610,7 +22610,7 @@ public class CoreV1Api {
         if (localVarAccept != null) localVarHeaderParams.put("Accept", localVarAccept);
 
         final String[] localVarContentTypes = {
-            "application/json-patch+json", "application/merge-patch+json", "application/strategic-merge-patch+json"
+            "application/merge-patch+json", "application/json-patch+json", "application/strategic-merge-patch+json"
         };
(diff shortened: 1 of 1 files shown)


hellowenqi:
CLA signed

brendandburns:
Unfortunately this code is generated, so we can't just do the change like this...

This is going to take some thought to get right. What is the use-case that you are trying for? That will help determine what the right thing to do is here...

Thanks!

hellowenqi:
```
final String[] localVarContentTypes = {
            "application/json-patch+json", "application/merge-patch+json", "application/strategic-merge-patch+json"
        };
        final String localVarContentType = apiClient.selectHeaderContentType(localVarContentTypes);
```
As the original code shows, function apiClient.selectHeaderContentType always select the first element of localVarContentTypes as content-type, which is "application/json-patch+json".
But when I call patchNode api using "application/json-patch+json", the server will return me error message "json: cannot unmarshal object into Go value of type jsonpatch.Patch".

you can reproduce it by the following command.

```
 curl -k -v -H "Accept: application/json" -XPATCH -d '{"metadata":{"labels":{"label-key":"label-value"}}}' -H "Content-Type: application/json-patch+json" http://localhost:8888/api/v1/nodes/minikube
```

brendandburns:
Yeah, I understand, but this is generated code, so I think what we need to do instead of this is to add a `PatchClient` or somesuch that extends from `APIClient` and overrides the `selectHeaderContentType` method in some way.

We can't change this file or else we won't be able to do code generation in the future.

hellowenqi:
I am very curious about the technology of generating code.  what is that?

Issue #343, opened 2018-08-21
Quantity.java is missing an equals method


saskiawagenaar:
kubernetes/src/main/java/io/kubernetes/client/custom/Quantity.java has no implemented equals method, which makes it difficult to compare deployments which have resource limits or requests.

brendandburns:
Closed by #344

PR #344, opened 2018-08-21
Implemented equals method for Quantity

saskiawagenaar:
Solves #343.


brendandburns:
LGTM, thanks!

Hendrik-H:
Hi, when implementing `equals` one should also implement `hashCode`. For example like this:
```
@Override
public int hashCode() {
    return Objects.hash(number, format);
}
```

PR #355, opened 2018-09-01
Added more example with Java SE 8 Lambda expression

yoshioterada:
I wrote the more example for the k8s Java Client with Java SE 8 Lambda expression.

This example will execute and show following in one example.

1.  scale the number of pod
2.  show all namespaces and pods in specific namespace
3.  show Services
4.  show logs for specific pod

Please refer to the following execution logs in my environment?
[log.txt](https://github.com/kubernetes-client/java/files/2342448/log.txt)

And I will add the diff as follows.
[diff.txt](https://github.com/kubernetes-client/java/files/2344049/diff.txt)

Thank you & Best Regards.
Yoshio.

### DIFF
--- a/examples/pom.xml
+++ b/examples/pom.xml
@@ -77,6 +77,19 @@
 					</execution>
 				</executions>
 			</plugin>
+			<plugin>
+				<groupId>com.coveo</groupId>
+				<artifactId>fmt-maven-plugin</artifactId>
+				<version>2.2.0</version>
+				<executions>
+					<execution>
+						<phase>test</phase>
+						<goals>
+							<goal>check</goal>
+						</goals>
+					</execution>
+				</executions>
+			</plugin>
 		</plugins>
 	</build>
 	<properties>
--- a/examples/pom.xml
+++ b/examples/pom.xml
@@ -77,23 +77,10 @@
 					</execution>
 				</executions>
 			</plugin>
-			<plugin>
-				<groupId>com.coveo</groupId>
-				<artifactId>fmt-maven-plugin</artifactId>
-				<version>2.2.0</version>
-				<executions>
-					<execution>
-						<phase>test</phase>
-						<goals>
-							<goal>check</goal>
-						</goals>
-					</execution>
-				</executions>
-			</plugin>
 		</plugins>
 	</build>
 	<properties>
-		<java.version>1.7</java.version>
+		<java.version>1.8</java.version>
 		<maven.compiler.source>${java.version}</maven.compiler.source>
 		<maven.compiler.target>${java.version}</maven.compiler.target>
 		<swagger-core-version>1.5.12</swagger-core-version>
--- a/examples/src/main/java/io/kubernetes/client/examples/Example.java
+++ b/examples/src/main/java/io/kubernetes/client/examples/Example.java
@@ -24,6 +24,7 @@ import io.kubernetes.client.models.V1NamespaceList;
 import io.kubernetes.client.models.V1PodList;
 import io.kubernetes.client.models.V1ServiceList;
 import io.kubernetes.client.util.Config;
+import java.io.IOException;
 import java.util.List;
 import java.util.Optional;
 import java.util.logging.Level;
@@ -49,21 +50,24 @@ public class Example {
     For API_SERVER_NAME, you can get the server name as follows.
     $ kubectl cluster-info|grep master
     Kubernetes master is running at https://*****************.hcp.japaneast.azmk8s.io:443
-    */
+     */
     private final static String API_SERVER_NAME = "https://*****************.hcp.japaneast.azmk8s.io";
     /*
     For ACCESS_TOKEN , you can get the token as follows
     $ kubectl describe secret $(kubectl get secrets | grep default | cut -f1 -d ' ') | grep -E '^token' | cut -f2 -d':' | tr -d '\t'
-    */
+     */
     private static final String ACCESS_TOKEN = "********************************";
 
     private final static Logger LOGGER = Logger.getLogger(Example.class.getName());
 
     /**
      * Constructor
+     * @throws java.io.IOException
      */
-    public Example() {
-        ApiClient client = Config.fromToken(API_SERVER_NAME, ACCESS_TOKEN, false);
+    public Example() throws IOException {
+        ApiClient client = Config.defaultClient();
+        //If you want to use specific k8s cluster and access token, please use following?
+        //ApiClient client = Config.fromToken(API_SERVER_NAME, ACCESS_TOKEN, false);
         Configuration.setDefaultApiClient(client);
         corev1Api = new CoreV1Api(client);
     }
@@ -77,8 +81,8 @@ public class Example {
         try {
             Example operation = new Example();
             operation.executeCommand();
-        } catch (ApiException ex) {
-            LOGGER.log(Level.SEVERE, null, ex);
+        } catch (ApiException | IOException ex) {
+            Logger.getLogger(Example.class.getName()).log(Level.SEVERE, null, ex);
         }
     }
 
@@ -201,7 +205,7 @@ public class Example {
     public void scaleDeployment(String deploymentName, int numberOfReplicas) throws ApiException {
         ExtensionsV1beta1Api extensionV1Api = new ExtensionsV1beta1Api();
         extensionV1Api.setApiClient(corev1Api.getApiClient());
-        ExtensionsV1beta1DeploymentList listNamespacedDeployment = extensionV1Api.listNamespacedDeployment(DEFAULT_NAME_SPACE,null,null,null,Boolean.FALSE,null,null,null,null,Boolean.FALSE);
+        ExtensionsV1beta1DeploymentList listNamespacedDeployment = extensionV1Api.listNamespacedDeployment(DEFAULT_NAME_SPACE, null, null, null, Boolean.FALSE, null, null, null, null, Boolean.FALSE);
 
         List<ExtensionsV1beta1Deployment> extensionsV1beta1DeploymentItems = listNamespacedDeployment.getItems();
         Optional<ExtensionsV1beta1Deployment> findedDeployment = extensionsV1beta1DeploymentItems.stream()
--- a/examples/src/main/java/io/kubernetes/client/examples/Example.java
+++ b/examples/src/main/java/io/kubernetes/client/examples/Example.java
@@ -45,6 +45,7 @@ public class Example {
 
     private final CoreV1Api corev1Api;
     private final static String DEFAULT_NAME_SPACE = "default";
+    private final static Integer TIME_OUT_VALUE = 180;
 
     /*
     For API_SERVER_NAME, you can get the server name as follows.
@@ -57,7 +58,7 @@ public class Example {
     $ kubectl describe secret $(kubectl get secrets | grep default | cut -f1 -d ' ') | grep -E '^token' | cut -f2 -d':' | tr -d '\t'
      */
     private static final String ACCESS_TOKEN = "********************************";
-
+    
     private final static Logger LOGGER = Logger.getLogger(Example.class.getName());
 
     /**
@@ -123,7 +124,7 @@ public class Example {
      * @throws ApiException
      */
     public List<String> getAllNameSapces() throws ApiException {
-        V1NamespaceList listNamespace = corev1Api.listNamespace("true", null, null, Boolean.FALSE, null, Integer.SIZE, null, Integer.BYTES, Boolean.FALSE);
+        V1NamespaceList listNamespace = corev1Api.listNamespace("true", null, null, Boolean.FALSE, null, 0, null, Integer.MAX_VALUE, Boolean.FALSE);
         List<String> list = listNamespace.getItems()
                 .stream()
                 .map(v1Namespace -> v1Namespace.getMetadata().getName())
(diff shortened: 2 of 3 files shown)


examples/pom.xml
```diff
@@ -77,23 +77,10 @@
 					</execution>
 				</executions>
 			</plugin>
-			<plugin>
```
brendandburns: Let's keep this plugin, I want to enforce code style.

examples/src/main/java/io/kubernetes/client/examples/Example.java
```diff
@@ -9,36 +9,230 @@
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
-*/
+ */
```
brendandburns: please delete added whitespace.

examples/src/main/java/io/kubernetes/client/examples/Example.java
```diff
@@ -9,36 +9,230 @@
...
-    CoreV1Api api = new CoreV1Api();
-    V1PodList list =
-        api.listPodForAllNamespaces(null, null, null, null, null, null, null, null, null);
-    for (V1Pod item : list.getItems()) {
-      System.out.println(item.getMetadata().getName());
+    private final CoreV1Api corev1Api;
+    private final static String DEFAULT_NAME_SPACE = "default";
+    private final static Integer TIME_OUT_VALUE = 180;
+
+    /*
```
brendandburns: delete this, it should be loaded from the `$HOME/.kube/config` file

examples/src/main/java/io/kubernetes/client/examples/Example.java
```diff
@@ -9,36 +9,230 @@
...
+    For API_SERVER_NAME, you can get the server name as follows.
+    $ kubectl cluster-info|grep master
+    Kubernetes master is running at https://*****************.hcp.japaneast.azmk8s.io:443
+     */
+    private final static String API_SERVER_NAME = "https://*****************.hcp.japaneast.azmk8s.io";
```
brendandburns: delete this, as above, it should be loaded from the `$HOME/.kube/config` file.

examples/src/main/java/io/kubernetes/client/examples/Example.java
```diff
@@ -9,36 +9,230 @@
...
+    /*
+    For ACCESS_TOKEN , you can get the token as follows
+    $ kubectl describe secret $(kubectl get secrets | grep default | cut -f1 -d ' ') | grep -E '^token' | cut -f2 -d':' | tr -d '\t'
+     */
+    private static final String ACCESS_TOKEN = "********************************";
```
brendandburns: delete this one too for the same reason.

examples/src/main/java/io/kubernetes/client/examples/Example.java
```diff
@@ -9,36 +9,230 @@
...
+import io.kubernetes.client.models.ExtensionsV1beta1DeploymentList;
+import io.kubernetes.client.models.ExtensionsV1beta1DeploymentSpec;
+import io.kubernetes.client.models.V1NamespaceList;
 import io.kubernetes.client.models.V1PodList;
+import io.kubernetes.client.models.V1ServiceList;
 import io.kubernetes.client.util.Config;
 import java.io.IOException;
+import java.util.List;
+import java.util.Optional;
+import java.util.logging.Level;
```
brendandburns: We're using `sl4j` instead of the built in logger, see:

https://github.com/kubernetes-client/java/blob/master/util/src/main/java/io/kubernetes/client/util/credentials/ClientCertificateAuthentication.java#L17

examples/src/main/java/io/kubernetes/client/examples/Example.java
```diff
@@ -9,36 +9,230 @@
...
-  public static void main(String[] args) throws IOException, ApiException {
-    ApiClient client = Config.defaultClient();
-    Configuration.setDefaultApiClient(client);
 
-    CoreV1Api api = new CoreV1Api();
-    V1PodList list =
-        api.listPodForAllNamespaces(null, null, null, null, null, null, null, null, null);
-    for (V1Pod item : list.getItems()) {
-      System.out.println(item.getMetadata().getName());
+    private final CoreV1Api corev1Api;
```
brendandburns: I don't think this needs to be declared as a class level variable.

examples/src/main/java/io/kubernetes/client/examples/Example.java
```diff
@@ -9,36 +9,230 @@
...
+     */
+    private static final String ACCESS_TOKEN = "********************************";
+    
+    private final static Logger LOGGER = Logger.getLogger(Example.class.getName());
+
+    /**
+     * Constructor
+     * @throws java.io.IOException
+     */
+    public Example() throws IOException {
```
brendandburns: I don't think we need a constructor, let's keep it all in the static `main` method.

examples/src/main/java/io/kubernetes/client/examples/Example.java
```diff
@@ -9,36 +9,230 @@
...
+        ApiClient client = Config.defaultClient();
+        //If you want to use specific k8s cluster and access token, please use following?
```
brendandburns: Delete this, I don't think it's necessary.

examples/src/main/java/io/kubernetes/client/examples/Example.java
```diff
@@ -9,36 +9,230 @@
...
+
+    /**
+     * Main method
+     *
+     * @param args
+     */
+    public static void main(String[] args) {
+        try {
+            Example operation = new Example();
+            operation.executeCommand();
```
brendandburns: let's just keep everything `static`, I don't think we need an `Example` object.

examples/src/main/java/io/kubernetes/client/examples/Example.java
```diff
@@ -9,36 +9,230 @@
...
+        } catch (ApiException | IOException ex) {
+            Logger.getLogger(Example.class.getName()).log(Level.SEVERE, null, ex);
+        }
+    }
+
+    public void executeCommand() throws ApiException {
+
```
brendandburns: delete newline.

examples/src/main/java/io/kubernetes/client/examples/Example.java
```diff
@@ -9,36 +9,230 @@
...
+        System.out.println("----- Print Log of Specific Pod End -----");
+    }
+
+    /**
+     * Get all namespaces in k8s cluster
+     *
+     * @return
+     * @throws ApiException
+     */
+    public List<String> getAllNameSapces() throws ApiException {
```
brendandburns: typo, this should be `getAllNamespaces`

examples/src/main/java/io/kubernetes/client/examples/Example.java
```diff
@@ -9,36 +9,230 @@
...
+
+        //Print all of the Services
+        System.out.println("----- Print list all Services Start -----");
+        List<String> services = getServices();
+        services.stream().forEach(System.out::println);
+        System.out.println("----- Print list all Services End -----");
+
+        //Print log of specific pod
+        System.out.println("----- Print Log of Specific Pod Start -----");
+        printLog(DEFAULT_NAME_SPACE, "account-service-cbd44cc8-tq5hb");
```
brendandburns: Since this pod won't exist, maybe use `getPods()[0]`

brendandburns:
Thanks for the PR!

I think that this is a great larger example. I'd prefer to retain `Example.java` as a really simple example for people to learn from. Perhaps you can add this as a new `ExpandedExample.java`?

Otherwise, I think I'd prefer just a bunch of `static` methods rather than instance level methods?

examples/pom.xml
yoshioterada: Brendan-san, I understood about it !!

examples/src/main/java/io/kubernetes/client/examples/Example.java
yoshioterada: I will delete the whitespace.

examples/src/main/java/io/kubernetes/client/examples/Example.java
yoshioterada: In my environment, old "config" exists and refer to the old config. As a result , it tried to access difference AKS Cluster. So I would like to show that there is a difference way to refer the cluster. So at this point, I would like to insert this code.

examples/src/main/java/io/kubernetes/client/examples/Example.java
yoshioterada: In my environment, old "config" exists and refer to the old config. As a result , it tried to access difference AKS Cluster. So I would like to show that there is a difference way to refer the cluster. So at this point, I would like to insert this code.

examples/src/main/java/io/kubernetes/client/examples/Example.java
yoshioterada: OK, I will modify to use the sl4j.

examples/src/main/java/io/kubernetes/client/examples/Example.java
yoshioterada: In the Constructor of Example() class, I initialized only one times for CoreV1Api instance as follows.
corev1Api = new CoreV1Api(client);

After that I reuse the instance valuable on following.
****************************************************************************
        V1NamespaceList listNamespace = corev1Api.listNamespace("true", null, null, Boolean.FALSE, null, 0, null, Integer.MAX_VALUE, Boolean.FALSE);

        V1PodList v1podList = corev1Api.listPodForAllNamespaces(null, null, null, null, null, null, null, null, null);

        V1PodList listNamespacedPod = corev1Api.listNamespacedPod(namespace, null, null, null, Boolean.FALSE, label, Integer.MAX_VALUE, null, TIME_OUT_VALUE, Boolean.FALSE);

        V1ServiceList listNamespacedService = corev1Api.listNamespacedService(DEFAULT_NAME_SPACE, null, null, null, Boolean.FALSE, null, Integer.MAX_VALUE, null, TIME_OUT_VALUE, Boolean.FALSE);

        extensionV1Api.setApiClient(corev1Api.getApiClient());
****************************************************************************

It may be OK to specify the "ApiClient client" instead of the "CoreV1Api".

However I thought that "CoreV1Api" is better.

examples/src/main/java/io/kubernetes/client/examples/Example.java
yoshioterada: This is related the above code, I thought that if I don't implement the Constructor, I need to write the creation code for the instance of both "ApiClient" and "CoreV1Api".
I thought that it is not appropriate for Natural Java programing. So I created the constructor.

        //If you want to use specific k8s cluster and access token, please use following?
        //ApiClient client = Config.fromToken(API_SERVER_NAME, ACCESS_TOKEN, false);
        corev1Api = new CoreV1Api(client);

examples/src/main/java/io/kubernetes/client/examples/Example.java
yoshioterada: Yes, I will modify it !! Thank you so much !!

examples/src/main/java/io/kubernetes/client/examples/Example.java
yoshioterada: Yes, I will modify it!! Thank you so much !!

examples/src/main/java/io/kubernetes/client/examples/Example.java
yoshioterada: I'm thinking that "everything static method" is not appropriate for Java programing language manner as Object Oriented programing. I will be able to include the code of executeCommand() into main() method. However I need to create a new Example Object to invoke another method.

examples/src/main/java/io/kubernetes/client/examples/Example.java
yoshioterada: If the above is OK, I will copy the code to main() method.

yoshioterada:
Dear Brendan-san, Based on your suggestion and I will modify some point like type and, I would like to create new file as ExpandedExample.java

examples/src/main/java/io/kubernetes/client/examples/Example.java
yoshioterada: In my environment, old "config" exists and refer to the old config. As a result , it tried to access difference AKS Cluster. So I would like to show that there is a difference way to refer the cluster. So at this point, I would like to insert this code.

yoshioterada:
Dear Brendan-san, I created new Java Class and modified the previous discussion. So if it is possible, could you review again?

brendandburns:
A few small comments, generally this looks good, I really would prefer that all methods are static though as I don't think object-oriented programming really applies here...

yoshioterada:
Dear Brendan-san.
I added the static modifiers for all of methods. 

Thank you & Best Regards.
Yoshio.

brendandburns:
Sorry a few more comments, then I think it's ready to merge.

Thanks for the patience!

examples/src/main/java/io/kubernetes/client/examples/ExpandedExample.java
```diff
@@ -0,0 +1,301 @@
+/*
+Copyright 2017 The Kubernetes Authors.
```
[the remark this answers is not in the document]
yoshioterada: Yes, I will modify it.

brendandburns:
/lgtm
/approve

Thanks for the patience!

PR #366, opened 2018-09-12
Dropped reference to gson internal API

ieugen:
* fixes https://github.com/kubernetes-client/java/issues/304
* JSON class is ignored by swagger code gen - see kubernetes/.swagger-codegen-ignore
* please document process around swagger code-gen so people can contribute
* please separate generated code from developed code


brendandburns:
Here's the improved instructions:

https://github.com/kubernetes-client/java/pull/371

PR #373, opened 2018-09-20
shutdown okhttp client after exec call

kondapally1989:
After exec call there is an idle time because connection pool still persists for some time as it is a async call. 
so it would be better to close connection pool immediately when ExecProcess.destroy() is called.


kondapally1989:
it seems after closing executor services, we need to reinitialise ApiClient if we want to again hit Kubernetes api which is unwanted behaviour. 

I will close this pull request and look into it again.

PR #375, opened 2018-09-27
Add utilities for copying files.

### LINKED ISSUE #353: Transferring files to pod
Hello,
I tried to send files to a pod with the lib using the Exec object. It works, but it can't ensure the integrity of the file.
Here is an example of how I do it : 

```
public class ExecExample {
    public static void main(String[] args) throws IOException, ApiException {
        String podName = "nginx-4217019353-k5sn9";
        String namespace = "default";

        Exec exec = new Exec();
        final Process proc =
                exec.exec(
                        namespace,
                        podName,
                        new String[]{"tee" ,"/tmp/test"},
                        true,
                        false);

        try {
            FileInputStream file = new FileInputStream("C:/test");
            OutputStream pod = proc.getOutputStream();
            IOUtils.copy(file, pod);
            pod.flush();
            IOUtils.closeQuietly(file);
            IOUtils.closeQuietly(pod);
        } catch (IOException ex) {
            ex.printStackTrace();
        }

        proc.destroy();

        System.exit(proc.exitValue());
    }
}
```

The problem is that there are missing bytes in my pod file. Not always the same amount and it's depending of the original file size. 
For example : 
- Sent 1024 bytes -> always received 1024 bytes = OK
- Sent 2048 bytes -> always received 2047 bytes = KO
- Sent 4096 bytes -> always received 4095 bytes = KO
- Sent 8192 bytes -> always received 8190 bytes = KO

I tested with the command `tar xf - -C /tmp` and sending a tar file, and I also tested the commands with kubectl. No problems in the both cases.
It's maybe a buffer problem, but I don't know when. IOUtils.copy() return the right amount of bytes.

Development environment : Windows
Kubernetes Java library version : 2.0.0
Kubernetes version : 1.8.2

brendandburns:
Closes #353

### DIFF
--- a/util/src/main/java/io/kubernetes/client/Copy.java
+++ b/util/src/main/java/io/kubernetes/client/Copy.java
@@ -1,0 +1,89 @@
+/*
+Copyright 2017, 2018 The Kubernetes Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package io.kubernetes.client;
+
+import com.google.common.io.ByteStreams;
+import io.kubernetes.client.models.V1Pod;
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Path;
+import org.apache.commons.codec.binary.Base64;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class Copy extends Exec {
+  private static final Logger log = LoggerFactory.getLogger(Copy.class);
+
+  /** Simple Copy constructor, uses default configuration */
+  public Copy() {
+    super(Configuration.getDefaultApiClient());
+  }
+
+  /**
+   * Copy Constructor
+   *
+   * @param apiClient The api client to use.
+   */
+  public Copy(ApiClient apiClient) {
+    super(apiClient);
+  }
+
+  public InputStream copyFile(String namespace, String pod, String srcPath)
+      throws ApiException, IOException {
+    return copyFile(namespace, pod, null, srcPath);
+  }
+
+  public InputStream copyFile(V1Pod pod, String srcPath) throws ApiException, IOException {
+    return copyFile(pod, null, srcPath);
+  }
+
+  public InputStream copyFile(V1Pod pod, String container, String srcPath)
+      throws ApiException, IOException {
+    return copyFile(
+        pod.getMetadata().getNamespace(), pod.getMetadata().getName(), container, srcPath);
+  }
+
+  public InputStream copyFile(String namespace, String pod, String container, String srcPath)
+      throws ApiException, IOException {
+    Process proc =
+        this.exec(
+            namespace,
+            pod,
+            new String[] {"sh", "-c", "cat " + srcPath + " | base64"},
+            container,
+            false,
+            false);
+    try {
+      proc.waitFor();
+    } catch (InterruptedException ex) {
+      log.warn("Interrupted waiting for copy to complete", ex);
+    }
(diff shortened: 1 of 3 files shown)


kondapally1989:
@brendandburns 
looks good to me.
may be we can add todo statements for
1. is tar available.
2. and can function names be like copyFileFrom/copyDirectoryFrom so that in future we can also add some thing like copyFileTo/copyDirectoryTo  ?

adding  copyFileTo  Pod  may not  be appropriate as it defeats the purpose of docker / kubernetes. but may be for testing people can use it.

brendandburns:
Comments addressed, please re-check.

Thanks!

util/src/main/java/io/kubernetes/client/Copy.java
```diff
@@ -0,0 +1,152 @@
...
+          }
+        }
+      }
+    }
+  }
+
+  public static void copyFile(String namespace, String pod, String srcPath, Path dest)
+      throws ApiException, IOException {
+    Copy c = new Copy();
+    InputStream is = c.copyFile(namespace, pod, null, srcPath);
```
kondapally1989: i think travis build is throwing error here , string to Path

util/src/main/java/io/kubernetes/client/Copy.java
```diff
@@ -0,0 +1,152 @@
...
+   *
+   * @param apiClient The api client to use.
+   */
+  public Copy(ApiClient apiClient) {
+    super(apiClient);
+  }
+
+  public InputStream copyFileFromPod(String namespace, String pod, String srcPath)
+      throws ApiException, IOException {
+    return copyFile(namespace, pod, null, srcPath);
```
kondapally1989: is it copyFileFromPod ?

util/src/main/java/io/kubernetes/client/Copy.java
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  public InputStream copyFileFromPod(V1Pod pod, String srcPath) throws ApiException, IOException {
+    return copyFile(pod, null, srcPath);
```
kondapally1989: is it copyFileFromPod ?

util/src/main/java/io/kubernetes/client/Copy.java
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  public InputStream copyFileFromPod(V1Pod pod, String container, String srcPath)
+      throws ApiException, IOException {
+    return copyFile(
```
kondapally1989: is it copyFileFromPod ?

util/src/main/java/io/kubernetes/client/Copy.java
```diff
@@ -0,0 +1,152 @@
...
+  public void copyFileFromPod(String namespace, String name, String srcPath, Path destination)
+      throws ApiException, IOException {
+    try (InputStream is = copyFileFromPod(namespace, name, srcPath);
+         FileOutputStream fos = new FileOutputStream(destination.toFile())) {
+      ByteStreams.copy(is, fos);
+      fos.flush();
+    }
+  }
+
+  public void copyDirectoryFromPod(V1Pod pod, String srcPath, Path destination)
```
kondapally1989: is it copyDirectoryToPod ?

util/src/main/java/io/kubernetes/client/Copy.java
```diff
@@ -0,0 +1,152 @@
...
+
+  public void copyDirectoryFromPod(V1Pod pod, String container, String srcPath, Path destination)
+      throws ApiException, IOException {
+    copyDirectoryToPod(pod.getMetadata().getNamespace(), pod.getMetadata().getName(), container, srcPath, destiation);
+  }
+
+
+  public void copyDirectoryFromPod(String namespace, String pod, String srcPath, Path destination)
+      throws ApiException, IOException {
+    copyDirectoryToPod(namespace, pod, null, srcPath, destiation);
```
kondapally1989: destination

util/src/main/java/io/kubernetes/client/Copy.java
kondapally1989: copyFileFromPod?

kondapally1989:
/hold cancel
i did /lgtm in one of previous comment but deleted the comment as ci build broke and added hold.

@brendandburns  once you fix those compile issues, i think it will automatically merge

brendandburns:
@kondapally1989 updated, thanks for the patience!

kondapally1989:
@brendandburns  format issue in compile time.

brendandburns:
@kondapally1989 sorry for the noise, I swear I fixed the formatting, but apparently not.  Please take another look.

PR #379, opened 2018-10-04
Update the versions used in the README to latest stable and beta

dadadom:
The versions in the README.md which is shown as the top-level document when looking at the repository, were outdated. Users might use those outdated versions and run into problems which have been fixed in the meantime.


dadadom:
CLA signed.

brendandburns:
Thanks!

/lgtm
/approve

PR #387, opened 2018-10-10
Changed to set USERPROFILE on Windows environment.

### LINKED ISSUE #386: Can't load $HOME/.Kube/config when running on Windows
`io.kubernetes.client.util.KubeConfig#ENV_HOME` is fixed at "HOME", but $HOME is not set under Windows environment.
Therefore `io.kubernetes.client.util.ClientBuilder#findConfigInHomeDir()` fails and can not read config file.

In "client-go" it corresponds to this problem the following code
https://github.com/kubernetes/client-go/blob/master/util/homedir/homedir.go

but it does not seem to consider "client-java".

yu-shiba:
fix #386


yu-shiba:
I signed it.

brendandburns:
/lgtm
/approve

Thanks!

PR #390, opened 2018-10-11
Expand HOME detection on Windows.

brendandburns:
Further fixes for #386 

This emulates the kubeconfig code found here:

https://github.com/kubernetes/client-go/blob/master/util/homedir/homedir.go

@yu-shiba


yu-shiba:
I also thought about the same thing.

I adopted the #387 approach because of emphasis on compatibility with existing assets,

1. `io.kubernetes.client.util.KubeConfig#ENV_HOME` is a public member and can not make drastic changes
2. The same fault occurs when referencing `io.kubernetes.client.util.KubeConfig#ENV_HOME` with other code
3. Respond with minimal change

Is there a problem with the above?

brendandburns:
I don't think anyone is using `ENV_HOME` and I prefer to behave the same way as kubectl/client-go

So I think I prefer this fix...

Thanks!

yu-shiba:
I agree with you.
#387 also wanted to do like this #390.

The code of #387 is not beautiful though what I say it myself...

brendandburns:
@yu-shiba if you have time, please take a look and review this code.

If it looks good, you can reply /lgtm

And then this code can merge.

Thanks!

yu-shiba:
@brendandburns Is it not good?
Feeling insufficient authority.

brendandburns:
self-lgtm on lgtm from @yu-shiba

Issue #86, opened 2017-10-04
Getting exception from deleteNamespacedStatefulSet() call


chlung:
Hi,

I was getting exception when I issued deleteNamespacedStatefulSet() call.  The call did delete statefulset, but it throw the exception back to caller.

Exception:
java.lang.IllegalStateException: Expected a string but was BEGIN_OBJECT at line 1 column 2004 path $.status

Here is my code,
V1DeleteOptions deleteOptions = new V1DeleteOptions()
Boolean orphanDependents = true

apiInstance.deleteNamespacedStatefulSet(name, namespace, deleteOptions, null, 60, orphanDependents, null)

Thanks.

brendandburns:
I'll investigate this, I suspect that it is a problem with the Swagger/OpenAPI spec indicating the wrong code for Delete...

But we'll need to investigate to be sure.

chlung:
@brendanburns thanks.

brendandburns:
Ok, I figured out the issue here...

When you delete a StatefulSet (and probably a ReplicaSet too) it initially returns a `StatefulSet` object, but eventually it returns a `Status` object.

This means that the Swagger should be `oneOf(StatefulSet, Status)`

But the Kubernetes Swagger here:

https://raw.githubusercontent.com/kubernetes/kubernetes/master/api/openapi-spec/swagger.json

Only has the `Status` object.

Thus, the generated Java code barfs when it tries to parse a `StatefulSet` object into a `Status` object.

Unfortunately `go-restful` when generates the swagger spec doesn't support `oneOf` as far as I know, thus we're going to either need to massage the swagger spec before client generation, improve go-restful's swagger generation or switch to a different way of generating swagger...

None of these are great options, and none of them will happen very quickly.  So for now, I think the best you can do is catch this Exception and move on...

Sorry!

@mbohlool @lwander @lavalamp fyi

chlung:
@brendandburns  ok.  I can handle it in my code.  Thanks for debugging this issue.  Should we keep this issue for tracking purpose or close it for now?

brendandburns:
Yes, let's definitely keep this open.

mbohlool:
`oneof` is not supported in OpeAPI v2. It is possible to fix generator when
we move to OpenAPI v3 (both our kube-openapi and swagger-codegen should
support it).

neuromantik33:
Bug also exists when trying to delete a namespaced Job. JSON output is here
```json
{
  "kind": "Job",
[...]

PR #406, opened 2018-10-28
Invoke async callback on unexpected exceptions

mpartel:
Currently e.g. #86 causes an async request's callback to never be called, because deserialization fails with an exception that is not an `ApiException`.

- The first commit ensures the callback is called for any exception.
- The second commit wraps JSON exceptions in an `ApiException`, as the method's javadoc specifies.

### DIFF
--- a/kubernetes/src/main/java/io/kubernetes/client/ApiClient.java
+++ b/kubernetes/src/main/java/io/kubernetes/client/ApiClient.java
@@ -831,12 +831,15 @@ public class ApiClient {
             }
 
             @Override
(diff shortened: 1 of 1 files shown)


kondapally1989:
@mpartel 
ApiClient is auto generated by swagger. and should not be edited manually.

mpartel:
Ah, so it is :(

Ok, I'll try to look for the original.

mpartel:
https://github.com/swagger-api/swagger-codegen/pull/8870

hellowenqi:
Are there any solutions to handle the exception for an async request up till now? #86

PR #412, opened 2018-11-02
Add merge in ProtoClient

TwinProduction:
`ProtoClient.update(...)` uses PUT, but really, it replaces the existing resource with the given name by the resource passed as parameter `obj`. It shouldn't have been named `update`, but instead, it should have been named `replace` for clarity. 

[reference](https://kubernetes.io/docs/reference/federation/v1/operations/#_replace_the_specified_service)

I didn't rename it regardless, because it would be a breaking change, but I did add the `merge(...)` method which uses `PATCH`, as I think it would be better to avoid using the `request(...)` method as much as possible for clarity.


TwinProduction:
This closes  #413

brendandburns:
Change looks good to me, but it needs formatting fixed.

To fix formatting errors, run "mvn fmt:format"

TwinProduction:
@brendandburns Fixed

PR #414, opened 2018-11-03
Fix formatting in Quantity.java

davidxia:
Indent two spaces instead of four to be consistent with the rest
of this project.


davidxia:
Use https://github.com/kubernetes-client/java/pull/414/files?w=1 to see this diff without whitespace changes.

PR #415, opened 2018-11-03
Remove extra `Quantity.fromString()` call

davidxia:
in Quantity constructor. Avoid doing extra work.
We only need to call it once.

### DIFF
--- a/kubernetes/src/main/java/io/kubernetes/client/custom/Quantity.java
+++ b/kubernetes/src/main/java/io/kubernetes/client/custom/Quantity.java
@@ -37,8 +37,9 @@ public class Quantity {
   }
 
   public Quantity(final String value) {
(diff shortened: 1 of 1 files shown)


kubernetes/src/main/java/io/kubernetes/client/custom/Quantity.java
```diff
@@ -37,8 +37,9 @@ public Quantity(final BigDecimal number, final Format format) {
     }
 
     public Quantity(final String value) {
```
davidxia: Aside from this PR, it seems weird to me that this constructor calls a static factory method, gets an instance of Quantity, and reassigns the attributes to itself.

brendandburns:
Needs rebase now.

davidxia:
Thanks. Rebased.

PR #416, opened 2018-11-03
Use two space indentation in some test classes

davidxia:
* JSONTest
* SuffixFormatterTest
* V1PodBuilderTest

Indent two spaces instead of four to be consistent with the rest
of this project.

### DIFF
--- a/kubernetes/src/test/java/io/kubernetes/client/JSONTest.java
+++ b/kubernetes/src/test/java/io/kubernetes/client/JSONTest.java
@@ -8,20 +8,20 @@ import static org.junit.Assert.*;
 
 public class JSONTest {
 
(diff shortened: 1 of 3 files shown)


davidxia:
https://github.com/kubernetes-client/java/pull/416/files?w=1 shows nothing other than whitespace changes.

davidxia:
Related to https://github.com/kubernetes-client/java/issues/414. Let me know if we should use checkstyle for this project and what to check for. In any case this PR should help.

brendandburns:
@davidxia checkstyle should already be running, the build should fail if mvn format isn't correct.

Not sure why these files pass...

/lgtm
/approve

PR #417, opened 2018-11-03
Fix Quantity serialization

davidxia:
Fix failing `YamlTest.testLoadAllFile()`

by adding a `CustomRepresenter` for `Quantity` to `Yaml`.

The serialized Deployment YAML's resource requests won't necessarily be in the
same format as the original. But this PR prevents the serialization from
returning empty maps which is an improvement.

In the future, another PR can be made which could guarantee the original YAML
is returned when serializing Deployment back to YAML.

### DIFF
--- a/util/src/test/resources/expected.yaml
+++ b/util/src/test/resources/expected.yaml
@@ -1,0 +1,49 @@
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: mock
+  name: mock
+spec:
+  ports:
+  - port: 99
+    protocol: TCP
+    targetPort: 9949
+  selector:
+    app: mock
+---
+apiVersion: apps/v1beta1
+kind: Deployment
+metadata:
+  labels:
+    app: helloworld
+  name: helloworld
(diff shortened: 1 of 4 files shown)


brendandburns:
This is failing formatting check:

You need to run:

```
mvn fmt:format
``

**Comment by davidxia** [comment]
@brendanburns thanks. Fixed the formatting.

File: util/src/test/resources/expected.yaml [hunk 1]
```diff
@@ -0,0 +1,49 @@
...
+        image: gcr.io/hightowerlabs/helloworld:0.0.1
+        imagePullPolicy: Always
+        name: helloworld
+        resources:
+          limits:
+            memory: 1Gi
+            cpu: 500m
+          requests:
+            memory: 4Gi
+            cpu: '2'
```
davidxia: Not sure why the quotes are here. I guess it's because `QuantityFormatter.toBase10String()` returns a String?

brendandburns:
THanks for the PR!

PR #422, opened 2018-11-06
[MINOR] - Remove some redundant fields

dguy:
Removing some redundant fields i found when browsing the code


brendandburns:
@dguy 

Thanks for the PR, but these files are mostly auto-generated.

If you want to do these clean-ups they need to land in the [swagger-codegen](https://github.com/swagger-api/swagger-codegen) repo, and then we'll regenerate the code.

PR #461, opened 2018-12-10
Java informer portal implementation

yue9944882:
Gopher testing

Plz ignore any funny part in this pull if u notice 🙂Nvmd, i will revert later

Also, am opening a consumer part of this pull at k/gen repo

/hold


examples/src/main/java/io/kubernetes/client/examples/InformerExample.java
```diff
@@ -0,0 +1,36 @@
...
+
+        SharedInformerFactory.registerClient(CoreV1Api.class);
+
+        ApiClient client = Config.defaultClient();
+        Configuration.setDefaultApiClient(client);
+
+        SharedInformerFactory factory = new SharedInformerFactory();
+
+        // Node informer
+        ClusterScopeInformer<V1Node> nodeInformer = factory.informerForClusterScopeResource(V1Node.class);
```
yue9944882: example stands out here  @mbohlool @brendandburns

brendandburns:
Thanks for starting this PR!

Can I get a quick doc added that gives a high-level guide to the approach here?

It doesn't seem to me like forking the http-gson code is sustainable (or necessary) but I'm probably not fully thinking about the problem correctly.

yue9944882:
@brendandburns sure, am all ears. where the link? 👍🏻

brendandburns:
@yue9944882 Here's the rough sketch:

https://github.com/kubernetes-client/java/pull/468

It's an implementation of a ListWatch cache, but it would be easy to add event handlers to implement a generic Informer....

util/pom.xml
```diff
@@ -139,8 +139,8 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
                 <configuration>
-                    <source>1.7</source>
-                    <target>1.7</target>
+                    <source>8</source>
```
yue9944882: lambda is supported since java8

yue9944882:
@brendandburns here's a brief note about our informer

[https://docs.google.com/document/d/1HFOcs39oTD3g-cBuWyLd5QaevPg6omCBWnCbPV5JvBg/edit](https://docs.google.com/document/d/1HFOcs39oTD3g-cBuWyLd5QaevPg6omCBWnCbPV5JvBg/edit)

yue9944882:
politely bumping this thread @brendandburns

brendandburns:
Sorry, didn't realize this PR was still active, I will review ASAP...

examples/src/main/java/io/kubernetes/client/examples/CacheExample.java
```diff
@@ -0,0 +1,3 @@
+package io.kubernetes.client.examples;
+
+public class CacheExample {}
```
brendandburns: ?? Delete this or fill it in...

util/src/main/java/io/kubernetes/client/informer/ResyncRunnable.java
```diff
@@ -0,0 +1,32 @@
+package io.kubernetes.client.informer;
+
+import io.kubernetes.client.informer.cache.Store;
+import io.kubernetes.client.informer.cache.function.ResyncFunc;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ResyncRunnable<ApiType> implements Runnable {
```
brendandburns: I'd love to see a JavaDoc for each of these classes explaining how they fit together for the benefit of future developers.

util/src/main/java/io/kubernetes/client/informer/SharedInformerFactory.java
```diff
@@ -0,0 +1,106 @@
...
+
+  private Map<Type, Future> startedInformers;
+
+  private ExecutorService informerExecutor;
+
+  /** Constructor. */
+  public SharedInformerFactory() {
+    informers = new ConcurrentHashMap<>();
+    startedInformers = new HashMap<>();
+    informerExecutor = Executors.newCachedThreadPool();
```
brendandburns: Probably want to add a constructor where the user can supply their own ThreadPool (for testing, or performance sensitive applications)

util/src/main/java/io/kubernetes/client/informer/SharedInformerFactory.java
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  // Start initializes all requested informers.
+  public synchronized void startAllRegisteredInformers() {
+    if (Collections.isEmptyMap(informers)) {
+      return;
+    }
+    informers.forEach(
+        (informerType, informer) -> {
+          Future future = startedInformers.get(informerType);
```
brendandburns: This is weird. I think you should do:

```java
if (!startedInformers.containsKey(informerType)) {
    startedInformers.put(informerType, informerExecutor.submit(informer::run));
}
```

util/src/main/java/io/kubernetes/client/informer/SharedInformerFactory.java
```diff
@@ -0,0 +1,106 @@
...
+        });
+  }
+
+  public synchronized void stopAllRegisteredInformers() {
+    if (Collections.isEmptyMap(informers)) {
+      return;
+    }
+    informers.forEach(
+        (informerType, informer) -> {
+          Future future = startedInformers.get(informerType);
```
brendandburns: Same here use `containsKey` Also, I think you want to use `HashMap.remove(...)` not set the type to null. Setting the type to null won't empty the hashmap.

util/src/main/java/io/kubernetes/client/informer/cache/Controller.java
```diff
@@ -0,0 +1,137 @@
...
+import io.kubernetes.client.informer.ListerWatcher;
+import io.kubernetes.client.informer.ResyncRunnable;
+import io.kubernetes.client.informer.cache.function.PopProcessFunc;
+import io.kubernetes.client.informer.cache.function.ResyncFunc;
+import io.kubernetes.client.util.common.Collections;
+import java.util.concurrent.*;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Cache is java a portable for k/client-go's informer-Controller */
```
brendandburns: This sentence doesn't make full sense...

util/src/main/java/io/kubernetes/client/informer/cache/DeltaFIFO.java
```diff
@@ -0,0 +1,505 @@
...
+  }
+
+  /**
+   * Delete items from the delta FIFO.
+   *
+   * @param obj the obj
+   */
+  @Override
+  public void delete(Object obj) {
+    String id = this.keyOf(obj);
```
brendandburns: shouldn't this be inside the lock?

util/src/main/java/io/kubernetes/client/informer/cache/DeltaFIFO.java
```diff
@@ -0,0 +1,505 @@
...
+    lock.writeLock().lock();
+    try {
+      this.populated = true;
+      if (this.knownObjects == null) {
+        boolean exists = this.items.containsKey(id);
```
brendandburns: inline this:

```java
if (this.items.containsKey(id)) {
...
}
```

util/src/main/java/io/kubernetes/client/informer/cache/DeltaFIFO.java
```diff
@@ -0,0 +1,505 @@
...
+        if (!exists) {
+          // Presumably, this was deleted when a relist happened.
+          // Don't provide a second report of the same deletion.
+          return;
+        }
+      } else {
+        // We only want to skip the "deletion" action if the object doesn't
+        // exist in knownObjects and it doesn't have corresponding item in items.
+        Object knownObj = this.knownObjects.getByKey(id);
+        boolean exists = this.items.containsKey(id);
```
brendandburns: Inline this too.

util/src/main/java/io/kubernetes/client/informer/cache/DeltaFIFO.java
```diff
@@ -0,0 +1,505 @@
...
+        if (knownObj == null && !exists) {
```
brendandburns: use `containsKey` instead of obj == null.

util/src/main/java/io/kubernetes/client/informer/cache/DeltaFIFO.java
```diff
@@ -0,0 +1,505 @@
...
+  public Deltas<Object> pop(PopProcessFunc<Deltas<Object>> func) throws InterruptedException {
+    lock.writeLock().lock();
+    try {
+      while (true) {
+        while (queue.isEmpty()) {
+          notEmpty.await();
+        }
+        // there should have data now
+        String id = this.queue.removeFirst();
+        boolean exist = this.items.containsKey(id);
```
brendandburns: Move this down to the conditional and inline.

util/src/main/java/io/kubernetes/client/informer/cache/ReflectorRunnable.java
```diff
@@ -0,0 +1,188 @@
...
+          break;
+      }
+      lastSyncResourceVersion = newResourceVersion;
+      if (log.isDebugEnabled()) {
+        log.debug("{}#Receiving resourceVersion {}", apiTypeClass, lastSyncResourceVersion);
+      }
+    }
+  }
+
+  // NOTE(yue9944882): This function tries to find latest resourceVersion from the HTTP_GONE
```
brendandburns: delete dead code.

util/src/main/java/io/kubernetes/client/informer/cache/Cache.java
```diff
@@ -0,0 +1,467 @@
...
+      }
+    } finally {
+      lock.unlock();
+    }
+  }
+
+  /** Resync. */
+  @Override
+  public void resync() {
+    // Do nothing by default
```
brendandburns: default

util/src/main/java/io/kubernetes/client/informer/cache/Cache.java
```diff
@@ -0,0 +1,467 @@
...
+  private Map<String, IndexFunc> indexers;
+
+  /** items stores object instances */
+  private Map<String, ApiType> items;
+
+  /** indices stores objects' keys by their indices */
+  private Map<String, Map<String, Set<String>>> indices = new HashMap<>();
+
+  /** lock provides thread-safety */
+  private ReentrantLock lock = new ReentrantLock();
```
brendandburns: Do you really need this? I would just use `synchronized`

util/src/main/java/io/kubernetes/client/informer/cache/Controller.java
```diff
@@ -0,0 +1,137 @@
...
+  private void processLoop() {
+    while (true) {
+      try {
+        Deltas deltas = this.queue.pop(this.processFunc);
+        if (Collections.isEmptyCollection(deltas.getDeltas())) {
+          return;
+        }
+      } catch (InterruptedException t) {
+        log.error("DefaultController#processLoop get interrupted {}", t.getMessage(), t);
+        return;
```
brendandburns: delete this line.

util/src/main/java/io/kubernetes/client/informer/cache/Controller.java
```diff
@@ -0,0 +1,137 @@
...
+      return "";
+    }
+    return reflector.getLastSyncResourceVersion();
+  }
+
+  /** processLoop drains the work queue. */
+  private void processLoop() {
+    while (true) {
+      try {
+        Deltas deltas = this.queue.pop(this.processFunc);
```
brendandburns: You don't actually do anything with `deltas`, remove declaration and inline.

util/src/main/java/io/kubernetes/client/informer/cache/DeltaFIFO.java
```diff
@@ -0,0 +1,505 @@
...
+import io.kubernetes.client.informer.exception.EmptyDeltasException;
+import io.kubernetes.client.util.common.Collections;
+import java.util.*;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+// DeltaFIFO is a java portable of k/client-go's DeltaFIFO
```
brendandburns: Isn't DeltaQueue a better name?

brendandburns:
Added a number of comments. 

Also, this needs much better unit test coverage.

Thanks!

examples/src/main/java/io/kubernetes/client/examples/CacheExample.java
yue9944882: removed

util/src/main/java/io/kubernetes/client/informer/ResyncRunnable.java
yue9944882: yeah and i updated other classes w/ more java docs. ptal

util/src/main/java/io/kubernetes/client/informer/SharedInformerFactory.java
yue9944882: added. also applied this comment to `SharedProcessor`.

util/src/main/java/io/kubernetes/client/informer/SharedInformerFactory.java
yue9944882: done

util/src/main/java/io/kubernetes/client/informer/SharedInformerFactory.java
yue9944882: nice catch 👍🏻done

util/src/main/java/io/kubernetes/client/informer/cache/Controller.java
yue9944882: updated w/ following content: 

>  Cache is a java port of k/client-go's ThreadSafeStore. It basically saves and indexes all the entries.

util/src/main/java/io/kubernetes/client/informer/cache/DeltaFIFO.java
yue9944882: [https://github.com/kubernetes/kubernetes/blob/cdbc9d8567c177b8c6505181f755440792b9ff06/staging/src/k8s.io/client-go/tools/cache/delta_fifo.go#L191-L197](https://github.com/kubernetes/kubernetes/blob/cdbc9d8567c177b8c6505181f755440792b9ff06/staging/src/k8s.io/client-go/tools/cache/delta_fifo.go#L191-L197)

ref its original go code from k/k. it's fine to be out of the lock i suppose?

util/src/main/java/io/kubernetes/client/informer/cache/DeltaFIFO.java
yue9944882: done

util/src/main/java/io/kubernetes/client/informer/cache/DeltaFIFO.java
yue9944882: done

util/src/main/java/io/kubernetes/client/informer/cache/DeltaFIFO.java
yue9944882: check 👍🏻

util/src/main/java/io/kubernetes/client/informer/cache/ReflectorRunnable.java
yue9944882: deleted

util/src/main/java/io/kubernetes/client/informer/cache/Controller.java
yue9944882: deleted

util/src/main/java/io/kubernetes/client/informer/cache/Controller.java
yue9944882: make sense. updated

util/src/main/java/io/kubernetes/client/informer/cache/Cache.java
yue9944882: am afraid i didn't get it. what's ur point plz?

util/src/main/java/io/kubernetes/client/informer/cache/Cache.java
yue9944882: mmm good question. technically, it would be fine to replace the lock instance by `synchronized`. wdyt if we do it in a follow-up? it's safer to keep the code on a same basis w/ its go code implementation (additionally it's proved to be working in our environment) as its first landing to this repo. i added a TODO note here.

util/src/main/java/io/kubernetes/client/informer/cache/DeltaFIFO.java
yue9944882: both naming sounds fine to me. while `DeltaFIFO` keeps the naming from its go code which might be better for code maintenance. wdyt?

[https://github.com/kubernetes/kubernetes/blob/cdbc9d8567c177b8c6505181f755440792b9ff06/staging/src/k8s.io/client-go/tools/cache/delta_fifo.go#L96](https://github.com/kubernetes/kubernetes/blob/cdbc9d8567c177b8c6505181f755440792b9ff06/staging/src/k8s.io/client-go/tools/cache/delta_fifo.go#L96)

yue9944882:
@brendandburns thx for reviewing 😋all comments addressed, ptal

> Also, this needs much better unit test coverage.

+9999. more tests are on the way.

yue9944882:
@brendandburns test added. ptal. thx!

util/src/main/java/io/kubernetes/client/informer/cache/Deltas.java
```diff
@@ -0,0 +1,77 @@
+package io.kubernetes.client.informer.cache;
+
+import io.kubernetes.client.util.common.Collections;
+import java.util.List;
+
+/**
+ * Deltas class is the queue'd entry in the DeltaFIFO class. It consists of one or more Delta class
+ * which records the operation type and its state of the object.
+ */
+public class Deltas<ApiType> {
```
brendandburns: This is basically exactly a [Java `Deque`](https://docs.oracle.com/javase/7/docs/api/java/util/Deque.html)

Let's remove the `Deltas` class and just use `Deque`? (or at the very least use a `Deque` instead of a `List`)

util/src/main/java/io/kubernetes/client/informer/cache/Deltas.java
```diff
@@ -0,0 +1,77 @@
...
+    return deltas.get(0);
+  }
+
+  public static class Delta<ApiType> {
+
+    private DeltaType type;
+
+    private ApiType obj;
+
+    public Delta(DeltaType type, ApiType obj) {
```
brendandburns: Perhaps use [MutablePair<DeltaType, ApiType>](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/tuple/MutablePair.html) instead of a custom class? that way toString, hashCode, etc will all work correctly.

util/src/main/java/io/kubernetes/client/informer/cache/Deltas.java
```diff
@@ -0,0 +1,77 @@
...
+    public ApiType getObj() {
+      return obj;
+    }
+
+    public void setObj(ApiType obj) {
+      this.obj = obj;
+    }
+  }
+
+  public enum DeltaType {
```
brendandburns: This feels like a duplicate of EventType?

util/src/main/java/io/kubernetes/client/informer/cache/Store.java
```diff
@@ -0,0 +1,68 @@
+package io.kubernetes.client.informer.cache;
+
+import java.util.List;
+
+public interface Store<ApiType> {
```
brendandburns: Isn't this basically just a Map?

util/src/main/java/io/kubernetes/client/informer/cache/function/IndexFunc.java
```diff
@@ -0,0 +1,9 @@
+package io.kubernetes.client.informer.cache.function;
+
+import java.util.List;
+
+@FunctionalInterface
+public interface IndexFunc<T> {
```
brendandburns: Let's remove this and use `java.util.function` instead?

util/src/main/java/io/kubernetes/client/informer/cache/function/KeyFunc.java
```diff
@@ -0,0 +1,6 @@
+package io.kubernetes.client.informer.cache.function;
+
+@FunctionalInterface
+public interface KeyFunc<T> {
```
brendandburns: As above.

util/src/main/java/io/kubernetes/client/informer/cache/function/PopProcessFunc.java
```diff
@@ -0,0 +1,6 @@
+package io.kubernetes.client.informer.cache.function;
+
+public interface PopProcessFunc<T> {
```
brendandburns: as above.

util/src/main/java/io/kubernetes/client/informer/cache/function/ResyncFunc.java
```diff
@@ -0,0 +1,7 @@
+package io.kubernetes.client.informer.cache.function;
+
+@FunctionalInterface
+public interface ResyncFunc {
```
brendandburns: here too...

util/src/main/java/io/kubernetes/client/informer/exception/EmptyDeltasException.java
```diff
@@ -0,0 +1,8 @@
+package io.kubernetes.client.informer.exception;
+
+public class EmptyDeltasException extends RuntimeException {
```
brendandburns: use `java.util.NoSuchElementException`?

util/src/main/java/io/kubernetes/client/informer/impl/DefaultSharedIndexInformer.java
```diff
@@ -0,0 +1,221 @@
...
+          this.processor.distribute(
+              new ProcessorListener.DeleteNotification(delta.getObj()), false);
+          break;
+      }
+    }
+  }
+
+  @Override
+  public void addIndexers(Map indexers) {
+    // Not implemented yet
```
brendandburns: throw NotImplementedException

brendandburns:
Many thanks for the tests! I added a few more comments about reducing the number of custom classes in the implementation so that it's easier to understand.

util/src/main/java/io/kubernetes/client/informer/impl/DefaultSharedIndexInformer.java
brendandburns: Ah, that doesn't exist in Java.

```java
throw new RuntimeException("unimplemented!");
```

brendandburns:
```suggestion
    throw new RuntimeException("unimplemented!");
```

util/src/main/java/io/kubernetes/client/informer/cache/Deltas.java
yue9944882: > This is basically exactly a Java Deque

yesssssss. that's a great point. replaced w/ Deque

util/src/main/java/io/kubernetes/client/informer/cache/Deltas.java
yue9944882: done

util/src/main/java/io/kubernetes/client/informer/cache/function/IndexFunc.java
yue9944882: done

util/src/main/java/io/kubernetes/client/informer/cache/Deltas.java
yue9944882: no, it isn't. delta type represents an inner state of the queuing items which has one more type "Sync" comparing to event type. while event type is exposed as an informer's API.

util/src/main/java/io/kubernetes/client/informer/impl/DefaultSharedIndexInformer.java
yue9944882: done

util/src/main/java/io/kubernetes/client/informer/cache/Store.java
yue9944882: mmm.. let me try if it can be replaced by `Map`..

yue9944882: ur point is making sense to me. i'm trying to make the `Store` interface inherit from `Map`. but this really brings a large refactor throughout the pull. 👀i'm not even confident if it will work the same as the verified one in our internal environment. wdyt if i add a TODO comment and refactor this in a follow-up pull?

brendandburns: TODO is fine with me.

brendandburns:
Thanks for all the changes and the patience. Add the TODO and this is LGTM and ready to merge.

util/src/main/java/io/kubernetes/client/informer/SharedInformerFactory.java
```diff
@@ -0,0 +1,136 @@
...
+      Function<CallGeneratorParams, Call> callGenerator,
+      Class<ApiType> apiTypeClass,
+      Class<ApiListType> apiListTypeClass,
+      long resyncPeriodInMillis) {
+    ListerWatcher<ApiType, ApiListType> listerWatcher =
+        listerWatcherFor(callGenerator, apiTypeClass, apiListTypeClass);
+    SharedIndexInformer<ApiType> informer =
+        new DefaultSharedIndexInformer<ApiType, ApiListType>(
+            apiTypeClass, listerWatcher, resyncPeriodInMillis);
+    this.informers.put(TypeToken.get(apiTypeClass).getType(), informer);
```
adohe: no need to add this, as the cache will always be overwritten.

util/src/main/java/io/kubernetes/client/informer/SharedInformerFactory.java
```diff
@@ -0,0 +1,136 @@
...
+  /** Stop all registered informers. */
+  public synchronized void stopAllRegisteredInformers() {
+    if (Collections.isEmptyMap(informers)) {
+      return;
+    }
+    informers.forEach(
+        (informerType, informer) -> {
+          if (startedInformers.containsKey(informerType)) {
+            startedInformers.remove(informerType);
+            informer.stop();
```
adohe: should remove informer after stop

util/src/main/java/io/kubernetes/client/informer/cache/DeltaFIFO.java
adohe: this has no need to inside the lock, just a function call with the obj.

adohe:
so glad to see this get merged @yue9944882 @brendandburns  as a follow up we could help build more utility, like work-queue, leader election. Let's just get this Java informer framework fully functional.

util/src/main/java/io/kubernetes/client/informer/SharedInformerFactory.java
yue9944882: the order is fine

util/src/main/java/io/kubernetes/client/informer/cache/Store.java
yue9944882: added, thx

brendandburns:
@adohe do you want me to wait on an LGTM from you before I merge? Otherwise, I think this is good to go...

adohe:
overall lgtm. Let's just get it merge @brendandburns

PR #465, opened 2018-12-14
Improve auth tests, fix a bug in credential parsing

### LINKED ISSUE #460: ClassCastException observed while trying to read ~/.kube/config file.
Hi,

Observing a class cast exception when kubernetes client tries to verify if the key has expired. 
```
Exception in thread "main" java.lang.ClassCastException: java.lang.String cannot be cast to java.util.Date
	at io.kubernetes.client.util.authenticators.GCPAuthenticator.isExpired(GCPAuthenticator.java:40)
	at io.kubernetes.client.util.KubeConfig.getAccessToken(KubeConfig.java:200)
	at io.kubernetes.client.util.credentials.KubeconfigAuthentication.<init>(KubeconfigAuthentication.java:33)
	at io.kubernetes.client.util.ClientBuilder.kubeconfig(ClientBuilder.java:212)
	at io.kubernetes.client.util.ClientBuilder.standard(ClientBuilder.java:96)
	at io.kubernetes.client.util.ClientBuilder.standard(ClientBuilder.java:75)
	at io.kubernetes.client.util.Config.defaultClient(Config.java:104)
```

The kubectl config was generated using gcloud. 

The config file snippet is as follows.
```
current-context: cluster-1
kind: Config
preferences: {}
users:
- name: cluster-1
  user:
    auth-provider:
      config:
        access-token:  REDACTED
        cmd-args: config config-helper --format=json
        cmd-path: /usr/lib/google-cloud-sdk/bin/gcloud
        expiry: "2018-12-10T08:46:57Z"
        expiry-key: '{.credential.token_expiry}'
        token-key: '{.credential.access_token}'
      name: gcp

```
Interestingly I observe that there is a unit test for the same. https://github.com/kubernetes-client/java/blob/master/util/src/test/java/io/kubernetes/client/util/KubeConfigTest.java

Any pointers on to resolve this issue would be super helpful...

brendandburns:
Fixes #460

### DIFF
--- a/util/src/main/java/io/kubernetes/client/util/authenticators/GCPAuthenticator.java
+++ b/util/src/main/java/io/kubernetes/client/util/authenticators/GCPAuthenticator.java
@@ -13,8 +13,11 @@ limitations under the License.
 package io.kubernetes.client.util.authenticators;
 
 import io.kubernetes.client.util.KubeConfig;
+import java.time.Instant;
 import java.util.Date;
 import java.util.Map;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * The Authenticator interface represents a plugin that can handle a specific type of authentication
@@ -24,6 +27,8 @@ public class GCPAuthenticator implements Authenticator {
   static {
     KubeConfig.registerAuthenticator(new GCPAuthenticator());
   }
+
+  private static final Logger log = LoggerFactory.getLogger(GCPAuthenticator.class);
 
   @Override
   public String getName() {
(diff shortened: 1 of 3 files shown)


util/src/main/java/io/kubernetes/client/util/authenticators/GCPAuthenticator.java
```diff
@@ -37,7 +43,19 @@ public String getToken(Map<String, Object> config) {
...
   @Override
   public boolean isExpired(Map<String, Object> config) {
-    Date expiry = (Date) config.get("expiry");
+    Object expiryObj = config.get("expiry");
+    Date expiry = null;
+    if (expiryObj instanceof Date) {
+      expiry = (Date) expiryObj;
+    } else if (expiryObj instanceof String) {
+      try {
+        expiry = DateFormat.getDateInstance().parse((String) expiryObj);
```
yue9944882: [https://github.com/kubernetes/kubernetes/blob/a521285b7cf8273fddf9c8e912070f9e386ffa0d/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/time.go#L151](https://github.com/kubernetes/kubernetes/blob/a521285b7cf8273fddf9c8e912070f9e386ffa0d/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/time.go#L151)

how about explicitly parsing timestamp w/ RFC3339 format here?

brendandburns: Turns out there's not a clean way to do that in Java:

https://stackoverflow.com/questions/6038136/how-do-i-parse-rfc-3339-datetimes-with-java/41569330

But I followed the advice to switch to Instant which uses ISO8601 which is nearly identical.

brendandburns:
Comment addressed, please take another look.

Issue #430, opened 2018-11-09
Upgrade to okhttp3 or better yet, use a HTTP client facade like Feign


asarkar:
I see that a `4.0.0.-beta-1` branch is in development. This will be a good time to upgrade the OkHttp client to the latest. The migration is relatively painless. The `2.7.5` version been used is almost 2.5 years old, and of course, has been abandoned in favor of [okhttp3](https://search.maven.org/search?q=g:com.squareup.okhttp3).

Another option, albeit more ambitious, would be to use a Facade like [Feign](https://github.com/OpenFeign/feign) that lets the users plugin in their own HTTP clients. That would be extremely useful in many projects where people heavily customize HTTP clients for timeouts, SSL settings and what not.

brendandburns:
This code is all auto-generated by swagger-codegen:

https://github.com/swagger-api/swagger-codegen/tree/master/modules/swagger-codegen/src/main/resources/Java/libraries

I see that there is a feign generator, you could try re-generating with that generator...

Alternately you could upgrade the okhttp-gson client to use okhttp3:

https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/Java/libraries/okhttp-gson/api.mustache

asarkar:
@brendandburns I'm not clear on what you're suggesting. I'm not compiling the code myself, I'm picking up the `3.0.0` from Maven Central. I also checked the 4.0.0.-beta-1 [pom.xml](https://github.com/kubernetes-client/java/blob/client-java-parent-4.0.0-beta1/kubernetes/pom.xml#L204), and the version is clearly 2.7.5.

brendandburns:
@asarkar I don't think that the 2.7 and 3.0 APIs are the same, thus I don't think we can just upgrade the pom file. 

either way, we can't make changes to the generated code (or pom) in this repository, it needs to be done in the swagger-codegen repository.

asarkar:
@brendandburns It's unlikely that swagger will agree to simply upgrading a client version; perhaps a new module could be created for okhttp3?
As for the API, from what i've seen, the difference is mostly in the package names, and addition of some fluent builders. It is true that the migration is not simply a POM change, but it is also true that it's very straightforward.

What do you think?

brendandburns:
Maybe we should just try generating with the `feign` module instead?

I think either path would be fine. Let me know if you need pointers to the code to get started.

alannakelly:
[...]

PR #483, opened 2019-01-16
bump okhttp to v3

tedli:
issue: #430


tedli:
Is the current master branch without any touched able to pass those tests?

brendandburns:
Sadly this code is all auto-generated, we can't make manual changes in this code since we won't be able to regenerate it as the Kubernetes API changes.

The right path here is to regenerate the code to use something like feign instead of okhttp. It's a TODO, but if you want to take it on, I'm happy to help guide it.

I'm going to close this PR since we can't merge it. Please let me know if there are more questions.

Thanks

PR #497, opened 2019-02-04
Mark com.microsoft.azure.adal4j optional

briandealwis:
I'm trying to use the Java client in an OSGi application, which is failing as the Java client library has a dependency on the `com.microsoft.aad.adal4j` package.  This package cannot be found as `com.microsoft.azure:adal4j` dependency does not carry OSGi metadata.

But since this library is only required to use the Azure authenticator, we should mark this dependency as optional, which also marks the `com.microsoft.aad.adal4j` package dependency as optional too.

(This PR also fixes some stray tabs.)

### DIFF
--- a/util/pom.xml
+++ b/util/pom.xml
@@ -70,6 +70,7 @@
             <groupId>com.microsoft.azure</groupId>
             <artifactId>adal4j</artifactId>
             <version>1.6.0</version>
+            <optional>true</optional>
         </dependency>
         <dependency>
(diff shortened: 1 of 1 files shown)


briandealwis:
I signed it!

brendandburns:
CNCF bot is still not happy. Did you use the same email as your github handle?

Otherwise, this looks fine to me.

brendandburns:
/lgtm
/approve

Thanks!

Issue #290, opened 2018-06-19
GCP authentication does not support refreshing tokens


omerkarj:
This relates to #143.

GCPAuthenticator does not implement the refresh() method.
How can the client be used without manually refreshing the access token if it is expired?

for reference, initializing the client (i.e. running the example) with an expired token will result in:
`Caused by: java.lang.IllegalStateException: Unimplemented
	at io.kubernetes.client.util.authenticators.GCPAuthenticator.refresh(GCPAuthenticator.java:49)
	at io.kubernetes.client.util.KubeConfig.getAccessToken(KubeConfig.java:188)
	at io.kubernetes.client.util.credentials.KubeconfigAuthentication.<init>(KubeconfigAuthentication.java:33)
	at io.kubernetes.client.util.ClientBuilder.kubeconfig(ClientBuilder.java:165)
	at io.kubernetes.client.util.ClientBuilder.standard(ClientBuilder.java:80)
	at io.kubernetes.client.util.Config.defaultClient(Config.java:104)
	at serivces.KubernetesService.<clinit>(KubernetesService.groovy:23)
	... 2 more`

brendandburns:
There's no way to fix this right now, someone needs to implement the token refresh.

bootstraponline:
This breaks using the client. I had started on an integration then noticed the calls stopped working. Maybe mark the issue as contributions welcome if this isn't on the roadmap.

```kotlin
val client = Config.defaultClient()
Configuration.setDefaultApiClient(client)

val api = CoreV1Api(client)

// kubectl get service --field-selector metadata.name=esp-my-pod
val fieldSelector = "metadata.name=esp-my-pod"

val result = api.listServiceForAllNamespaces(
        null,
        fieldSelector,
        null,
        "",
        1,
        null,
        null,
        2 * 60,
        false)
```

```java
Exception in thread "main" java.lang.IllegalStateException: Unimplemented
	at io.kubernetes.client.util.authenticators.GCPAuthenticator.refresh(GCPAuthenticator.java:49)
	at io.kubernetes.client.util.KubeConfig.getAccessToken(KubeConfig.java:188)
	at io.kubernetes.client.util.credentials.KubeconfigAuthentication.<init>(KubeconfigAuthentication.java:33)
	at io.kubernetes.client.util.ClientBuilder.kubeconfig(ClientBuilder.java:165)
	at io.kubernetes.client.util.ClientBuilder.standard(ClientBuilder.java:80)
[...]

Issue #238, opened 2018-04-14
Port exec-based client auth provider to Java client


dbenhur:
With 1.10, this exec-based client auth credential provider alpha feature was introduced  [kubernetes/features#541](https://github.com/kubernetes/features/issues/541) for client-go.  Exec-based auth plugin should be supported by this java client in addition to client-go. Umbrella issue: [kubernetes#62185](https://github.com/kubernetes/kubernetes/issues/62185)

/kind feature
/sig api-machinery

aniruddh-a:
Is anyone working on this issue ?

brendandburns:
@aniruddh-a sorry for the delay in responding. No one is working on this. If you're inetersted we'd love a PR.

Thanks

mustafaakin:
Is help still needed? We implemented it in a hacky-way but would love to contribute it in a cleaner way, it is simple enough to do so.

vkahawala:
Hi @mustafaakin I would verymuch like to know how you solved this issue. I saw your post under [aws-iam-authenticator](https://github.com/kubernetes-sigs/aws-iam-authenticator/issues/168). I was able to get that piece of code working. But I'm having trouble using that token with kubernetes java client.

Thanks

jglick:
Seems to be `KubeConfig.getAccessToken` which needs patching. I did so for another client: https://github.com/fabric8io/kubernetes-client/pull/1224

jglick:
Since this is implemented at least in a basic form, I think this issue could be closed and follow-ups filed as needed for the items I noted in #512 or anything else that comes up.

PR #512, opened 2019-02-20
#238: support client.authentication.k8s.io/v1beta1 ExecCredential

jglick:
Implementation of #238 allowing `~/.kube/config` to use the new generic exec credentials rather than a vendor-specific plugin or hard-coded token.

Planning to do before taking out of “draft” status:

- [X] basic implementation (run a command, get a token)
- [X] observed working with GKE
- [x] `KubeConfigTest` addition
- [x] relativization to basedir
- [x] environment variables
- [x] more robustness against bad inputs, logging, etc.

Other things which I think could be left for follow-ups depending on demand:

- [ ] observed working with EKS (I could make a test cluster if necessary)
- [ ] time-based expiry (see https://github.com/kubernetes-client/java/issues/290#issuecomment-453180074 or inline comments for why this would require broader refactoring)
- [ ] 401-based expiry (even trickier)
- [ ] client certificate support

### DIFF
--- a/util/src/main/java/io/kubernetes/client/util/KubeConfig.java
+++ b/util/src/main/java/io/kubernetes/client/util/KubeConfig.java
@@ -249,6 +249,7 @@ public class KubeConfig {
    *     href="https://kubernetes.io/docs/reference/access-authn-authz/authentication/#client-go-credential-plugins">
    *     Authenticating » client-go credential plugins</a>
    */
+  @SuppressWarnings("unchecked")
   private String tokenViaExecCredential(Map<String, Object> execMap) {
     if (execMap == null) {
       return null;
@@ -260,8 +261,28 @@ public class KubeConfig {
       return null;
     }
     String command = (String) execMap.get("command");
-    List<Map<String, String>> env = (List) execMap.get("env");
-    List<String> args = (List) execMap.get("args");
+    JsonElement root = runExec(command, (List) execMap.get("args"), (List) execMap.get("env"));
+    if (root == null) {
+      return null;
+    }
+    // TODO verify .apiVersion and .kind = ExecCredential
+    JsonObject status = root.getAsJsonObject().get("status").getAsJsonObject();
+    JsonElement token = status.get("token");
+    if (token == null) {
+      // TODO handle clientCertificateData/clientKeyData
+      // (KubeconfigAuthentication is not yet set up for that to be dynamic)
+      log.warn("No token produced by {}", command);
+      return null;
+    }
+    return token.getAsString();
+    // TODO cache tokens between calls, up to .status.expirationTimestamp
+    // TODO a 401 is supposed to force a refresh,
+    // but KubeconfigAuthentication hardcodes AccessTokenAuthentication which does not support that
+    // and anyway ClientBuilder only calls Authenticator.provide once per ApiClient;
+    // we would need to do it on every request
+  }
+
+  private JsonElement runExec(String command, List<String> args, List<Map<String, String>> env) {
     // TODO relativize command to basedir of config file (requires KubeConfig to be given a basedir)
     List<String> argv = new ArrayList<>();
     argv.add(command);
@@ -270,7 +291,9 @@ public class KubeConfig {
     }
     ProcessBuilder pb = new ProcessBuilder(argv);
     if (env != null) {
-      // TODO apply
+      for (Map<String, String> entry : env) {
+        pb.environment().put(entry.get("name"), entry.get("value"));
+      }
     }
     pb.redirectError(ProcessBuilder.Redirect.INHERIT);
     try {
@@ -288,25 +311,11 @@ public class KubeConfig {
         log.error("{} failed with exit code {}", command, r);
         return null;
       }
-      // TODO verify .apiVersion and .kind = ExecCredential
-      JsonObject status = root.getAsJsonObject().get("status").getAsJsonObject();
-      JsonElement token = status.get("token");
-      if (token == null) {
-        // TODO handle clientCertificateData/clientKeyData
-        // (KubeconfigAuthentication is not yet set up for that to be dynamic)
-        log.warn("No token produced by {}", command);
-        return null;
-      }
-      return token.getAsString();
+      return root;
     } catch (IOException | InterruptedException x) {
       log.error("Failed to run " + command, x);
       return null;
     }
-    // TODO cache tokens between calls, up to .status.expirationTimestamp
-    // TODO a 401 is supposed to force a refresh,
-    // but KubeconfigAuthentication hardcodes AccessTokenAuthentication which does not support that
-    // and anyway ClientBuilder only calls Authenticator.provide once per ApiClient;
-    // we would need to do it on every request
   }
 
   public boolean verifySSL() {
--- a/util/src/main/java/io/kubernetes/client/util/KubeConfig.java
+++ b/util/src/main/java/io/kubernetes/client/util/KubeConfig.java
@@ -257,8 +257,8 @@ public class KubeConfig {
       return null;
     }
     String apiVersion = (String) execMap.get("apiVersion");
-    if (!"client.authentication.k8s.io/v1beta1".equals(apiVersion)) {
-      // TODO or v1alpha1 is apparently identical and could be supported
+    if (!"client.authentication.k8s.io/v1beta1".equals(apiVersion)
+        && !"client.authentication.k8s.io/v1alpha1".equals(apiVersion)) {
       log.error("Unrecognized user.exec.apiVersion: {}", apiVersion);
       return null;
     }
--- a/util/src/main/java/io/kubernetes/client/util/KubeConfig.java
+++ b/util/src/main/java/io/kubernetes/client/util/KubeConfig.java
@@ -283,6 +283,7 @@ public class KubeConfig {
       log.warn("No token produced by {}", command);
       return null;
     }
+    log.debug("Obtained a token from {}", command);
     return token.getAsString();
     // TODO cache tokens between calls, up to .status.expirationTimestamp
     // TODO a 401 is supposed to force a refresh,
--- a/util/src/main/java/io/kubernetes/client/util/KubeConfig.java
+++ b/util/src/main/java/io/kubernetes/client/util/KubeConfig.java
@@ -222,59 +222,9 @@ public class KubeConfig {
         }
       }
     }
-    Object exec = currentUser.get("exec");
-    if (exec != null) {
-      // TODO extract to helper method for clarity
-      Map<String, Object> execMap = (Map<String, Object>) exec;
-      // https://kubernetes.io/docs/reference/access-authn-authz/authentication/#client-go-credential-plugins
-      String apiVersion = (String) execMap.get("apiVersion");
-      if ("client.authentication.k8s.io/v1beta1".equals(apiVersion)) { // TODO or v1alpha1 is apparently identical and could be supported
-        String command = (String) execMap.get("command");
-        List<Map<String, String>> env = (List) execMap.get("env");
-        List<String> args = (List) execMap.get("args");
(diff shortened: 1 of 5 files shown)


util/src/main/java/io/kubernetes/client/util/KubeConfig.java
```diff
@@ -214,6 +222,60 @@ public String getAccessToken() {
         }
       }
     }
+    Object exec = currentUser.get("exec");
+    if (exec != null) {
+      // TODO extract to helper method for clarity
+      Map<String, Object> execMap = (Map<String, Object>) exec;
```
brendandburns: It's probably better to declare an actual Java object that represents the `exec` field rather than using generic JSON accessors, it makes the code pretty ugly.

util/src/main/java/io/kubernetes/client/util/KubeConfig.java
```diff
@@ -214,6 +222,60 @@ public String getAccessToken() {
...
+          Process proc = pb.start();
+          JsonElement root = null;
+          try (InputStream is = proc.getInputStream();
+               Reader r = new InputStreamReader(is, StandardCharsets.UTF_8)) {
+            root = new JsonParser().parse(r);
+          } catch (JsonParseException x) {
+            log.error("Failed to parse output of " + command, x);
+          }
+          int r = proc.waitFor();
+          if (r == 0) {
```
brendandburns: flip this around for less indents:

```java
if (r != 0) {
   log.error(...)
   return;
}
...
```

util/src/main/java/io/kubernetes/client/util/KubeConfig.java
```diff
@@ -214,6 +222,60 @@ public String getAccessToken() {
...
+            if (root != null) {
```
brendandburns: Same thing with this conditional.

util/src/main/java/io/kubernetes/client/util/KubeConfig.java
```diff
@@ -214,6 +222,60 @@ public String getAccessToken() {
...
+              // TODO verify .apiVersion and .kind = ExecCredential
+              JsonObject status = root.getAsJsonObject().get("status").getAsJsonObject();
```
brendandburns: Let's create a real Java object for this response value too.

util/src/main/java/io/kubernetes/client/util/KubeConfig.java
brendandburns: (I realized I'm being hypocritical given the above, so feel free to ignore this...)

util/src/main/java/io/kubernetes/client/util/KubeConfig.java
brendandburns: As above, feel free to skip if you want.

util/src/main/java/io/kubernetes/client/util/KubeConfig.java
```diff
@@ -214,6 +222,60 @@ public String getAccessToken() {
         }
       }
     }
+    Object exec = currentUser.get("exec");
```
brendandburns: Please break this out into a separate method.

Actually, I think it would be good like this:

`getExecAuthProviderToken()`

```java
public String getExecAuthProviderToken(...) {
   String command = getAuthCommand(...);
   Object response = runCommand(command);
   return extractToken(response);
}
```

Or something like that.

brendandburns:
Thanks for starting this! I have a few comments, also this needs unit tests.

util/src/main/java/io/kubernetes/client/util/KubeConfig.java
jglick: > Please break this out into a separate method.

Yes, already noted as per comment two lines down.

util/src/main/java/io/kubernetes/client/util/KubeConfig.java
jglick: > hypocritical given the above

You mean the `authProviderMap` code?

After years of bad experiences I have a general preference for plain old Java code over use of fancy reflection-based (de)serialization libraries, but if you have a strong preference for some particular approach (GSON?) I will follow your guidelines.

util/src/main/java/io/kubernetes/client/util/KubeConfig.java
jglick: Yes, when extracting to its own method it is possible to have multiple `return` statements which I find easier to follow.

util/src/main/java/io/kubernetes/client/util/KubeConfig.java
jglick: (ditto)

util/src/main/java/io/kubernetes/client/util/KubeConfig.java
jglick: (ditto)

util/src/main/java/io/kubernetes/client/util/KubeConfig.java
```diff
@@ -229,6 +244,99 @@ public String getAccessToken() {
...
+    // TODO cache tokens between calls, up to .status.expirationTimestamp
+    // TODO a 401 is supposed to force a refresh,
+    // but KubeconfigAuthentication hardcodes AccessTokenAuthentication which does not support that
+    // and anyway ClientBuilder only calls Authenticator.provide once per ApiClient;
+    // we would need to do it on every request
+  }
+
+  private JsonElement runExec(String command, List<String> args, List<Map<String, String>> env) {
+    List<String> argv = new ArrayList<>();
+    if (command.startsWith("./")) {
```
brendandburns: This is Linux specific, use `File.isAbsolute(command)` instead.

brendandburns: and also, presumably you want to catch `../foo/bar` too which this doesn't currently cover.

util/src/main/java/io/kubernetes/client/util/KubeConfig.java
```diff
@@ -229,6 +244,99 @@ public String getAccessToken() {
...
+      for (Map<String, String> entry : env) {
+        pb.environment().put(entry.get("name"), entry.get("value"));
+      }
+    }
+    pb.redirectError(ProcessBuilder.Redirect.INHERIT);
+    try {
+      Process proc = pb.start();
+      JsonElement root;
+      try (InputStream is = proc.getInputStream();
+          Reader r = new InputStreamReader(is, StandardCharsets.UTF_8)) {
```
brendandburns: indentation is weird here, is this really correct?

util/src/main/java/io/kubernetes/client/util/KubeConfig.java
```diff
@@ -229,6 +244,99 @@ public String getAccessToken() {
...
+        log.error("Failed to parse output of " + command, x);
+        return null;
+      }
+      int r = proc.waitFor();
+      if (r != 0) {
+        log.error("{} failed with exit code {}", command, r);
+        return null;
+      }
+      return root;
+    } catch (IOException | InterruptedException x) {
```
brendandburns: why swallow the exception here? wouldn't you rather throw?

brendandburns:
Looks pretty good, some minor nits.

util/src/main/java/io/kubernetes/client/util/KubeConfig.java
jglick: > This is Linux specific

Not really. It should work fine on Windows as well.

> use `File.isAbsolute(command)` instead

That is not going to work since the _normal_ case is a plain command from the path like `aws-iam-authenticator`.

> presumably you want to catch `../foo/bar` too

As noted in the TODO comment, it is unclear. The [official documentation](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#configuration) does not give any definition of “[r]elative command paths”. Is `bin/script` to be treated this way, or only `./bin/script`? `.\bin\script`? Is `/usr/local/bin/script` legal, despite clearly not being “relative”? Test cases for the Golang client which actually run a script [only cover](https://github.com/kubernetes/client-go/blob/9c9f7f424e65af6ded246bd4ba1a1cec988acc7b/plugin/pkg/client/auth/exec/exec_test.go#L639) the `./` prefix. From the [Golang implementation](https://github.com/kubernetes/client-go/blob/9c9f7f424e65af6ded246bd4ba1a1cec988acc7b/plugin/pkg/client/auth/exec/exec.go#L304-L312) it is not even clear that this system works at all: it appears to resolve paths against the process current working directory, which might be completely unrelated to the location of `$KUBECONFIG`—I presume the test passes merely because the Go test runner changes directory to the parent of `exec_test.go`.

util/src/main/java/io/kubernetes/client/util/KubeConfig.java
jglick: What looks incorrect? I do not have a choice about indentation—it is whatever `mvn fmt:format` picks.

util/src/main/java/io/kubernetes/client/util/KubeConfig.java
jglick: It is not swallowed—it is sent to the log.

As to whether it is appropriate to throw exceptions out of `getAccessToken` (presumably wrapped in `RuntimeException`?) as opposed to logging and proceeding with other authentication modes and ultimately returning null (unauthenticated), I have no strong opinion. I was following the example of existing code which proceeds despite errors (`Failed to read token file` for example).

On the other hand `AzureActiveDirectoryAuthenticator.refresh` appears to throw an undeclared exception out of `getAccessToken`. `KubeconfigAuthentication.<init>` is permitted to throw `IOException` so perhaps `getAccessToken` should as well? That would mean reworking some existing code which catches and logs exceptions, and adding `IOException` to the signature of `Authenticator.refresh`.

util/src/main/java/io/kubernetes/client/util/KubeConfig.java
jglick: Experimentally, `kubectl` seems to work with for example `../bin/gke-auth` even when run from an arbitrary directory. Unclear to me from looking at sources how that works. At any rate, the best I can think of is to treat any `command` containing either `/` or `\` as a (possibly absolute) file path to be resolved against the `$(basedir $KUBECONFIG)`.

util/src/main/java/io/kubernetes/client/util/KubeConfig.java
brendandburns: ok, the 'R' aligned with the '(' looks weird to me, but if mvn:format says so, I guess it's right :P

util/src/main/java/io/kubernetes/client/util/KubeConfig.java
brendandburns: ok, sounds like it's messy, let's take it as a follow-on cleanup exercise...

brendandburns:
/lgtm
/approve

Thanks!

util/src/main/java/io/kubernetes/client/util/KubeConfig.java
jglick: Yeah my preference would be to line up the `I` and the `R`.

util/src/main/java/io/kubernetes/client/util/KubeConfig.java
jglick: Noted in PR description.

util/src/main/java/io/kubernetes/client/util/KubeConfig.java
klewi: @jglick - is the item "observed working with EKS (I could make a test cluster if necessary)" still open, and if not, could I help out with that?

jglick: Yes it is still open. If you are using EKS, please help test this system.

PR #520, opened 2019-03-04
Added support for KUBECONFIG environment variable with multiple paths.

roldevg:
Use first kubeconfig file.

Kubectl documentation: 
https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/#linux-1

### DIFF
--- a/util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
+++ b/util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
@@ -130,7 +130,7 @@ public class ClientBuilder {
       final String[] filePaths = kubeConfigEnv.split(File.pathSeparator);
       final String kubeConfigPath = filePaths[0];
       if (filePaths.length > 1) {
-        log.debug(
+        log.warn(
             "Found multiple kubeconfigs files, $KUBECONFIG: " + kubeConfigEnv + " using first: {}",
             kubeConfigPath);
       }
--- a/util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
+++ b/util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
@@ -104,16 +104,38 @@ public class ClientBuilder {
   }
 
   private static File findConfigFromEnv() {
-    String kubeConfigPath = System.getenv(ENV_KUBECONFIG);
+    final KubeConfigEnvParser kubeConfigEnvParser = new KubeConfigEnvParser();
+
+    final String kubeConfigPath =
+        kubeConfigEnvParser.parseKubeConfigPath(System.getenv(ENV_KUBECONFIG));
     if (kubeConfigPath == null) {
       return null;
     }
+
     final File kubeConfig = new File(kubeConfigPath);
     if (kubeConfig.exists()) {
       return kubeConfig;
     } else {
       log.debug("Could not find file specified in $KUBECONFIG");
       return null;
+    }
+  }
+
+  private static class KubeConfigEnvParser {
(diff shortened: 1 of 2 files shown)


roldevg:
I signed it

roldevg:
I signed it

roldevg:
I signed it

roldevg:
I signed it

util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
```diff
@@ -117,6 +121,24 @@ private static File findConfigFromEnv() {
...
+  private static class KubeConfigEnvParser {
+    private String parseKubeConfigPath(String kubeConfigEnv) {
+      if (kubeConfigEnv == null) {
+        return null;
+      }
+
+      final String[] filePaths = kubeConfigEnv.split(File.pathSeparator);
+      final String kubeConfigPath = filePaths[0];
+      if (filePaths.length > 1) {
+        log.info(
```
brendandburns: I think this should be log.warning, not log.info since we're explicitly doing something that isn't quite what the user expects...

util/src/test/java/io/kubernetes/client/util/ClientBuilderTest.java
```diff
@@ -76,6 +77,14 @@ public void testDefaultClientReadsKubeConfig() throws Exception {
     assertEquals("http://kubeconfig.dir.com", client.getBasePath());
   }
 
+  @Test
+  public void testDefaultClientReadsKubeConfigMultiple() throws Exception {
+    final String kubeConfigEnv = KUBECONFIG_FILE_PATH + File.pathSeparator + KUBECONFIG_FILE_PATH;
```
brendandburns: Let's make the second file something different like `/doesn't/exist` or somesuch to validate the "chooses first" is correct.

brendandburns:
Thanks for the PR! generally LGTM, a few small comments.

util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
roldevg: Yeah, you are right, fixed.

util/src/test/java/io/kubernetes/client/util/ClientBuilderTest.java
roldevg: fixed

util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
```diff
@@ -117,6 +121,24 @@ private static File findConfigFromEnv() {
...
+    private String parseKubeConfigPath(String kubeConfigEnv) {
+      if (kubeConfigEnv == null) {
+        return null;
+      }
+
+      final String[] filePaths = kubeConfigEnv.split(File.pathSeparator);
+      final String kubeConfigPath = filePaths[0];
+      if (filePaths.length > 1) {
+        log.debug(
+            "Found multiple kubeconfigs files, $KUBECONFIG: " + kubeConfigEnv + " using first: {}",
```
brendandburns: Hrm, isn't debug even lower level than warning? I think this should be warning.

brendandburns:
looks good, but I think we want a higher log level, not a lower log level.

Thanks!

util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
roldevg: Sorry, I have not understood. Fixed.

PR #453, opened 2018-12-04
Upgraded gson dependency to 2.8.0

craigatgoogle:
Resolves: https://github.com/kubernetes-client/java/issues/452

PR #73, opened 2017-09-20
Rename parent POM & util POM to make client-java the primary artifact name

lwander:
Closes https://github.com/kubernetes-client/java/issues/64

PR #521, opened 2019-03-06
Delete non-Maven build scripts

jglick:
When this project was created, in #1, it included a `kubernetes/build.gradle` and related files. It _also_ included a Maven `kubernetes/pom.xml`, oh and a Scala-style `kubernetes/build.sbt` for good measure. The initial `kubernetes/.travis.yml` suggests that this was all just dumped here by Swagger, not deliberate. Later in #5, #73, etc. the structure was rearranged, all assuming a Maven build—as does the current (top-level) `.travis.yml` and `README.md`. In fact the current structure has a root POM and declares the `kubernetes` subdirectory as a `module`, which implies that it is to be built via Maven. The `build.gradle` is clearly out of date—for example, it [does not reflect](https://github.com/kubernetes-client/java/blob/2d22dced9b7fc4f77f4f72720539f2ae155df125/kubernetes/build.gradle#L100) #453.

My practical interest in this is that the NetBeans IDE currently prefers `build.gradle` to `pom.xml` when both are present in a directory, so the `kubernetes` subdirectory gets opened as a Gradle project and it is a mess. If build scripts and related metadata are not actually current and usable, they are better deleted to avoid confusion.


brendandburns:
Fine with me, but please add to https://github.com/kubernetes-client/java/blob/master/kubernetes/.swagger-codegen-ignore or else we'll just regenerate them next time we codegen.

Thanks!

jglick:
> please add to `.swagger-codegen-ignore`

Done, though I am not sure how to test this. (Not familiar with Swagger.)

PR #522, opened 2019-03-07
Fixes watch conflict handling in informer and more test coverage

yue9944882:


util/src/main/java/io/kubernetes/client/informer/cache/ReflectorRunnable.java
```diff
@@ -99,7 +99,7 @@ public void run() {
             return;
           }
           log.error("{}#Watch failed as {} unexpected", apiTypeClass, t.getMessage(), t);
-          continue;
+          return;
```
yue9944882: the first bug uncovered by conflicting watch event

/cc @brendandburns

PR #525, opened 2019-03-09
add workqueue implementation for kube-style controller

adohe:
ref [#523](https://github.com/kubernetes-client/java/issues/523).

Need more tests.

### DIFF
--- a/extended/src/main/java/io/kubernetes/client/extended/wait/Wait.java
+++ b/extended/src/main/java/io/kubernetes/client/extended/wait/Wait.java
@@ -1,0 +1,49 @@
+package io.kubernetes.client.extended.wait;
+
+import java.time.Duration;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Supplier;
+
+public class Wait {
+
+  /**
+   * Poll tries a condition func until it returns true, an exception, or the timeout is reached.
+   *
+   * @param interval the check interval
+   * @param timeout the timeout period
+   * @param condition the condition func
+   */
+  public static boolean poll(Duration interval, Duration timeout, Supplier<Boolean> condition) {
+    ScheduledExecutorService executorService = Executors.newSingleThreadScheduledExecutor();
+    AtomicBoolean result = new AtomicBoolean(false);
+    long dueDate = System.currentTimeMillis() + timeout.toMillis();
+    ScheduledFuture<?> future =
+        executorService.scheduleAtFixedRate(
+            () -> {
+              try {
+                result.set(condition.get());
+              } catch (Exception e) {
+                result.set(false);
+              }
+            },
+            interval.toMillis(),
+            interval.toMillis(),
+            TimeUnit.MILLISECONDS);
+    try {
+      while (System.currentTimeMillis() < dueDate) {
+        if (result.get()) {
+          future.cancel(true);
+          return true;
+        }
+      }
+    } catch (Exception e) {
+      return result.get();
+    }
+    future.cancel(true);
+    return result.get();
+  }
+}
--- a/extended/src/main/java/io/kubernetes/client/extended/workqueue/DefaultDelayingQueue.java
+++ b/extended/src/main/java/io/kubernetes/client/extended/workqueue/DefaultDelayingQueue.java
@@ -1,0 +1,136 @@
+package io.kubernetes.client.extended.workqueue;
+
+import com.google.common.primitives.Longs;
+import java.time.Duration;
+import java.time.Instant;
+import java.time.temporal.Temporal;
+import java.util.Map;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.DelayQueue;
+import java.util.concurrent.Delayed;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.TimeUnit;
+
+/** The default delaying queue implementation. */
+public class DefaultDelayingQueue<T> extends DefaultWorkQueue<T> implements DelayingQueue<T> {
+
+  public static Duration heartBeatInterval = Duration.ofSeconds(10);
+
+  private DelayQueue<WaitForEntry<T>> delayQueue;
+  private ConcurrentMap<T, WaitForEntry<T>> waitingEntryByData;
+  protected BlockingQueue<WaitForEntry<T>> waitingForAddQueue;
+
+  public DefaultDelayingQueue(ExecutorService waitingWorker) {
+    this.delayQueue = new DelayQueue<>();
+    this.waitingEntryByData = new ConcurrentHashMap<>();
+    this.waitingForAddQueue = new LinkedBlockingQueue<>(1000);
+    waitingWorker.submit(this::waitingLoop);
+  }
+
+  public DefaultDelayingQueue() {
+    this(Executors.newSingleThreadExecutor());
+  }
+
+  public void addAfter(T item, Duration duration) {
+    // don't add if we're already shutting down
+    if (super.isShuttingDown()) {
+      return;
+    }
+
+    // immediately add things w/o delay
+    if (duration.isZero()) {
+      super.add(item);
+      return;
+    }
+    WaitForEntry<T> entry = new WaitForEntry<>(item, duration.addTo(Instant.now()));
+    this.waitingForAddQueue.offer(entry);
+  }
+
+  private void waitingLoop() {
+    try {
+      while (true) {
+        // underlying work-queue is shutting down, quit the loop.
+        if (super.isShuttingDown()) {
+          return;
+        }
+        // peek the first item from the delay queue
+        WaitForEntry<T> entry = delayQueue.peek();
+        // default next ready-at time to "never"
+        Duration nextReadyAt = heartBeatInterval;
+        if (entry != null) {
+          // the delay-queue isn't empty, so we deal with the item in the following logic:
+          // 1. check if the item is ready to fire
+          //   a. if ready, remove it from the delay-queue and push it into underlying work-queue
+          //   b. if not, refresh the next ready-at time.
+          Instant now = Instant.now();
+          if (!Duration.between(entry.readyAtMillis, now).isNegative()) {
+            delayQueue.remove(entry);
+            super.add(entry.data);
+            this.waitingEntryByData.remove(entry.data);
+            continue;
+          } else {
+            nextReadyAt = Duration.between(now, entry.readyAtMillis);
+          }
+        }
+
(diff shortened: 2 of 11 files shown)


extended/src/main/java/io/kubernetes/client/workqueue/DefaultDelayingQueue.java
```diff
@@ -0,0 +1,70 @@
...
+
+import java.time.Duration;
+import java.util.concurrent.DelayQueue;
+import java.util.concurrent.Delayed;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * The default delaying queue implementation.
```
brendandburns: Please add more documentation about what this is and how it is used. We can't assume people are familiar with the client go approach.

extended/src/main/java/io/kubernetes/client/workqueue/DefaultDelayingQueue.java
```diff
@@ -0,0 +1,70 @@
...
+ */
+public class DefaultDelayingQueue extends DefaultWorkQueue implements DelayingQueue {
+
+    private DelayQueue      delayQueue    = new DelayQueue();
+    private ExecutorService waitingWorker = Executors.newSingleThreadExecutor();
+
+    public DefaultDelayingQueue() {
```
brendandburns: It's probably worth making the `ExecutorService` an optional parameter, for testing and other reasons.

extended/src/main/java/io/kubernetes/client/workqueue/DefaultDelayingQueue.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.concurrent.DelayQueue;
+import java.util.concurrent.Delayed;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * The default delaying queue implementation.
+ */
+public class DefaultDelayingQueue extends DefaultWorkQueue implements DelayingQueue {
```
brendandburns: Use generic types instead of just passing `Object`

extended/src/main/java/io/kubernetes/client/workqueue/DefaultDelayingQueue.java
```diff
@@ -0,0 +1,70 @@
...
+                }
+                WaitForEntry entry = (WaitForEntry) delayQueue.take();
+                super.add(entry.data);
+            }
+        } catch (InterruptedException e) {
+            return;
+        }
+    }
+
+    private class WaitForEntry implements Delayed {
```
brendandburns: make this class static. (alternately, just use an anonymous class above)

I think I prefer the anonymous class approach.

extended/src/main/java/io/kubernetes/client/workqueue/DefaultRateLimitingQueue.java
```diff
@@ -0,0 +1,78 @@
+package io.kubernetes.client.workqueue;
+
+import java.time.Duration;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * The default rate limiting queue implementation.
+ */
+public class DefaultRateLimitingQueue extends DefaultDelayingQueue implements RateLimitingQueue {
```
brendandburns: As above wrt generics.

extended/src/main/java/io/kubernetes/client/workqueue/DefaultWorkQueue.java
```diff
@@ -0,0 +1,123 @@
...
+    private Set<Object>         processing;
+
+    private AtomicBoolean shuttingDown = new AtomicBoolean(false);
+
+    private final ReentrantLock lock = new ReentrantLock();
+
+    private final Condition notEmpty = lock.newCondition();
+
+    public DefaultWorkQueue() {
+        this.queue = new LinkedList<>();
```
brendandburns: Lists.newLinkedList() (or don't use Sets below...)

extended/src/main/java/io/kubernetes/client/workqueue/DefaultWorkQueue.java
```diff
@@ -0,0 +1,123 @@
...
+
+    // Things that are currently being processed are in the processing set.
+    // These things may be simultaneously in the dirty set. When we finish
+    // processing something and remove it from this set, we'll check if
+    // it's in the dirty set, and if so, add it to the queue.
+    private Set<Object>         processing;
+
+    private AtomicBoolean shuttingDown = new AtomicBoolean(false);
+
+    private final ReentrantLock lock = new ReentrantLock();
```
brendandburns: Get rid of this lock and just used `synchronized` on the methods.

extended/src/main/java/io/kubernetes/client/workqueue/DefaultWorkQueue.java
```diff
@@ -0,0 +1,123 @@
...
+            lock.unlock();
+        }
+    }
+
+    @Override
+    public Object get() {
+        lock.lock();
+        try {
+            while (queue.size() == 0 && !shuttingDown.get()) {
+                notEmpty.await();
```
brendandburns: use `this.wait()` and `this.notify()` instead of condition variable.

extended/src/main/java/io/kubernetes/client/workqueue/WorkQueue.java
```diff
@@ -0,0 +1,49 @@
...
+     *
+     * @return current queue length
+     */
+    int length();
+
+    /**
+     * Get blocks until it can return an item to be processed. If shutdown = true,
+     * the caller should end their process.
+     *
+     * @return the object
```
brendandburns: which object?

extended/src/main/java/io/kubernetes/client/workqueue/WorkQueue.java
```diff
@@ -0,0 +1,49 @@
+package io.kubernetes.client.workqueue;
+
+/**
+ * The workqueue interface defines the queue behavior.
+ */
+public interface WorkQueue {
```
brendandburns: Make this a generic instead of using `Object`

extended/src/main/java/io/kubernetes/client/workqueue/RateLimiter.java
```diff
@@ -0,0 +1,30 @@
+package io.kubernetes.client.workqueue;
+
+import java.time.Duration;
+
+public interface RateLimiter {
```
brendandburns: There are lots of other retry libraries out there (e.g. https://github.com/jhalterman/failsafe) do we need to implement this ourselves?

extended/src/main/java/io/kubernetes/client/workqueue/DefaultWorkQueue.java
```diff
@@ -0,0 +1,123 @@
...
+            }
+            if (queue.size() == 0) {
+                // We must be shutting down
+                return null;
+            }
+            Object obj = this.queue.poll();
+            this.processing.add(obj);
+            this.dirty.remove(obj);
+            return obj;
+        } catch (Throwable t) {
```
brendandburns: Why do you want to catch here? this seems like a bad idea... (at the very least we should log)

extended/src/main/java/io/kubernetes/client/workqueue/DefaultWorkQueue.java
```diff
@@ -0,0 +1,123 @@
...
+                notEmpty.signal();
+            }
+        } finally {
+            lock.unlock();
+        }
+    }
+
+    @Override
+    public void shutDown() {
+        this.shuttingDown.compareAndSet(false, true);
```
brendandburns: I think you need to `notify()` here too...

extended/src/main/java/io/kubernetes/client/workqueue/DefaultWorkQueue.java
```diff
@@ -0,0 +1,123 @@
...
+import com.google.common.collect.Sets;
+
+import java.util.LinkedList;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.ReentrantLock;
+
+/**
+ * The default workqueue implementation.
```
brendandburns: You should document how you expect this to be used.

brendandburns:
I have three high level comments:
1) Please use generics everywhere.
2) Please add much better documentation about usage, I don't think anyone reading this library will know how to use it.
3) If we only ever expect to have one implementation, let's not have interfaces, let's just have the implementation.

Other comments: More tests are definitely required, some sample code would be great.

yue9944882:
thx so much for help w/ this @adohe !

do you think about splitting this pull into two parts? one for initializing the new module, another for the implementation so that it won't be the blocker for other tasks. thx!

yue9944882:
@adohe seems recently too swamped to bump the thread. let me take over it. we'll finally get a java controller example into this repo.

extended/src/main/java/io/kubernetes/client/workqueue/DefaultDelayingQueue.java
adohe: done.

extended/src/main/java/io/kubernetes/client/workqueue/DefaultDelayingQueue.java
adohe: just use static inner class.

extended/src/main/java/io/kubernetes/client/workqueue/RateLimiter.java
adohe: didn't notice this before, will take a deep look

extended/src/main/java/io/kubernetes/client/workqueue/DefaultWorkQueue.java
adohe: this depends on how to handle exception, throw exception out here seems better

extended/src/main/java/io/kubernetes/client/workqueue/DefaultWorkQueue.java
adohe: sure, should use `notifyAll()` here

adohe:
comments addressed, I am working on much better usage documentation and test cases. cc @brendandburns @yue9944882

yue9944882:
ci looks not happy

cizezsy:
Any update on this PR? I really want to use this feature in client-java :)

yue9944882:
@cizezsy the guy is recently too busy to push the progress. and also, the pull is a bit more complex so it might take time. welcome to split an another pass if you want to contribute.

extended/src/main/java/io/kubernetes/client/extended/function/ConditionFunc.java
```diff
@@ -0,0 +1,6 @@
+package io.kubernetes.client.extended.function;
+
+@FunctionalInterface
+public interface ConditionFunc {
```
yue9944882: remove it, use functional.Supplier as replacement

yue9944882:
/lgtm
/approve
/hold

lazy consensus if @brendandburns would like another rounds of review

PR #526, opened 2019-03-12
add new extended module

adohe:
@brendandburns @yue9944882 open a separate PR to add new extended module


extended/pom.xml
```diff
@@ -0,0 +1,85 @@
...
+                        <goals>
+                            <goal>check</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+    <properties>
+        <java.version>1.7</java.version>
```
yue9944882: why not `8`? aligning w/ the util module might be better

extended/pom.xml
```diff
@@ -0,0 +1,85 @@
...
+    <name>client-java-extended</name>
+    <version>5.0.0-SNAPSHOT</version>
+    <url>https://github.com/kubernetes-client/java</url>
+
+    <parent>
+        <artifactId>client-java-parent</artifactId>
+        <groupId>io.kubernetes</groupId>
+        <version>5.0.0-SNAPSHOT</version>
+    </parent>
+
```
yue9944882:
```suggestion
<dependencies>
        <dependency>
            <groupId>io.kubernetes</groupId>
            <artifactId>client-java-api</artifactId>
            <version>${project.version}</version>
        </dependency>
        <dependency>
            <groupId>io.kubernetes</groupId>
            <artifactId>client-java-proto</artifactId>
            <version>${project.version}</version>
        </dependency>
</dependencies>
```

better to add some common dependencies here beforehand to avoid conflict?

brendandburns: yep, makes sense to me.

adohe: done~ ptal

yue9944882:
ref #523

PR #527, opened 2019-03-13
Adding pagination utility

kondapally1989:
pagination tool  https://github.com/kubernetes-client/java/issues/523
@brendandburns  @yue9944882


util/src/main/java/io/kubernetes/client/util/PagerParams.java
```diff
@@ -0,0 +1,30 @@
+package io.kubernetes.client.util;
+
+public class PagerParams {
+
+  private Integer limit;
+  private String _continue;
+
+  public PagerParams() {}
```
brendandburns: delete

util/src/test/java/io/kubernetes/client/pager/PagerTest.java
```diff
@@ -0,0 +1,90 @@
...
+      Resources.getResource("namespace-list.json").getPath();
+  private static final int PORT = 8087;
+  @Rule public WireMockRule wireMockRule = new WireMockRule(PORT);
+
+  @Before
+  public void setup() throws IOException {
+    client = new ClientBuilder().setBasePath("http://localhost:" + PORT).build();
+  }
+
+  @Test
```
brendandburns: Let's add/expand the tests so that they actually cover the pagination.

util/src/main/java/io/kubernetes/client/pager/Pager.java
```diff
@@ -0,0 +1,107 @@
...
+  private Integer limit;
+  private ApiClient client;
+  private Call call;
+  private Type listType;
+  private Function<PagerParams, Call> listFunc;
+
+  /**
+   * Pagination in kubernetes list call depends on continue and limit variable
+   *
+   * @param listFunc
```
brendandburns: Either document the parameters, or delete these. (and below)

util/src/main/java/io/kubernetes/client/pager/Pager.java
```diff
@@ -0,0 +1,107 @@
+package io.kubernetes.client.pager;
```
brendandburns: Add copyright boilerplate here.

util/src/main/java/io/kubernetes/client/util/PagerParams.java
```diff
@@ -0,0 +1,30 @@
+package io.kubernetes.client.util;
```
brendandburns: Add boilerplate here.

brendandburns: It's weird that this is in a different package than `Pager`. move to `pager`? or `util.pager`

util/src/main/java/io/kubernetes/client/util/PagerParams.java
```diff
@@ -0,0 +1,30 @@
...
+
+public class PagerParams {
+
+  private Integer limit;
```
brendandburns: nit, I prefer `int` to `Integer` but I don't feel strongly.

util/src/main/java/io/kubernetes/client/util/PagerParams.java
```diff
@@ -0,0 +1,30 @@
...
+  private String _continue;
```
brendandburns: nit: remove underscore.

util/src/main/java/io/kubernetes/client/util/PagerParams.java
```diff
@@ -0,0 +1,30 @@
...
+
+  public void setLimit(Integer limit) {
+    this.limit = limit;
+  }
+
+  public String getContinue() {
+    return _continue;
+  }
+
+  public void setContinue(String _continue) {
```
brendandburns: remove underscore.

brendandburns:
This is great!  Some small-ish comments.

util/src/main/java/io/kubernetes/client/util/PagerParams.java
yue9944882: isn't `continue` a reserved key word?

util/src/main/java/io/kubernetes/client/pager/Pager.java
```diff
@@ -0,0 +1,107 @@
...
+import io.kubernetes.client.ApiException;
+import io.kubernetes.client.models.V1ListMeta;
+import io.kubernetes.client.util.PagerParams;
+import io.kubernetes.client.util.Reflect;
+import io.kubernetes.client.util.exception.ObjectMetaReflectException;
+import java.io.IOException;
+import java.lang.reflect.Type;
+import java.util.function.Function;
+
+public class Pager {
```
yue9944882: if we don't want to make the pager shared by multiple resource types, better make it type parameterized.

```suggestion
public class Pager<ApiType, ApiListType> {
```

util/src/main/java/io/kubernetes/client/pager/Pager.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+    return Boolean.TRUE;
+  }
+
+  /**
+   * returns next chunk of List. size of list depends on limit set in constructor.
+   *
+   * @return Object
+   */
+  public <T> T next() {
```
yue9944882: nit: s/\<ApiType\>/\<T\>/g

util/src/test/java/io/kubernetes/client/pager/PagerTest.java
yue9944882: +1. add let's add test when the list receives a `400` due to invalid `limit` or `continue`, maybe expecting a checked exception here?

examples/src/main/java/io/kubernetes/client/examples/PagerExample.java
```diff
@@ -0,0 +1,76 @@
...
+ * <p>Easiest way to run this: mvn exec:java
+ * -Dexec.mainClass="io.kubernetes.client.examples.PagerExample"
+ *
+ * <p>From inside $REPO_DIR/examples
+ */
+public class PagerExample {
+  public static void main(String[] args) throws IOException {
+
+    ApiClient client = Config.defaultClient();
+    client.getHttpClient().setReadTimeout(60, TimeUnit.SECONDS);
```
yue9944882: (might be an unrelated question) what's the point of this?

util/src/main/java/io/kubernetes/client/util/PagerParams.java
yue9944882: i'm fine w/ either. a zero `int` and a null `Integer` actually works the same. 

[https://github.com/kubernetes/kubernetes/blob/d5a3db003916b1d33b503ccd2e4897e094d8af77/staging/src/k8s.io/apiserver/pkg/storage/etcd3/store.go#L494-L497](https://github.com/kubernetes/kubernetes/blob/d5a3db003916b1d33b503ccd2e4897e094d8af77/staging/src/k8s.io/apiserver/pkg/storage/etcd3/store.go#L494-L497)

util/src/main/java/io/kubernetes/client/util/PagerParams.java
```diff
@@ -0,0 +1,30 @@
...
+  public PagerParams(Integer limit, String _continue) {
+    this.limit = limit;
+    this._continue = _continue;
+  }
+
+  public Integer getLimit() {
+    return limit;
+  }
+
+  public void setLimit(Integer limit) {
```
yue9944882: how about making the `limit` an immutable value here? (which is, settable by constructor but didn't have setter/getter). otherwise, what's the use-case of changing page size dynamically?

util/src/main/java/io/kubernetes/client/util/PagerParams.java
brendandburns: good point, but perhaps `continueToken`? underscore is pretty unusual in Java variable names...

util/src/main/java/io/kubernetes/client/util/PagerParams.java
brendandburns: continueToken as above?

util/src/main/java/io/kubernetes/client/pager/Pager.java
kondapally1989: added.

util/src/main/java/io/kubernetes/client/util/PagerParams.java
kondapally1989: changed to continueToken.

util/src/main/java/io/kubernetes/client/util/PagerParams.java
kondapally1989: changed to continueToken.

util/src/main/java/io/kubernetes/client/pager/Pager.java
kondapally1989: added

util/src/main/java/io/kubernetes/client/util/PagerParams.java
kondapally1989: moved to pager package.

kondapally1989: added

util/src/test/java/io/kubernetes/client/pager/PagerTest.java
kondapally1989: added 2 more tests for failure cases.

examples/src/main/java/io/kubernetes/client/examples/PagerExample.java
brendandburns: This just prevents timeouts for large bodies/slow connections. I think the default is 10 seconds?

It could be that this is cargo-cult copied from other examples :)

brendandburns:
This is good enough for now, I'd still like to see a little more coverage in the unit tests, but we can defer to an additional PR.

/lgtm
/approve

brendandburns:
@yue9944882 @adohe @kondapally1989 

Let's move this into the `extended` module (separate PR after this merges is fine)

PR #538, opened 2019-03-27
Leader Election #523

atttx123:
leader election #523 
@yue9944882 @brendandburns 

by the way, i'm not sure how to write the unit test code for the leader election.

### DIFF
--- a/examples/src/main/java/io/kubernetes/client/examples/LeaderElectionExample.java
+++ b/examples/src/main/java/io/kubernetes/client/examples/LeaderElectionExample.java
@@ -1,0 +1,68 @@
+package io.kubernetes.client.examples;
+
+import io.kubernetes.client.leaderelection.LeaderCallbacks;
+import io.kubernetes.client.leaderelection.LeaderElectionConfig;
+import io.kubernetes.client.leaderelection.LeaderElector;
+import io.kubernetes.client.leaderelection.resourcelock.ConfigMapLock;
+import io.kubernetes.client.leaderelection.resourcelock.ResourceLockConfig;
+import io.kubernetes.client.util.Config;
+
+import java.io.IOException;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * A simple example of how to use the LeaderElection Java API
+ *
+ * @author yu
+ */
+public class LeaderElectionExample {
+    public static void main(String[] args) throws IOException {
+        final String name = "my-cm-lock";
+        final String namespace = "kube-system";
+
+        ResourceLockConfig lockConfig = new ResourceLockConfig();
+        lockConfig.setIdentity(String.format("%s-%s", name, System.currentTimeMillis()));
+        ConfigMapLock lock = new ConfigMapLock(name, namespace, Config.defaultClient(), lockConfig);
+
+        LeaderCallbacks callbacks = new LeaderCallbacks() {
+            @Override
+            public void onStartedLeading() {
+                System.out.println("onStartedLeading");
+                try {
+                    Thread.sleep(TimeUnit.SECONDS.toMillis(86400));
+                } catch (InterruptedException e) {
+                    e.printStackTrace();
+                }
+            }
+
+            @Override
+            public void onStopLeading() {
+                System.out.println("onStopLeading");
+            }
+
+            @Override
+            public void onNewLeader(String identity) {
+                System.out.println("onNewLeader identity: " + identity);
+                try {
+                    Thread.sleep(TimeUnit.SECONDS.toMillis(86400));
+                } catch (InterruptedException e) {
+                    e.printStackTrace();
+                }
+            }
+        };
+        LeaderElectionConfig electionConfig = new LeaderElectionConfig(
+                lock,
+                TimeUnit.SECONDS.toMillis(30),
+                TimeUnit.SECONDS.toMillis(15),
+                TimeUnit.SECONDS.toMillis(5),
+                callbacks,
+                true,
+                name);
+        LeaderElector leaderElector = new LeaderElector(electionConfig);
+        try {
+            leaderElector.run();
+        } finally {
+            leaderElector.cancel();
+        }
+    }
+}
--- a/util/src/main/java/io/kubernetes/client/leaderelection/LeaderElectionConfig.java
+++ b/util/src/main/java/io/kubernetes/client/leaderelection/LeaderElectionConfig.java
@@ -1,0 +1,146 @@
+package io.kubernetes.client.leaderelection;
+
+import io.kubernetes.client.leaderelection.resourcelock.Lock;
+
+/**
+ * Created by IDEA on 2019-03-24 16:24
+ *
+ * @author yu
+ */
+public class LeaderElectionConfig {
+    /**
+     * Lock is the resource that will be used for locking
+     */
+    private Lock lock;
+    /**
+     * LeaseDuration is the duration that non-leader candidates will
+     * wait to force acquire leadership. This is measured against time of
+     * last observed ack.
+     */
+    private long leaseDuration;
+    /**
+     * RenewDeadline is the duration that the acting master will retry
+     * refreshing leadership before giving up.
+     */
+    private long renewDeadline;
+    /**
+     * RetryPeriod is the duration the LeaderElector clients should wait
+     * between tries of actions.
+     */
+    private long retryPeriod;
+    /**
+     * Callbacks are callbacks that are triggered during certain lifecycle
+     * events of the LeaderElector
+     */
+    private LeaderCallbacks callbacks;
+    /**
+     * ReleaseOnCancel should be set true if the lock should be released
+     * when the run context is cancelled. If you set this to true, you must
+     * ensure all code guarded by this lease has successfully completed
+     * prior to cancelling the context, or you may have two processes
+     * simultaneously acting on the critical path.
+     */
+    private boolean releaseOnCancel;
+    /**
+     * Name is the name of the resource lock for debugging
+     */
+    private String name;
+
+    public LeaderElectionConfig(
+            Lock lock, long leaseDuration, long renewDeadline, long retryPeriod, LeaderCallbacks callbacks,
+            boolean releaseOnCancel, String name) {
+        this.lock = lock;
+        this.leaseDuration = leaseDuration;
+        this.renewDeadline = renewDeadline;
+        this.retryPeriod = retryPeriod;
+        this.callbacks = callbacks;
+        this.releaseOnCancel = releaseOnCancel;
+        this.name = name;
+    }
+
+    /**
+     * Lock is the resource that will be used for locking
+     */
+    public Lock getLock() {
+        return lock;
+    }
+
+    public void setLock(Lock lock) {
+        this.lock = lock;
+    }
+
+    /**
+     * LeaseDuration is the duration that non-leader candidates will
+     * wait to force acquire leadership. This is measured against time of
+     * last observed ack.
+     */
+    public long getLeaseDuration() {
+        return leaseDuration;
+    }
+
+    public void setLeaseDuration(long leaseDuration) {
+        this.leaseDuration = leaseDuration;
+    }
+
+    /**
+     * RenewDeadline is the duration that the acting master will retry
+     * refreshing leadership before giving up.
+     */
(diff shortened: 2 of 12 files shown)


util/src/main/java/io/kubernetes/client/leaderelection/resourcelock/ResourceLockConfig.java
```diff
@@ -0,0 +1,40 @@
...
+public class ResourceLockConfig {
+    /**
+     * Identity is the unique string identifying a lease holder across
+     * all participants in an election.
+     */
+    private String identity;
+    /**
+     * EventRecorder is optional.
+     */
+    private EventRecorder eventRecorder;
```
yue9944882: remove this. we should design/implement event utilities in another pull. i updated the issue to reflect the task.

util/src/main/java/io/kubernetes/client/leaderelection/resourcelock/ResourceLock.java
```diff
@@ -0,0 +1,178 @@
...
+        this.namespace = namespace;
+        this.client = client;
+        this.apiVersion = apiVersion;
+        this.kind = kind;
+        this.type = type;
+        this.resourceLockConfig = resourceLockConfig;
+    }
+
+    private String getRootUrl() {
+        return String.format("/api/%s/namespaces/%s/%s",
```
yue9944882: we're introducing `coordination` api in the future releases, so it doesn't have to be a legacy`/api` path. dealing w/ format printing url here doesn't make sense. better make it an interface or lambda returning an (generic?) object

util/src/main/java/io/kubernetes/client/leaderelection/resourcelock/ResourceLock.java
```diff
@@ -0,0 +1,178 @@
...
+
+    private final ResourceLockConfig resourceLockConfig;
+    private JsonObject object;
+
+    public ResourceLock(String name, String namespace, ApiClient client, String apiVersion, String kind, String type,
+                        ResourceLockConfig resourceLockConfig) {
+        this.name = name;
+        this.namespace = namespace;
+        this.client = client;
+        this.apiVersion = apiVersion;
```
yue9944882: we don't have a second version for `/api` since that it's a legacy path

examples/src/main/java/io/kubernetes/client/examples/LeaderElectionExample.java
```diff
@@ -0,0 +1,68 @@
...
+                TimeUnit.SECONDS.toMillis(15),
+                TimeUnit.SECONDS.toMillis(5),
+                callbacks,
+                true,
+                name);
+        LeaderElector leaderElector = new LeaderElector(electionConfig);
+        try {
+            leaderElector.run();
+        } finally {
+            leaderElector.cancel();
```
yue9944882: sorry to be dense. what's the point of cancelling if the `elector.run` already returned?

util/src/test/java/io/kubernetes/client/util/leaderelection/LeaderElectionTest.java
```diff
@@ -0,0 +1,70 @@
...
+import io.kubernetes.client.leaderelection.resourcelock.ConfigMapLock;
+import io.kubernetes.client.leaderelection.resourcelock.ResourceLockConfig;
+import io.kubernetes.client.util.Config;
+import org.junit.Test;
+
+import java.io.IOException;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Created by IDEA on 2019-03-27 16:06
```
yue9944882: nit: minimize docs

yue9944882:
and btw, cncf seems unhappy

/check-cla

atttx123:
I signed it

examples/src/main/java/io/kubernetes/client/examples/LeaderElectionExample.java
atttx123: i just fellow the logic in [`leaderelection.go`](https://github.com/kubernetes/client-go/blob/master/tools/leaderelection/leaderelection.go#L172),  the `Run` function will behave like this:
* block the main thread
* if itself get the leader lock, run a goroutine for `OnStartedLeading`
* if others get the leader lock, this lock will fail into a while sleep loop, unit itself get the leader lock
so, the `leaderElector.run()` will block the thread

util/src/main/java/io/kubernetes/client/leaderelection/resourcelock/ResourceLockConfig.java
atttx123: `EventRecorder`is just a interface for ResourceLock to record the event, 
[`client-go/leaderelection/resroucelock/interface.go`](https://github.com/kubernetes/client-go/blob/master/tools/leaderelection/resourcelock/interface.go#L53),

yue9944882: don't design it if it doesn't intend to use it

examples/src/main/java/io/kubernetes/client/examples/LeaderElectionExample.java
yue9944882: `cancel` here seems supposed to called in another thread. what'll read the boolean if `run` already exit?

atttx123: yes，`cancel()` set  an AtomicBoolean like a stop channel in golang

util/src/main/java/io/kubernetes/client/leaderelection/resourcelock/ResourceLockConfig.java
atttx123: OK, i will delete this interface.

util/src/main/java/io/kubernetes/client/leaderelection/resourcelock/ResourceLock.java
atttx123: Good idea, i will change this and add more unit test, it will take a time to complete

atttx123: there is a not good design in `k8s-java-client`, because all k8s generated resource type  do not have an  common parent class. if `ConfigMap` and `Endpoint` extends `HashedMetaObject`, then this interface will be more meaningful.

atttx123:
@yue9944882 
sorry to be dense. how can i solve the `Non complying file` error ?

util/src/main/java/io/kubernetes/client/leaderelection/LeaderElectionConfig.java
```diff
@@ -0,0 +1,150 @@
...
+
+  /**
+   * Lock is the resource that will be used for locking
+   */
+  private Lock lock;
+  /**
+   * LeaseDuration is the duration that non-leader candidates will wait to force acquire leadership.
+   * This is measured against time of last observed ack.
+   */
+  private long leaseDuration;
```
yue9944882: `java.time.Duration` looks better?

util/src/main/java/io/kubernetes/client/leaderelection/LeaderElector.java
```diff
@@ -0,0 +1,315 @@
...
+        try {
+          Callable<Boolean> callable = () -> {
+            boolean done = false;
+            while (!done && !cancel.get()) {
+              done = tryAcquireOrRenew();
+              Thread.sleep(config.getRetryPeriod());
+            }
+            return done;
+          };
+          executor.invokeAll(
```
yue9944882:
```suggestion
          executor.scheduleAtFixedRate(() -> {
              // ... tryAcquireOrRenew
          }, 0, getRetryPeriod().toMillis(), TimeUnit.MILLISECONDS)
```

util/src/main/java/io/kubernetes/client/leaderelection/LeaderElector.java
```diff
@@ -0,0 +1,315 @@
...
+  private Long observedTime;
+  /**
+   * used to implement OnNewLeader(), may lag slightly from the value observedRecord.HolderIdentity
+   * if the transition has not yet been reported.
+   */
+  private String reportedLeader;
+
+  private AtomicBoolean cancel = new AtomicBoolean(false);
+
+  private ExecutorService executor;
```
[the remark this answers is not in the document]
yue9944882: additionally, i think we should split it into two separate pools, one for running lease and one for hooks. the latter one should be a fixed pool at size 1.

util/src/main/java/io/kubernetes/client/leaderelection/LeaderElector.java
```diff
@@ -0,0 +1,315 @@
...
+
+  private Random random = new Random();
+  private Logger logger = LoggerFactory.getLogger(LeaderElector.class);
+
+  private LeaderElectionConfig config;
+  /**
+   * internal bookkeeping
+   */
+  private LeaderElectionRecord observedRecord;
+  private Long observedTime;
```
yue9944882: nit: makes it either a `Date` or a `long`(built-in)

util/src/main/java/io/kubernetes/client/leaderelection/LeaderElector.java
```diff
@@ -0,0 +1,315 @@
...
+        config.getLock().identity(),
+        (int) TimeUnit.MILLISECONDS.toSeconds(config.getLeaseDuration()), now, now);
+    LeaderElectionRecord oldLeaderElectionRecord;
+    try {
+      // 1. obtain or create the ElectionRecord
+      oldLeaderElectionRecord = config.getLock().get();
+    } catch (ResourceNotFoundException e) {
+      try {
+        config.getLock().create(leaderElectionRecord);
+      } catch (ApiException e1) {
```
yue9944882: a better name for `e1`. we don't like to do extra reverse engineering work, do we?

util/src/main/java/io/kubernetes/client/leaderelection/resourcelock/ResourceLock.java
yue9944882: >  if ConfigMap and Endpoint extends HashedMetaObject, then this interface will be more meaningful.

we have limitted kinds of resources for leader-electing. i presume just multiple independant resource lock classes is fine. separated class like `ConfigMapResourceLock`, `EndpointsResourceLock`, `CoordinationResourceLock`..

@brendandburns for thoughts

util/src/main/java/io/kubernetes/client/leaderelection/LeaderElectionRecord.java
```diff
@@ -0,0 +1,133 @@
...
+   * leases with empty identities and will wait for the full lease interval to expire before
+   * attempting to reacquire. This value is set to empty when a client voluntarily steps down.
+   */
+  private String holderIdentity;
+  private int leaseDurationSeconds;
+  private String acquireTime;
+  private String renewTime;
+  private int leaderTransitions = 0;
+
+  public LeaderElectionRecord() {
```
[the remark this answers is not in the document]
atttx123: `LeaderElectionRecord ` means record saved in crd.metadata.annotations.

util/src/main/java/io/kubernetes/client/leaderelection/LeaderElector.java
atttx123: they are two different functions:
```java
invokeAll(Collection<? extends Callable<T>> tasks, long timeout, TimeUnit unit)
scheduleAtFixedRate(Runnable command, long initialDelay, long period, TimeUnit unit)
```

util/src/main/java/io/kubernetes/client/leaderelection/LeaderElectionRecord.java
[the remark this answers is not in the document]
yue9944882: please remove the unused non-param constructor

util/src/main/java/io/kubernetes/client/leaderelection/LeaderElector.java
yue9944882: isn't it better if we retry w/ the help of a scheduled thread pool?

yue9944882:
/close

closing in favor of #590

PR #543, opened 2019-04-01
Refactor pager to support colon-loop java grammar

yue9944882:
Implement `Iterable` interface for pager so that we can loop like:

```java
for (V1Pod pod: pager) {
    // ...
}
```


util/src/main/java/io/kubernetes/client/pager/Pager.java
```diff
@@ -64,43 +74,67 @@ public Boolean hasNext() {
...
-  }
-
-  /**
-   * returns next chunk of list. size of list depends on limit set in constructor or nextLimit.
-   *
-   * @param nextLimit
-   * @return
-   */
-  public <ApiType> ApiListType next(Integer nextLimit) {
+  @Override
```
brendandburns: nit: newline here.

brendandburns:
/lgtm
/approve

(one small nit, we can address in cleanup sometime later)

PR #544, opened 2019-04-04
Fixes patch format to support various patch formats

yue9944882:
Fixes: #127

### DIFF
--- a/kubernetes/src/main/java/io/kubernetes/client/ApiClient.java
+++ b/kubernetes/src/main/java/io/kubernetes/client/ApiClient.java
@@ -73,6 +73,10 @@ public class ApiClient {
 
     private HttpLoggingInterceptor loggingInterceptor;
 
+    public static final String PATCH_FORMAT_JSON_PATCH = "application/json-patch+json";
+    public static final String PATCH_FORMAT_JSON_MERGE_PATCH = "application/merge-patch+json";
+    public static final String PATCH_FORMAT_STRATEGIC_MERGE_PATCH = "application/strategic-merge-patch+json";
+
     /*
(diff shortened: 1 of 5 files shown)


kubernetes/src/main/java/io/kubernetes/client/ApiClient.java
```diff
@@ -73,6 +73,10 @@
 
     private HttpLoggingInterceptor loggingInterceptor;
 
+    public static final String PATCH_FORMAT_JSON_PATCH = "application/json-patch+json";
+    public static final String PATCH_FORMAT_JSON_MERGE_PATCH = "application/merge-patch+json";
+    public static final String PATCH_FORMAT_STRATEGIC_MERGE_PATCH = "application/strategic-merge-patch+json";
+
```
kondapally1989: Shouldn't apiclient be not modified as its generated by swagger.

yue9944882: it was already modified and skipped by setting rule here

[https://github.com/kubernetes-client/java/blob/2e43cbe52c2379755653f18a29957f3896a6722e/kubernetes/.swagger-codegen-ignore#L11](https://github.com/kubernetes-client/java/blob/2e43cbe52c2379755653f18a29957f3896a6722e/kubernetes/.swagger-codegen-ignore#L11)

brendandburns: Yeah, we really shouldn't, but there were a bunch of fixes we needed...

We should try regenerating in case there are bug fixes we could pick up...

For now, this is fine.

PR #558, opened 2019-04-22
Feat: Provide thread-safety for pager

yue9944882:
/cc @brendandburns 

Before this pull, `Pager` is not thread-safe. By adding a private non-static iterator class to store the pagination progress, the pager can be shared across threads now. Refactored the test a bit for validation.

### DIFF
--- a/util/src/main/java/io/kubernetes/client/pager/Pager.java
+++ b/util/src/main/java/io/kubernetes/client/pager/Pager.java
@@ -15,7 +15,6 @@ package io.kubernetes.client.pager;
 import com.squareup.okhttp.Call;
 import io.kubernetes.client.ApiClient;
 import io.kubernetes.client.ApiException;
-import io.kubernetes.client.models.V1ListMeta;
 import io.kubernetes.client.util.Reflect;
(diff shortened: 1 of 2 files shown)


util/src/main/java/io/kubernetes/client/pager/Pager.java
```diff
@@ -137,4 +85,53 @@ public String getContinueToken() {
...
+    private boolean started;
+    private String continueToken;
+    private Call call;
+    private int offsetCurrentPage;
+    private int currentPageSize;
+
+    /**
+     * returns false if kubernetes server has exhausted List.
+     *
+     * @return
```
brendandburns: Either fill in more details or delete this. (and below too)

Let's include JavaDoc tags if we're using them, but delete them if not.

brendandburns:
One pretty minor comment. Then LGTM

util/src/main/java/io/kubernetes/client/pager/Pager.java
yue9944882: updated and filled docs

PR #569, opened 2019-05-14
Upgrade to sundrio 0.18.0

iocanel:


yue9944882:
hi @iocanel, could you expand more detail about the change? what's the purpose for bumping the dependency?

iocanel:
@yue9944882: Hi, my understanding is that the kubernetes-client is affected by: 
https://github.com/sundrio/sundrio/issues/124 among other bugs that have been fixed since then.

yue9944882:
thanks @iocanel, let's get this in

/lgtm
/approve

PR #572, opened 2019-05-23
Bump new 5.0.0-beta1 release version

yue9944882:
/cc @brendandburns

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2,7 +2,7 @@
   <modelVersion>4.0.0</modelVersion>
   <artifactId>client-java-parent</artifactId>
   <groupId>io.kubernetes</groupId>
-  <version>5.0.0-SNAPSHOT</version>
+  <version>5.0.0-beta1</version>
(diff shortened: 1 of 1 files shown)


brendandburns:
Don't we need this for the other sub-folder pom.xml files too?

brendandburns:
And do we want to skip straight to a full release instead of a beta? We're going to immediately cut 6.0.0-beta1 anyway to pick up the 1.14 changes...

yue9944882:
> Don't we need this for the other sub-folder pom.xml files too?

do we? the sub-modules are supposed will inherit from parent's version?

> And do we want to skip straight to a full release instead of a beta? We're going to immediately cut 6.0.0-beta1 anyway to pick up the 1.14 changes...

ohh tricked. it's 5.x works for 1.13 and 6.x for 1.14.

brendandburns:
Are the poms set up that way? I feel like the last time I changed this I had to do it everywhere...

PR #575, opened 2019-05-30
Fix a problem that lister cannot get resource correctly

cizezsy:


yue9944882:
/lgtm
/approve

that's awkward..

PR #155, opened 2017-12-18
Add timeout configuration to WatchExample - Closes #150

scholzj:

PR #576, opened 2019-05-30
Add timeout configuration to InformerExample

cizezsy:
Informer will retry list and watch when read timeout, better to add it to example...

### DIFF
--- a/examples/src/main/java/io/kubernetes/client/examples/InformerExample.java
+++ b/examples/src/main/java/io/kubernetes/client/examples/InformerExample.java
@@ -11,11 +11,20 @@ import io.kubernetes.client.models.V1NodeList;
 import io.kubernetes.client.models.V1ObjectMeta;
 import io.kubernetes.client.util.CallGeneratorParams;
 import io.kubernetes.client.util.Config;
+import java.util.concurrent.TimeUnit;
 
+/**
+ * A simple example of how to use the Java API
+ *
+ * <p>Easiest way to run this: mvn exec:java
+ * -Dexec.mainClass="io.kubernetes.client.examples.InformerExample"
+ *
+ * <p>From inside $REPO_DIR/examples
+ */
 public class InformerExample {
   public static void main(String[] args) throws Exception {
-
     ApiClient client = Config.defaultClient();
+    client.getHttpClient().setReadTimeout(60, TimeUnit.SECONDS);
     Configuration.setDefaultApiClient(client);
 
     SharedInformerFactory factory = new SharedInformerFactory();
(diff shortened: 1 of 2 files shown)


examples/src/main/java/io/kubernetes/client/examples/InformerExample.java
```diff
@@ -11,11 +11,20 @@
...
+ * <p>Easiest way to run this: mvn exec:java
+ * -Dexec.mainClass="io.kubernetes.client.examples.InformerExample"
+ *
+ * <p>From inside $REPO_DIR/examples
+ */
 public class InformerExample {
   public static void main(String[] args) throws Exception {
-
     ApiClient client = Config.defaultClient();
+    client.getHttpClient().setReadTimeout(60, TimeUnit.SECONDS);
```
yue9944882: it's reasonable to apply a timeout to the list calls, but the watches normally span 10-ish minutes from informer which is jitterr'd server-side. basically expecting an infinite client-side timeout is fine and that's the client works in k/client-go.

cizezsy: Yeah, but IMHO if we set an infinite client-side timeout, it might lead a dead connection and causes thread pool gets exhausted because we don't know whether the server is alive. How client-go resolve that?

Also, you'are right, 60 seconds is too short, I use this setting because WatchExample uses it :P

yue9944882:
```suggestion
    client.getHttpClient().setReadTimeout(0); // infinite timeout
```
oops.. am afraid the timeout settings in the watch example is misleading then from #155. that makes the watch runners shutdown in a minute-ish timeout and leads to frequent re-list which can potentially overloads the server. we should also explicitly set the timeout to `0` in the watch example. @brendandburns wdyt?

yue9944882: > How does client-go resolve this?

no timeout in client-side

cizezsy: OK, changed it, thanks~

examples/src/main/java/io/kubernetes/client/examples/InformerExample.java
```diff
@@ -11,11 +11,20 @@
...
+ * <p>Easiest way to run this: mvn exec:java
+ * -Dexec.mainClass="io.kubernetes.client.examples.InformerExample"
+ *
+ * <p>From inside $REPO_DIR/examples
+ */
 public class InformerExample {
   public static void main(String[] args) throws Exception {
-
     ApiClient client = Config.defaultClient();
+    client.getHttpClient().setReadTimeout(0, TimeUnit.SECONDS); // infinite timeout
```
yue9944882: apply the infinite timeout to the watch example

cizezsy: done

yue9944882:
thank you @cizezsy for fixing this

PR #577, opened 2019-05-31
Release 5.0

brendandburns:
@yue9944882 

release 5.0.0 has been pushed to Sonatype and should be duplicated to maven central in 24hr or so...

### DIFF
--- a/extended/pom.xml
+++ b/extended/pom.xml
@@ -11,7 +11,7 @@
     <parent>
         <artifactId>client-java-parent</artifactId>
         <groupId>io.kubernetes</groupId>
-        <version>5.0.0</version>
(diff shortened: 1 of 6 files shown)


extended/pom.xml
```diff
@@ -1,7 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0"
```
brendandburns: note, this change happened when maven rewrote the new pom file, I suggest we keep it to eliminate future diffs.

yue9944882: thanks! sgtm

as a follow-up, i'll do the openapi-generator migration and 1.14 api re-generation

/lgtm

PR #584, opened 2019-06-04
Chore: Update compatibility matrix version to 5.0.0

yue9944882:
/cc @brendandburns


brendandburns:
Thanks! Let's make sure a similar update is part of reving to 1.14 API too.

PR #586, opened 2019-06-06
Bugfix: Revise failing patch example

yue9944882:
the patch example is failing. the supported values for `dryRun` is [null, "All"]


yue9944882:
closing in favor of #587

PR #590, opened 2019-06-08
leader election

wu8685:
Leader election for #523
Please take a look @yue9944882

### DIFF
--- a/examples/src/main/java/io/kubernetes/client/examples/LeaderElectionExample.java
+++ b/examples/src/main/java/io/kubernetes/client/examples/LeaderElectionExample.java
@@ -1,0 +1,38 @@
+package io.kubernetes.client.examples;
+
+import io.kubernetes.client.ApiClient;
+import io.kubernetes.client.Configuration;
+import io.kubernetes.client.extended.leaderelection.LeaderElectionConfig;
+import io.kubernetes.client.extended.leaderelection.LeaderElector;
+import io.kubernetes.client.extended.leaderelection.resourcelock.EndpointsLock;
+import io.kubernetes.client.util.Config;
+import java.time.Duration;
+
+/**
+ * A simple example of how to use the Java API
+ *
+ * <p>Easiest way to run this: mvn exec:java
+ * -Dexec.mainClass="io.kubernetes.client.examples.LeaderElectionExample"
+ *
+ * <p>From inside $REPO_DIR/examples
+ */
+public class LeaderElectionExample {
+  public static void main(String[] args) throws Exception {
+    ApiClient client = Config.defaultClient();
+    Configuration.setDefaultApiClient(client);
+
+    EndpointsLock lock = new EndpointsLock("kube-system", "leader-election", "foo");
+
+    LeaderElectionConfig leaderElectionConfig =
+        new LeaderElectionConfig(lock, Duration.ofMillis(10000), null, Duration.ofMillis(5000));
+    LeaderElector leaderElector = new LeaderElector(leaderElectionConfig);
+
+    leaderElector.run(
+        () -> {
+          System.out.println("Do something when getting leadership.");
+        },
+        () -> {
+          System.out.println("Do something when losing leadership.");
+        });
+  }
+}
--- a/util/src/main/java/io/kubernetes/client/util/leaderelection/LeaderElectionRecord.java
+++ b/util/src/main/java/io/kubernetes/client/util/leaderelection/LeaderElectionRecord.java
@@ -1,0 +1,96 @@
+package io.kubernetes.client.util.leaderelection;
+
+import java.util.Date;
+
+public class LeaderElectionRecord {
+
+  private String holderIdentity;
+
+  private int leaseDurationSeconds;
+
+  private Date acquireTime;
+
+  private Date renewTime;
+
+  private int leaderTransitions;
+
+  public LeaderElectionRecord() {}
+
+  public LeaderElectionRecord(
+      String holderIdentity,
+      int leaseDurationSeconds,
+      Date acquireTime,
+      Date renewTime,
+      int leaderTransitions) {
+    this.holderIdentity = holderIdentity;
+    this.leaseDurationSeconds = leaseDurationSeconds;
+    this.acquireTime = acquireTime;
+    this.renewTime = renewTime;
+    this.leaderTransitions = leaderTransitions;
+  }
+
+  public String getHolderIdentity() {
+    return holderIdentity;
+  }
+
+  public void setHolderIdentity(String holderIdentity) {
+    this.holderIdentity = holderIdentity;
+  }
+
+  public int getLeaseDurationSeconds() {
+    return leaseDurationSeconds;
+  }
+
+  public void setLeaseDurationSeconds(int leaseDurationSeconds) {
+    this.leaseDurationSeconds = leaseDurationSeconds;
+  }
+
+  public Date getAcquireTime() {
+    return acquireTime;
+  }
+
+  public void setAcquireTime(Date acquireTime) {
+    this.acquireTime = acquireTime;
+  }
+
+  public Date getRenewTime() {
+    return renewTime;
+  }
+
+  public void setRenewTime(Date renewTime) {
+    this.renewTime = renewTime;
+  }
+
+  public int getLeaderTransitions() {
+    return leaderTransitions;
+  }
+
+  public void setLeaderTransitions(int leaderTransitions) {
+    this.leaderTransitions = leaderTransitions;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) return true;
+    if (!(o instanceof LeaderElectionRecord)) return false;
+
+    LeaderElectionRecord that = (LeaderElectionRecord) o;
+
+    if (leaseDurationSeconds != that.leaseDurationSeconds) return false;
+    if (leaderTransitions != that.leaderTransitions) return false;
+    if (holderIdentity == null) {
+      if (that.holderIdentity != null) return false;
+    } else {
+      if (!holderIdentity.equals(that.holderIdentity)) return false;
+    }
+    if (acquireTime == null) {
+      if (that.acquireTime != null) return false;
+    } else {
+      if (!acquireTime.equals(that.acquireTime)) return false;
+    }
+    if (renewTime == null) {
+      return that.holderIdentity == null;
+    }
+    return renewTime.equals(that.renewTime);
+  }
+}
--- a/util/src/main/java/io/kubernetes/client/util/leaderelection/LeaderElector.java
+++ b/util/src/main/java/io/kubernetes/client/util/leaderelection/LeaderElector.java
@@ -1,0 +1,216 @@
+package io.kubernetes.client.util.leaderelection;
+
+import io.kubernetes.client.ApiException;
+import java.net.HttpURLConnection;
+import java.time.Duration;
+import java.util.Date;
+import java.util.concurrent.*;
+import java.util.concurrent.atomic.AtomicBoolean;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class LeaderElector {
+
+  private static final Logger log = LoggerFactory.getLogger(LeaderElector.class);
+
+  private static final double JITTER_FACTOR = 1.2;
+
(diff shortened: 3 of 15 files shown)


util/src/main/java/io/kubernetes/client/util/leaderelection/LeaderElector.java
```diff
@@ -0,0 +1,210 @@
...
+import java.util.concurrent.*;
+import java.util.concurrent.atomic.AtomicBoolean;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class LeaderElector {
+
+  private static final Logger log = LoggerFactory.getLogger(LeaderElector.class);
+
+  private static final int STATUS_NOT_FOUND = 404;
```
brendandburns: use `HttpURLConnection.HTTP_NOT_FOUND` instead of defining your own...

util/src/main/java/io/kubernetes/client/util/leaderelection/LeaderElector.java
```diff
@@ -0,0 +1,210 @@
...
+  private ScheduledExecutorService scheduledWorkers = Executors.newSingleThreadScheduledExecutor();
+  private ExecutorService leaseWorkers = Executors.newSingleThreadExecutor();
+  private ExecutorService hookWorkers = Executors.newSingleThreadExecutor();
+
+  public LeaderElector(LeaderElectionConfig config) {
+    this.config = config;
+  }
+
+  public void run(Runnable startLeadingHook, Runnable stopLeadingHook) {
+    log.info("Start leader election with lock", config.getLock().describe());
```
brendandburns: I feel like info level for all of this will be a little spammy. Maybe debug log level?

util/src/main/java/io/kubernetes/client/util/leaderelection/LeaderElector.java
```diff
@@ -0,0 +1,210 @@
...
+    try {
+      if (!acquire()) {
+        // Fail to acquire leadership
+        return;
+      }
+      log.info("Successfully acquired lease, became leader");
+      // Hook on start leading
+      hookWorkers.submit(startLeadingHook);
+      renew();
```
brendandburns: It feels weird to me that this function loops forever... I'd rather see the `while` loop here for clarity.

util/src/main/java/io/kubernetes/client/util/leaderelection/LeaderElector.java
```diff
@@ -0,0 +1,210 @@
...
+    } finally {
+      scheduledFuture.cancel(true);
+    }
+  }
+
+  private boolean tryAcquireOrRenew() {
+    Date now = new Date();
+    Lock lock = config.getLock();
+    LeaderElectionRecord leaderElectionRecord = new LeaderElectionRecord();
+    leaderElectionRecord.setHolderIdentity(lock.identity());
```
brendandburns: this feels odd. Why aren't these part of the constructor?

util/src/main/java/io/kubernetes/client/util/leaderelection/resourcelock/EndpointsLock.java
```diff
@@ -0,0 +1,115 @@
...
+import io.kubernetes.client.util.leaderelection.LeaderElectionRecord;
+import io.kubernetes.client.util.leaderelection.Lock;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicReference;
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class EndpointsLock implements Lock {
```
brendandburns: I feel like this duplicates a lot of code with the `ConfigMapLock` does it make sense to refactor a parent class for both of these? `AbstractResourceLock` or somesuch?

util/src/main/java/io/kubernetes/client/util/leaderelection/resourcelock/ConfigMapLock.java
```diff
@@ -0,0 +1,115 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicReference;
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ConfigMapLock implements Lock {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(ConfigMapLock.class);
```
brendandburns: I think we've used `log` elsewhere...

util/src/main/java/io/kubernetes/client/util/leaderelection/resourcelock/ConfigMapLock.java
```diff
@@ -0,0 +1,115 @@
...
+
+  // Namespace and name describes the configMap object
+  // that the LeaderElector will attempt to lead.
+  private final String namespace;
+  private final String name;
+  private final String identity;
+
+  private CoreV1Api coreV1Client = new CoreV1Api();
```
brendandburns: Probably want to give users the option to pass this in instead of using the default client?

util/src/main/java/io/kubernetes/client/util/leaderelection/resourcelock/ConfigMapLock.java
```diff
@@ -0,0 +1,115 @@
...
+  @Override
+  public boolean update(LeaderElectionRecord record) {
+    try {
+      V1ConfigMap configMap = configMapRefer.get();
+      configMap
+          .getMetadata()
+          .putAnnotationsItem(
+              LeaderElectionRecordAnnotationKey,
+              coreV1Client.getApiClient().getJSON().serialize(record));
+      V1ConfigMap replicedConfigMap =
```
brendandburns: `replacedConfigMap` ?

util/src/main/java/io/kubernetes/client/util/leaderelection/Lock.java
```diff
@@ -0,0 +1,16 @@
+package io.kubernetes.client.util.leaderelection;
+
+public interface Lock {
```
brendandburns: Since this is an interface, I think it needs more documentation.

examples/src/main/java/io/kubernetes/client/examples/LeaderElectionExample.java
```diff
@@ -0,0 +1,34 @@
...
+public class LeaderElectionExample {
+  public static void main(String[] args) throws Exception {
+    ApiClient client = Config.defaultClient();
+    Configuration.setDefaultApiClient(client);
+
+    CoreV1Api coreV1Api = new CoreV1Api(client);
+    EndpointsLock lock = new EndpointsLock("kube-system", "leader-election", "foo", coreV1Api);
+
+    LeaderElectionConfig leaderElectionConfig = new LeaderElectionConfig();
+    leaderElectionConfig.setLock(lock);
```
brendandburns: These feel like they should be in the constructor.

util/src/main/java/io/kubernetes/client/util/leaderelection/Lock.java
yue9944882: +1

util/src/main/java/io/kubernetes/client/util/leaderelection/Lock.java
```diff
@@ -0,0 +1,16 @@
...
+
+  String LeaderElectionRecordAnnotationKey = "control-plane.alpha.kubernetes.io/leader";
```
yue9944882: the annotation key should not be defined as a part of the interface, the underlying implementation of resource lock in k/l is migrating to `coordination.Lease` which provides corresponding api fields in the spec instead of annotation hacks.

util/src/main/java/io/kubernetes/client/util/leaderelection/Lock.java
```diff
@@ -0,0 +1,16 @@
...
+
+  LeaderElectionRecord get() throws Exception;
```
yue9944882: can this be an `ApiException`?

util/src/main/java/io/kubernetes/client/util/leaderelection/resourcelock/EndpointsLock.java
yue9944882: it's helping reducing code redundancy tho but having an `AbstractResourceLock` is a bit overkilling i suppose. for now, configmap and endpoints are the only supported resources and we aren't leaning to extend more. also, soon future on, the `coordination.Lease` API will be replacing the two.

util/src/main/java/io/kubernetes/client/util/leaderelection/Lock.java
brendandburns: +1,

util/src/main/java/io/kubernetes/client/util/leaderelection/LeaderElector.java
wu8685: How about to keep this kind of logs in info level, and change the logs inside election loop to debug level?

util/src/main/java/io/kubernetes/client/util/leaderelection/LeaderElector.java
```diff
@@ -0,0 +1,214 @@
...
+                }
+                acquired.set(success);
+              } catch (CancellationException e) {
+                log.info("Processing tryAcquireOrRenew successfully canceled");
+              } catch (Throwable t) {
+                future.cancel(true);
+              }
+            },
+            0,
+            Double.valueOf(retryPeriodMillis * (1.2 * Math.random() + 1)).longValue(),
```
yue9944882: const JITTER_FACTOR = 1.2

util/src/main/java/io/kubernetes/client/util/leaderelection/LeaderElector.java
yue9944882: how about adding a comment here: 

> // renew() won't return until it fails acquiring/renewing the lease

util/src/main/java/io/kubernetes/client/util/leaderelection/LeaderElector.java
```diff
@@ -0,0 +1,214 @@
...
+    }
+  }
+
+  private boolean acquire() {
+    if (log.isDebugEnabled()) {
+      log.debug("Attempting to acquire leader lease...");
+    }
+    Duration retryPeriod = config.getRetryPeriod();
+    long retryPeriodMillis = retryPeriod.toMillis();
+    AtomicBoolean acquired = new AtomicBoolean(false);
```
yue9944882: it doesnt have to be an atomic value

util/src/main/java/io/kubernetes/client/util/leaderelection/LeaderElector.java
wu8685: I changed the method name to `renewLoop`. Is it better?

util/src/main/java/io/kubernetes/client/util/leaderelection/LeaderElectionRecord.java
```diff
@@ -0,0 +1,101 @@
...
+  public int getLeaderTransitions() {
+    return leaderTransitions;
+  }
+
+  public void setLeaderTransitions(int leaderTransitions) {
+    this.leaderTransitions = leaderTransitions;
+  }
+
+  @Override
+  public int hashCode() {
```
yue9944882: can we remove this?

util/src/main/java/io/kubernetes/client/util/leaderelection/resourcelock/EndpointsLock.java
```diff
@@ -0,0 +1,120 @@
...
+
+  private CoreV1Api coreV1Client;
+
+  private AtomicReference<V1Endpoints> endpointsRefer = new AtomicReference<>(null);
+
+  public EndpointsLock(String namespace, String name, String identity) {
+    this.namespace = namespace;
+    this.name = name;
+    this.identity = identity;
+    this.coreV1Client = new CoreV1Api();
```
yue9944882:
```java
this(namespace, name, identity, Configuration.getDefaultApiClient())
```

PR #587, opened 2019-06-06
Feat: Structurelize patch object

yue9944882:
@brendandburns we saw that many developers are trying to pass an object directly into the patch call, making this structurelized can prevent them from that. additionally, this pull moves the patch format consts into the custom.V1Patch, which makes it easier to find

PR #594, opened 2019-06-14
Follow-up of #587: addressing minor comments

yue9944882:
ref: #587


brendandburns:
Thanks for following up on this!

/lgtm
/approve

Issue #596, opened 2019-06-14
Some files are missing from maven-deployed sources


victornoel:
Hi,

There seems to some missing files in the deployed sources to maven.

For example `io.kubernetes.client.ApiException` is provided by `client-java-5.0.0.jar` but is missing from `client-java-5.0.0-sources.jar`.

I don't know if there are others.

yue9944882:
@victornoel `ApiException` lives in `client-java-api` instead of `client-java`

victornoel:
@yue9944882 then there is a problem with the deployed maven artifacts: `io.kubernetes.client.ApiException` is present in `client-java-5.0.0.jar` also, so in Eclipse, m2e tries to look for its source in `client-java-5.0.0-sources.jar`.

```
➜  ~ unzip -l ~/.m2/repository/io/kubernetes/client-java/5.0.0/client-java-5.0.0.jar | grep ApiException
     3182  2019-05-31 04:49   io/kubernetes/client/ApiException.class
➜  ~ unzip -l ~/.m2/repository/io/kubernetes/client-java-api/5.0.0/client-java-api-5.0.0.jar | grep ApiException 
     3182  2019-05-31 04:49   io/kubernetes/client/ApiException.class

```

yue9944882:
i suppose it's caused by maven bundle packaging plugins.. i'm not really familiar w/ how OSGi packaging works but i sent #597 to strip the duplicated class. it's working fine locally.

victornoel:
@yue9944882 nice, thx!

PR #597, opened 2019-06-14
Chore: Exclude embeded dependencies from bundle packaging

yue9944882:
Fixes: #596


yue9944882:
i see warnings from my IDE tooling after this change

> The package is not exported by the bundle dependencies

hopefully this is okay

PR #606, opened 2019-07-01
Fix command in README

davidxia:
The `GEN_ROOT` env var already has the `gen` path segment.


brendandburns:
Thanks for the PR, but the current doc is actually correct given the previous `git clone ...` command.

I'm definitely open to more clear instructions either way so that no one gets confused...

Thanks!

davidxia:
ah yes, my mistake

PR #618, opened 2019-07-11
serverside apply & example

runzexia:
Signed-off-by: runzexia <runzexia@yunify.com>

### DIFF
--- a/examples/src/main/java/io/kubernetes/client/examples/PatchExample.java
+++ b/examples/src/main/java/io/kubernetes/client/examples/PatchExample.java
@@ -47,6 +47,8 @@ public class PatchExample {
       "{\"metadata\":{\"$deleteFromPrimitiveList/finalizers\":[\"example.com/test\"]}}";
   static String jsonDeploymentStr =
       "{\"kind\":\"Deployment\",\"apiVersion\":\"extensions/v1beta1\",\"metadata\":{\"name\":\"hello-node\",\"finalizers\":[\"example.com/test\"],\"labels\":{\"run\":\"hello-node\"}},\"spec\":{\"replicas\":1,\"selector\":{\"matchLabels\":{\"run\":\"hello-node\"}},\"template\":{\"metadata\":{\"creationTimestamp\":null,\"labels\":{\"run\":\"hello-node\"}},\"spec\":{\"terminationGracePeriodSeconds\":30,\"containers\":[{\"name\":\"hello-node\",\"image\":\"hello-node:v1\",\"ports\":[{\"containerPort\":8080}],\"resources\":{}}]}},\"strategy\":{}},\"status\":{}}";
+  static String applyYamlStr =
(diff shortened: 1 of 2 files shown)


examples/src/main/java/io/kubernetes/client/examples/PatchExample.java
```diff
@@ -47,6 +47,8 @@
       "{\"metadata\":{\"$deleteFromPrimitiveList/finalizers\":[\"example.com/test\"]}}";
   static String jsonDeploymentStr =
       "{\"kind\":\"Deployment\",\"apiVersion\":\"extensions/v1beta1\",\"metadata\":{\"name\":\"hello-node\",\"finalizers\":[\"example.com/test\"],\"labels\":{\"run\":\"hello-node\"}},\"spec\":{\"replicas\":1,\"selector\":{\"matchLabels\":{\"run\":\"hello-node\"}},\"template\":{\"metadata\":{\"creationTimestamp\":null,\"labels\":{\"run\":\"hello-node\"}},\"spec\":{\"terminationGracePeriodSeconds\":30,\"containers\":[{\"name\":\"hello-node\",\"image\":\"hello-node:v1\",\"ports\":[{\"containerPort\":8080}],\"resources\":{}}]}},\"strategy\":{}},\"status\":{}}";
+  static String applyYamlStr =
```
yue9944882: let's make it an true yaml string (instead of json content)

examples/src/main/java/io/kubernetes/client/examples/PatchExample.java
```diff
@@ -88,6 +90,16 @@ public static void main(String[] args) throws IOException {
                   null);
       System.out.println("strategic-merge-patched deployment" + deploy3);
 
+      // apply-yaml a deployment
```
yue9944882: explicitly note that server-side apply is yet alpha feature, and has to be enabled in the server

examples/src/main/java/io/kubernetes/client/examples/PatchExample.java
runzexia: Use yaml util for transformation? The header name is called apply yaml, but it is json when making a request to the server.

Or extend V1Patch? Let V1Patch support object, jsonString, yamlString

yue9944882: wait, the java client is encoding in json, but the server should receive yaml as its content-type is yaml.. does it really work? it'll take a shot locally

runzexia: I think that's just the header name?
https://github.com/kubernetes/kubernetes/blob/cb862e332329736dc1b893da6f757b808fa802aa/test/integration/apiserver/apply/apply_test.go#L269

brendandburns:
@runzexia is this PR still active?

Thanks!

examples/src/main/java/io/kubernetes/client/examples/PatchExample.java
yue9944882: after reaching out the maintainers working on this, i finally know that the so-called yaml here is actually including json. or rather, json is a valid yaml..

yue9944882:
/lgtm

brendan has approve

yue9944882:
getting this in

/approve

PR #595, opened 2019-06-14
[WIP] Migrating to openapi generator

yue9944882:

PR #619, opened 2019-07-12
Chore: Remove duplicated patch format from ApiClient

yue9944882:
/cc @brendandburns


brendandburns:
We should really start re-generating this file, it's going to cause problems...

/lgtm
/approve

yue9944882:
@brendandburns agree, previously i tried that in #595, it's safe to use the generated api-client since most fixes we expected has merged into the upstream.

PR #624, opened 2019-07-15
Add workqueue support

cizezsy:
Add workqueue support, see #523. Feel free to close this PR, if you think #525 is better :P

Will add more unit tests later.


yue9944882:
/cc @adohe

he's getting back on wednesday, let's hold until some input from him. by skimming the sketch, i find the rate-limiter enhancement compelling. do you consider port [the leaky-bucket-based rate-limiter](https://github.com/kubernetes/kubernetes/blob/b69cc78e5807e30ef9406e16e07fc87bd109cdcf/staging/src/k8s.io/client-go/util/workqueue/default_rate_limiters.go#L48-L63) to the tree?

cizezsy:
@yue9944882 Added it :P I will add more unit tests.

Also, the BucketRateLimiter is based on token bucket according to the documentation of rate.Limiter
>  It implements a "token bucket" of size b, initially full and refilled at rate r tokens per second.

extended/src/main/java/io/kubernetes/client/extended/workqueue/DelayingWorkQueue.java
```diff
@@ -0,0 +1,92 @@
+package io.kubernetes.client.extended.workqueue;
```
brendandburns: Why not just use `ScheduledThredPoolExecutor`?

https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ScheduledThreadPoolExecutor.html

I think it's better to use built-ins unless absolutely necessary...

extended/src/main/java/io/kubernetes/client/extended/workqueue/WorkQueue.java
```diff
@@ -0,0 +1,122 @@
...
+ *   <li>Multiple consumers and producers. In particular, it is allowed for an item to be reenqueued
+ *       while it is being processed.
+ *   <li>Shutdown notifications.
+ * </ul>
+ */
+public class WorkQueue<T> {
+  // queue defines the order in which we will work on items. Every
+  // element of queue should be in the dirty set and not in the
+  // processing set.
+  private LinkedList<T> queue;
```
brendandburns: Let's use a `BlockingQueue` here

(https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/BlockingQueue.html)

To limit the potential memory size.

extended/src/main/java/io/kubernetes/client/extended/workqueue/ratelimiter/RateLimiter.java
```diff
@@ -0,0 +1,23 @@
+package io.kubernetes.client.extended.workqueue.ratelimiter;
+
+import java.time.Duration;
+
+public interface RateLimiter<T> {
```
brendandburns: Why not use RateLimiter from Guava?

https://github.com/google/guava/blob/master/guava/src/com/google/common/util/concurrent/RateLimiter.java

brendandburns:
I put in some suggestions about existing libraries/classes you could use instead of re-implementing things...

extended/src/main/java/io/kubernetes/client/extended/workqueue/ratelimiter/RateLimiter.java
yue9944882: the guava rate-limiter's API is different from the one we want for work-queue here i suppose. guava's rate-limiter is basically expecting an invocation to be blocked until permitted, while for work-queue, the so-called `RateLimiter` here is more like a backoff-provider, it will return how much time an inbound item need to be waiting for next round of processing.. probably we should consider rename this to sth like `BackOffGetter`, or `BackOffFactory`? @brendandburns @cizezsy wdyt?

extended/src/main/java/io/kubernetes/client/extended/workqueue/WorkQueue.java
yue9944882: let's use the work-queue implementation from another pull #524

extended/src/main/java/io/kubernetes/client/extended/workqueue/DelayingWorkQueue.java
yue9944882: let's use the work-queue implementation from another pull #524

yue9944882:
>  you could use instead of re-implementing things...

+1 on we should avoid re-implementing wheels as much as possible

extended/src/main/java/io/kubernetes/client/extended/workqueue/ratelimiter/RateLimiter.java
cizezsy: Yeah, I have looked at Guava RateLimiter before, but it does not satisfy our requirement. 

Regard the name, maybe we should keep the name same with client-go's so that others can quickly know how to use client-java?

yue9944882:
@cizezsy another work-queue has got in, how do you think about rebase and replace the rate-limiter onto that?

cizezsy:
@yue9944882 ok, works for me :P

cizezsy:
@yue9944882 rebased and added some unit test :)

extended/src/main/java/io/kubernetes/client/extended/workqueue/ratelimiter/ItemExponentialFailureRateLimiter.java
```diff
@@ -0,0 +1,51 @@
...
+
+  public ItemExponentialFailureRateLimiter(Duration baseDelay, Duration maxDelay) {
+    this.baseDelay = baseDelay;
+    this.maxDelay = maxDelay;
+
+    failures = new HashMap<>();
+  }
+
+  @Override
+  public synchronized Duration when(T item) {
```
yue9944882:
```suggestion
  @Override
  public synchronized Duration when(T item) {
    int exp = failures.getOrDefault(item, 0);
    failures.put(item, exp + 1);
    long d = maxDelay.toMillis() >> exp;
    return d > baseDelay.toMillis() ? baseDelay.multipliedBy(1 << exp) : maxDelay;
  }
```

how do you think about this approach?

extended/src/main/java/io/kubernetes/client/extended/workqueue/ratelimiter/BucketRateLimiter.java
```diff
@@ -0,0 +1,56 @@
+package io.kubernetes.client.extended.workqueue.ratelimiter;
+
+import io.github.bucket4j.*;
+import java.time.Duration;
+
+/** A light-weight token bucket implementation for RateLimiter. */
+public class BucketRateLimiter<T> implements RateLimiter<T> {
```
yue9944882: thanks, the new bucketing library looks compelling. i tried simplify implemtation of `when` a bit learning from https://github.com/vladimir-bukhtoyarov/bucket4j/issues/96, wdyt?
 
```suggestion
public class BucketRateLimiter<T> implements RateLimiter<T> {

  private Bucket bucket;
  private long tokensInQueue;
  private long tokensGeneratedInPeriod;
  private Duration period;

  /**
   * @param capacity                Capacity is the maximum number of tokens can be consumed.
   * @param tokensGeneratedInPeriod Tokens generated in period.
   * @param period                  Period that generating specific number of tokens.
   */
  public BucketRateLimiter(long capacity, long tokensGeneratedInPeriod, Duration period) {
    Bandwidth bandwidth =
            Bandwidth.classic(capacity, Refill.greedy(tokensGeneratedInPeriod, period));

    this.bucket = Bucket4j.builder().
            addLimit(bandwidth).
            withSynchronizationStrategy(SynchronizationStrategy.SYNCHRONIZED).
            build();
  }

  @Override
  public synchronized Duration when(T item) {
    DelayGetter delayGetter = new DelayGetter();
    bucket.asAsyncScheduler().consume(1, delayGetter).complete(null);
    return delayGetter.getDelay();
  }

  @Override
  public void forget(T item) {
  }

  @Override
  public int numRequeues(T item) {
    return 0;
  }

  private class DelayGetter extends ScheduledThreadPoolExecutor {
    private Duration delay = Duration.ZERO;

    @Override
    public ScheduledFuture<?> schedule(Runnable command, long delay, TimeUnit unit) {
      this.delay = Duration.ofNanos(unit.toNanos(delay));
      return null;
    }

    public DelayGetter() {
      super(0);
    }

    public Duration getDelay() {
      return delay;
    }
  }
}

```

extended/src/main/java/io/kubernetes/client/extended/workqueue/ratelimiter/BucketRateLimiter.java
```diff
@@ -0,0 +1,56 @@
+package io.kubernetes.client.extended.workqueue.ratelimiter;
+
+import io.github.bucket4j.*;
```
adohe: please not using wild card in Java import statement, more info could be found here: https://stackoverflow.com/questions/147454/why-is-using-a-wild-card-with-a-java-import-statement-bad

extended/src/main/java/io/kubernetes/client/extended/workqueue/ratelimiter/ItemExponentialFailureRateLimiter.java
```diff
@@ -0,0 +1,51 @@
...
+    int exp = failures.getOrDefault(item, 0);
+    failures.put(item, exp + 1);
+
+    double backOff = baseDelay.toNanos() * Math.pow(2, exp);
+    if (backOff > Long.MAX_VALUE) {
```
brendandburns: Shouldn't this be:
`if (backOff > maxDelay) {` ...

Or alternately

`backoff = Math.min(backoff, maxDelay);`

extended/src/main/java/io/kubernetes/client/extended/workqueue/ratelimiter/ItemExponentialFailureRateLimiter.java
adohe: IMHO, Java has many built-in thread-safe data structure, try to avoid using synchronized method here, `java.util.concurrent.ConcurrentHashMap` should be a good fit for this case, and is much more high efficient.

extended/src/main/java/io/kubernetes/client/extended/workqueue/ratelimiter/ItemFastSlowRateLimiter.java
```diff
@@ -0,0 +1,48 @@
...
+  public ItemFastSlowRateLimiter(Duration fastDelay, Duration slowDelay, int maxFastAttempts) {
+    this.fastDelay = fastDelay;
+    this.slowDelay = slowDelay;
+    this.maxFastAttempts = maxFastAttempts;
+
+    failures = new HashMap<>();
+  }
+
+  @Override
+  public synchronized Duration when(T item) {
```
adohe: same as above, not using synchronized method here.

extended/src/main/java/io/kubernetes/client/extended/workqueue/ratelimiter/BucketRateLimiter.java
cizezsy: OK, will replace with single class imports.

extended/src/main/java/io/kubernetes/client/extended/workqueue/ratelimiter/BucketRateLimiter.java
cizezsy: oops, just realized my implementation is wrong :(

I will replace with your implementation, thanks!


withSynchronizationStrategy(SynchronizationStrategy.SYNCHRONIZED).

extended/src/main/java/io/kubernetes/client/extended/workqueue/ratelimiter/ItemExponentialFailureRateLimiter.java
cizezsy: I think `ConcurrentHashMap` is not thread-safe when we are doing multiple operations...

extended/src/main/java/io/kubernetes/client/extended/workqueue/ratelimiter/ItemExponentialFailureRateLimiter.java
cizezsy: Yeah, you are right, will change it :)

cizezsy:
Sorry for the delay, I am too busy in last week....

I have changed the implementation of `BucketRateLimiter` and `ItemExponentialFailureRateLimiter` as comment, and added some unit tests.

yue9944882:
LGTM, @brendandburns may want another round of review?

extended/src/main/java/io/kubernetes/client/extended/workqueue/ratelimiter/BucketRateLimiter.java
```diff
@@ -0,0 +1,64 @@
...
+    return 0;
+  }
+
+  private class DelayGetter extends ScheduledThreadPoolExecutor {
+    private Duration delay = Duration.ZERO;
+
+    @Override
+    public ScheduledFuture<?> schedule(Runnable command, long delay, TimeUnit unit) {
+      this.delay = Duration.ofNanos(unit.toNanos(delay));
+      return null;
```
brendandburns: Should this really return null?

extended/src/main/java/io/kubernetes/client/extended/workqueue/ratelimiter/ItemExponentialFailureRateLimiter.java
brendandburns: You can use a combination of `ConcurrentHashMap` and `AtomicInteger` here.

extended/src/main/java/io/kubernetes/client/extended/workqueue/ratelimiter/ItemFastSlowRateLimiter.java
brendandburns: again `ConcurrentHashMap` and `AtomicInteger` will provide what you need.

brendandburns:
A few additional comments.

extended/src/main/java/io/kubernetes/client/extended/workqueue/ratelimiter/BucketRateLimiter.java
cizezsy: @brendanburns 

The return value will not be used by Bucket4j according to the comment of https://github.com/vladimir-bukhtoyarov/bucket4j/issues/96, so I think it ok to return a null value.
>  ScheduledFuture that returned also not used by Bucket4j, so do not spend efforts to provide correctly result.

extended/src/main/java/io/kubernetes/client/extended/workqueue/ratelimiter/ItemExponentialFailureRateLimiter.java
cizezsy: OK, replaced with AtomicLongMap now.

extended/src/main/java/io/kubernetes/client/extended/workqueue/ratelimiter/ItemFastSlowRateLimiter.java
cizezsy: Replaced with AtomicLongMap.

extended/src/main/java/io/kubernetes/client/extended/workqueue/ratelimiter/BucketRateLimiter.java
```diff
@@ -0,0 +1,64 @@
...
+        Bandwidth.classic(capacity, Refill.greedy(tokensGeneratedInPeriod, period));
+    this.bucket =
+        Bucket4j.builder()
+            .addLimit(bandwidth)
+            .withSynchronizationStrategy(SynchronizationStrategy.SYNCHRONIZED)
+            .build();
+  }
+
+  @Override
+  public synchronized Duration when(T item) {
```
yue9944882:
```suggestion
  public Duration when(T item) {
```

no need for the modifier, isn't it?

cizezsy: Yes, will remove it.

yue9944882:
/lgtm
/approve
/hold

holding for lazy consensus with @brendandburns

PR #635, opened 2019-07-24
Out-of-box controller runtime framework and builder library

yue9944882:
Controller builder libraries.

### DIFF
--- a/extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
+++ b/extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
@@ -1,0 +1,225 @@
+package io.kubernetes.client.extended.controller;
+
+import io.kubernetes.client.extended.controller.reconciler.Reconciler;
+import io.kubernetes.client.extended.controller.reconciler.Request;
+import io.kubernetes.client.extended.controller.reconciler.Result;
+import io.kubernetes.client.extended.wait.Wait;
+import io.kubernetes.client.extended.workqueue.RateLimitingQueue;
+import java.time.Duration;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * The default implementation of a controller. A typical controller contains:
+ *
+ * <p>- a reconciler implemented by developers specifying what to do in reaction of notifications.
+ * see more documentation on the Reconciler interface - a work-queue continuously filled with task
+ * items managed by Informer framework - a set of worker threads acutally running reconciler.
+ */
+public class DefaultController implements Controller {
+  private static final Logger log = LoggerFactory.getLogger(DefaultController.class);
+
+  private Reconciler reconciler;
+  private String name;
+  private int workerCount;
+  private ScheduledExecutorService workerThreadPool;
+
+  private RateLimitingQueue<Request> workQueue;
+  private Supplier<Boolean>[] readyFuncs;
+
+  private Duration readyTimeout;
+  private Duration readyCheckInternal;
+
+  /**
+   * Instantiates a new Default controller.
+   *
+   * @param reconciler the reconciler
+   * @param workQueue the work queue
+   * @param readyFuncs the ready funcs
+   */
+  public DefaultController(
+      Reconciler reconciler,
+      RateLimitingQueue<Request> workQueue,
+      Supplier<Boolean>... readyFuncs) {
+    this.reconciler = reconciler;
+    this.workQueue = workQueue;
+    this.readyFuncs = readyFuncs;
+    this.readyTimeout = Duration.ofSeconds(30);
+    this.readyCheckInternal = Duration.ofSeconds(1);
+  }
+
+  // preFlightCheck checks if the controller is ready for working.
+  private boolean preFlightCheck() {
+    if (workerCount <= 0) {
+      log.error("Fail to start controller {}: worker count must be positive.");
+      return false;
+    }
+    if (workerThreadPool == null) {
+      log.error("Fail to start controller {}: missing worker thread-pool.");
+      return false;
+    }
+    if (!isReady()) {
+      log.error("Timed out waiting for cache to be synced.");
+      return false;
+    }
+    return true;
+  }
+
+  // isReady checks if any ready functions complaining before working.
+  private boolean isReady() {
+    boolean synced = true;
+    if (this.readyFuncs.length > 0) {
+      synced =
+          Wait.poll(
+              readyCheckInternal,
+              readyTimeout,
+              () -> {
+                boolean ready = true;
+                for (Supplier<Boolean> cacheHasSyncedFunc : this.readyFuncs) {
+                  ready = ready && cacheHasSyncedFunc.get();
+                }
+                return ready;
+              });
+    }
+    return synced;
+  }
+
+  @Override
+  public void run() {
+    if (!preFlightCheck()) {
+      log.error("Controller {} failed pre-run check, exiting..", this.name);
+      return;
+    }
+
+    // spawns worker threads for the controller.
+    CountDownLatch latch = new CountDownLatch(workerCount);
+    for (int i = 0; i < this.workerCount; i++) {
+      final int workerIndex = i;
+      workerThreadPool.scheduleWithFixedDelay(
+          () -> {
+            log.debug("Starting controller {} worker {}..", this.name, workerIndex);
+            this.worker();
+            latch.countDown();
+            log.debug("Exiting controller {} worker {}..", this.name, workerIndex);
+          },
+          0,
+          1,
+          TimeUnit.SECONDS);
+    }
+    try {
+      log.debug("Controller {} bootstrapping..", this.name);
+      latch.await();
+    } catch (InterruptedException e) {
+      log.error("Aborting controller.", e);
+    } finally {
+      log.info("Controller {} exited", this.name);
+    }
+  }
+
+  @Override
+  public void shutdown() {
+    // shutdown work-queue before the thread-pool
+    workQueue.shutDown();
+    workerThreadPool.shutdown();
+  }
+
+  private void worker() {
+    // taking tasks from work-queue in a loop
+    while (!workQueue.isShuttingDown()) {
+      Request request = null;
+      try {
+        request = workQueue.get();
+      } catch (InterruptedException e) {
+        // we're reaching here mostly because of forcibly shutting down the controller.
+        log.error("Controller worker interrupted.. keeps working until work-queue shutdown", e);
+        if (request != null) {
+          workQueue.addRateLimited(request);
+        }
+      }
+      // request is expected to be null, when the work-queue is shutting-down.
+      if (request == null) {
+        log.info("Controller {} worker exiting because work-queue has shutdown..");
+        return;
+      }
+      log.debug("Controller {} start reconciling {}..", request);
+
+      // do reconciliation, invoke user customized logic.
+      Result result = this.reconciler.reconcile(request);
+
+      try {
+        // checks whether do a re-queue (on failure)
+        if (result.isRequeue()) {
+          if (result.getRequeueAfter() == null) {
+            log.debug("Controller {} reconciling {} failed, requeuing {}..", request);
+            workQueue.addRateLimited(request);
+          } else {
+            log.debug(
+                "Controller {} reconciling {} failed, requeuing after {}..",
+                request,
+                result.getRequeueAfter());
+            workQueue.addAfter(request, result.getRequeueAfter());
+          }
+          return;
+        } else {
+          workQueue.forget(request);
+        }
+      } finally {
+        workQueue.done(request);
+        log.debug("Controller {} finished reconciling {}..", request);
+      }
+    }
+  }
+
+  public String getName() {
+    return name;
+  }
+
+  public void setName(String name) {
+    this.name = name;
+  }
+
+  public int getWorkerCount() {
+    return workerCount;
+  }
+
+  public void setWorkerCount(int workerCount) {
+    this.workerCount = workerCount;
+  }
+
+  public ExecutorService getWorkerThreadPool() {
+    return workerThreadPool;
+  }
+
+  public void setWorkerThreadPool(ScheduledExecutorService workerThreadPool) {
+    this.workerThreadPool = workerThreadPool;
+  }
+
+  public Reconciler getReconciler() {
+    return reconciler;
+  }
+
+  public void setReconciler(Reconciler reconciler) {
+    this.reconciler = reconciler;
+  }
+
+  public Duration getReadyTimeout() {
+    return readyTimeout;
+  }
+
+  public void setReadyTimeout(Duration readyTimeout) {
+    this.readyTimeout = readyTimeout;
+  }
+
+  public Duration getReadyCheckInternal() {
+    return readyCheckInternal;
+  }
+
+  public void setReadyCheckInternal(Duration readyCheckInternal) {
+    this.readyCheckInternal = readyCheckInternal;
+  }
+}
--- a/extended/src/main/java/io/kubernetes/client/extended/controller/WorkQueueResourceEventHandler.java
+++ b/extended/src/main/java/io/kubernetes/client/extended/controller/WorkQueueResourceEventHandler.java
@@ -1,0 +1,80 @@
+package io.kubernetes.client.extended.controller;
+
+import io.kubernetes.client.extended.controller.reconciler.Request;
+import io.kubernetes.client.extended.workqueue.WorkQueue;
+import io.kubernetes.client.informer.ResourceEventHandler;
+import java.util.function.BiPredicate;
+import java.util.function.Function;
+import java.util.function.Predicate;
+
+/**
+ * An event handler plumbs work-queue into a controller.
+ *
+ * @param <ApiType> the type parameter
+ */
+public class WorkQueueResourceEventHandler<ApiType> implements ResourceEventHandler<ApiType> {
+
+  private final WorkQueue<Request> workQueue;
+  private final Function<ApiType, Request> workKeyGenerator;
+  private Predicate<ApiType> onAddFilterPredicate;
+  private BiPredicate<ApiType, ApiType> onUpdateFilterPredicate;
+  private BiPredicate<ApiType, Boolean> onDeleteFilterPredicate;
+
+  /**
+   * Instantiates a new Work queue resource event handler.
+   *
+   * @param workQueue the work queue
+   * @param workKeyGenerator the work key generator
+   */
+  public WorkQueueResourceEventHandler(
+      WorkQueue<Request> workQueue, Function<ApiType, Request> workKeyGenerator) {
+    this.workQueue = workQueue;
+    this.workKeyGenerator = workKeyGenerator;
+  }
+
+  public Predicate<ApiType> getOnAddFilterPredicate() {
+    return onAddFilterPredicate;
+  }
+
+  public void setOnAddFilterPredicate(Predicate<ApiType> onAddFilterPredicate) {
+    this.onAddFilterPredicate = onAddFilterPredicate;
+  }
+
+  public BiPredicate<ApiType, ApiType> getOnUpdateFilterPredicate() {
+    return onUpdateFilterPredicate;
+  }
+
+  public void setOnUpdateFilterPredicate(BiPredicate<ApiType, ApiType> onUpdateFilterPredicate) {
+    this.onUpdateFilterPredicate = onUpdateFilterPredicate;
+  }
+
+  public BiPredicate<ApiType, Boolean> getOnDeleteFilterPredicate() {
+    return onDeleteFilterPredicate;
+  }
+
+  public void setOnDeleteFilterPredicate(BiPredicate<ApiType, Boolean> onDeleteFilterPredicate) {
+    this.onDeleteFilterPredicate = onDeleteFilterPredicate;
+  }
+
+  @Override
+  public void onAdd(ApiType obj) {
+    if (onAddFilterPredicate == null || onAddFilterPredicate.test(obj)) {
+      workQueue.add(workKeyGenerator.apply(obj));
+    }
+  }
+
+  @Override
+  public void onUpdate(ApiType oldObj, ApiType newObj) {
+    if (onUpdateFilterPredicate == null || onUpdateFilterPredicate.test(oldObj, newObj)) {
+      workQueue.add(workKeyGenerator.apply(newObj));
+    }
+  }
+
+  @Override
+  public void onDelete(ApiType obj, boolean deletedFinalStateUnknown) {
+    if (onDeleteFilterPredicate == null
+        || onDeleteFilterPredicate.test(obj, deletedFinalStateUnknown)) {
+      workQueue.add(workKeyGenerator.apply(obj));
+    }
+  }
+}
--- a/extended/src/main/java/io/kubernetes/client/extended/controller/reconciler/Reconciler.java
+++ b/extended/src/main/java/io/kubernetes/client/extended/controller/reconciler/Reconciler.java
@@ -1,0 +1,33 @@
+package io.kubernetes.client.extended.controller.reconciler;
+
+/**
+ * Reconciler defines interface for reconciling resource. Users are expected to implement this
+ * interface to develop their own controllers watching events on resources. A reconciler works by
+ * comparing the state specified in an object by a user against the actual cluster state, and then
+ * perform operations to make the actual cluster state reflect the state specified by the user.
+ * Typically, reconcile is triggered by a Controller in response to cluster Events (e.g. Creating,
+ * Updating, Deleting Kubernetes objects) or external Events (GitHub Webhooks, polling external
+ * sources, etc).
+ *
+ * <p>Example reconcile Logic:
+ *
+ * <p>1. Read an object and all the Pods it owns. 2. Observe that the object spec specifies 5
+ * replicas but actual cluster contains only 1 Pod replica. 3. Create 4 Pods and set their
+ * OwnerReferences to the object.
+ *
+ * <p>Reconciliation is level-based, meaning action isn't driven off changes in individual Events,
+ * but instead is driven by actual cluster state read from the apiserver or a local cache. For
+ * example if responding to a Pod Delete Event, the Request won't contain that a Pod was deleted,
+ * instead the reconcile function observes this when reading the cluster state and seeing the Pod as
+ * missing.
+ */
+public interface Reconciler {
+
+  /**
+   * Reconcile result.
+   *
+   * @param request the reconcile request, triggered by watch events
+   * @return the result
+   */
+  Result reconcile(Request request);
+}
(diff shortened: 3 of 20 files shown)


extended/src/main/java/io/kubernetes/client/extended/controller/ControllerBuilder.java
```diff
@@ -0,0 +1,163 @@
...
+import java.util.function.BiPredicate;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.function.Supplier;
+
+public class ControllerBuilder {
+
+  public static final int DEFAULT_WORKER_COUNT = 16;
+
+  private int workerCount = DEFAULT_WORKER_COUNT;
```
brendandburns: Do this assignment in the constructor.

brendandburns: and all below.

extended/src/main/java/io/kubernetes/client/extended/controller/ControllerBuilder.java
```diff
@@ -0,0 +1,163 @@
...
+    }
+    controller.setReconciler(reconciler);
+
+    if (leaderElector != null) {
+      return new LeaderElectingController(this.leaderElector, controller);
+    }
+    return controller;
+  }
+
+  public class ControllerWatchBuilder<ApiType, ApiListType> {
```
brendandburns: Let's extract this class out into it's own file.

extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
```diff
@@ -0,0 +1,192 @@
...
+import io.kubernetes.client.extended.workqueue.RateLimitingQueue;
+import java.time.Duration;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.function.Supplier;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DefaultController implements Controller {
+
```
brendandburns: delete whitespace.

extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
```diff
@@ -0,0 +1,192 @@
...
+  private static final Logger log = LoggerFactory.getLogger(DefaultController.class);
+
+  private Reconciler reconciler;
+  private String name;
+  private int workerCount;
+  private ExecutorService workerThreadPool;
+
+  private final RateLimitingQueue<Reconciler.Request> workQueue;
```
brendandburns: why are these final? I don't think that's needed.

extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
```diff
@@ -0,0 +1,192 @@
...
+      Supplier<Boolean>... readyFuncs) {
+    this.reconciler = reconciler;
+    this.workQueue = workQueue;
+    this.readyFuncs = readyFuncs;
+  }
+
+  @Override
+  public void run() {
+    if (workerCount <= 0) {
+      log.error("Fail to start controller {}: worker count must be positive.");
```
brendandburns: need a `return;`?

extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
```diff
@@ -0,0 +1,192 @@
...
+    }
+    if (workerThreadPool == null) {
```
brendandburns: here too?

extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
```diff
@@ -0,0 +1,192 @@
...
+      Reconciler reconciler,
+      RateLimitingQueue<Reconciler.Request> workQueue,
+      Supplier<Boolean>... readyFuncs) {
+    this.reconciler = reconciler;
+    this.workQueue = workQueue;
+    this.readyFuncs = readyFuncs;
+  }
+
+  @Override
+  public void run() {
```
brendandburns: This function is too long/complicated, I'd abstract out an `isSynced()` method or something like that.

extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
```diff
@@ -0,0 +1,192 @@
...
+      log.error("Timed out waiting for cache to be synced.");
+      return;
+    }
+
+    CountDownLatch latch = new CountDownLatch(workerCount);
+    for (int i = 0; i < this.workerCount; i++) {
+      final int workerIndex = i;
+      workerThreadPool.submit(
+          () -> {
+            if (log.isDebugEnabled()) {
```
brendandburns: Why do you have an if check here?

brendandburns: (and below)

extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
```diff
@@ -0,0 +1,192 @@
...
+              log.debug("Exiting controller {} worker {}..", this.name, workerIndex);
+            }
+          });
+    }
+    try {
+      if (log.isDebugEnabled()) {
+        log.debug("Controller {} bootstrapping..", this.name);
+      }
+      latch.await();
+    } catch (Throwable t) {
```
brendandburns: Don't blindly catch `Throwable()` catch the specific exception(s) you care about.

extended/src/main/java/io/kubernetes/client/extended/controller/ControllerBuilder.java
yue9944882: updated

extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
yue9944882: done

extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
yue9944882: removed final modifier

extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
yue9944882: nice catch!

extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
yue9944882: updated

extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
yue9944882: it's for improving performance, so that we won't reach the debugging branches on non-debugging deployment.. it's conventional but makes the code less readable on the flip side of the coin

extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
yue9944882: updated

extended/src/main/java/io/kubernetes/client/extended/controller/ControllerBuilder.java
```diff
@@ -0,0 +1,262 @@
...
+   * @param <ApiListType> the type parameter
+   */
+  public class ControllerWatchBuilder<ApiType, ApiListType> {
+
+    private ControllerWatchBuilder(Class<ApiType> apiTypeClass) {
+      this.apiTypeClass = apiTypeClass;
+      this.workKeyGenerator = Controllers.defaultReflectiveKeyFunc();
+    }
+
+    private Class<ApiType> apiTypeClass;
```
brendandburns: move instances up above constructor.

extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
brendandburns: No, but isn't there a 100% equivalent check in the `log.debug(...)`

Like I would guess that the first line of code in `log.debug(...)` is:

`if (!log.isDebugEnabled()) return;`

extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
brendandburns: Looks like you're still catching `Throwable` ?

extended/src/main/java/io/kubernetes/client/extended/controller/ControllerBuilder.java
yue9944882: the point of making `ControllerWatchBuilder` a non-static nested-class of `ControllerBuilder` is to force the usage so that we can only build a controller-watch when building a controller, i.e. enclosing the usage of watch-controller into controller-builder. additionally, the watch-builder is accessing a few fields from parent like informer-factory if we move it a separate class, then there will be bunch of copies when instantiating watch-builder from controller-builder. is that convincing?

extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
```diff
@@ -0,0 +1,295 @@
...
+            if (log.isDebugEnabled()) {
+              log.debug(
+                  "Controller {} reconciling {} failed, requeuing after {}..",
+                  request,
+                  result.getRequeueAfter());
+            }
+            workQueue.addAfter(request, result.getRequeueAfter());
+          }
+        }
+      } catch (InterruptedException e) {
```
brendandburns: What in this code throws `InterruptedException` can we wrap the potential throw more tightly?

extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
```diff
@@ -0,0 +1,295 @@
...
+          if (log.isDebugEnabled()) {
+            log.debug("Controller {} finished reconciling {}..", request);
+          }
+        }
+      }
+    }
+  }
+
+  /**
+   * Gets name.
```
brendandburns: This Javadoc seems super redundant, delete?

extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
```diff
@@ -0,0 +1,295 @@
...
+   *
+   * @return the name
+   */
+  public String getName() {
+    return name;
+  }
+
+  /**
+   * Sets name.
```
brendandburns: as above.

brendandburns: (and below too)

extended/src/main/java/io/kubernetes/client/extended/controller/WorkQueueResourceEventHandler.java
```diff
@@ -0,0 +1,116 @@
...
+   * @param workKeyGenerator the work key generator
+   */
+  public WorkQueueResourceEventHandler(
+      WorkQueue<Reconciler.Request> workQueue,
+      Function<ApiType, Reconciler.Request> workKeyGenerator) {
+    this.workQueue = workQueue;
+    this.workKeyGenerator = workKeyGenerator;
+  }
+
+  private final WorkQueue<Reconciler.Request> workQueue;
```
brendandburns: move instance variables above constructor.

extended/src/main/java/io/kubernetes/client/extended/controller/WorkQueueResourceEventHandler.java
```diff
@@ -0,0 +1,116 @@
...
+
+  private final Function<ApiType, Reconciler.Request> workKeyGenerator;
+
+  private Predicate<ApiType> onAddFilterPredicate;
+  private BiPredicate<ApiType, ApiType> onUpdateFilterPredicate;
+  private BiPredicate<ApiType, Boolean> onDeleteFilterPredicate;
+
+  /**
+   * Gets on add filter predicate.
```
brendandburns: delete redundant javadoc (here and everywhere)

extended/src/main/java/io/kubernetes/client/extended/controller/WorkQueueResourceEventHandler.java
```diff
@@ -0,0 +1,116 @@
...
+   *
+   * @param onDeleteFilterPredicate the on delete filter predicate
+   */
+  public void setOnDeleteFilterPredicate(BiPredicate<ApiType, Boolean> onDeleteFilterPredicate) {
+    this.onDeleteFilterPredicate = onDeleteFilterPredicate;
+  }
+
+  @Override
+  public void onAdd(ApiType obj) {
+    if (onAddFilterPredicate != null && onAddFilterPredicate.negate().test(obj)) {
```
brendandburns: Personally this feels cleaner as:

```java
if (onAddFilterPredicate == null || onAddFilterPredicate.test(obj)) {
   workQueue.add(...);
}
```

but it's your call.

extended/src/main/java/io/kubernetes/client/extended/controller/WorkQueueResourceEventHandler.java
```diff
@@ -0,0 +1,116 @@
...
+      return;
+    }
+    workQueue.add(workKeyGenerator.apply(obj));
+  }
+
+  @Override
+  public void onUpdate(ApiType oldObj, ApiType newObj) {
+    if (onUpdateFilterPredicate != null && onUpdateFilterPredicate.negate().test(oldObj, newObj)) {
```
brendandburns: as above.

extended/src/main/java/io/kubernetes/client/extended/controller/WorkQueueResourceEventHandler.java
```diff
@@ -0,0 +1,116 @@
...
+      return;
+    }
+    workQueue.add(workKeyGenerator.apply(newObj));
+  }
+
+  @Override
+  public void onDelete(ApiType obj, boolean deletedFinalStateUnknown) {
+    if (onDeleteFilterPredicate != null
```
brendandburns: here too.

extended/src/main/java/io/kubernetes/client/extended/controller/reconciler/Reconciler.java
```diff
@@ -0,0 +1,203 @@
...
+    public void setName(String name) {
+      this.name = name;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+      if (this == o) return true;
+      if (o == null || getClass() != o.getClass()) return false;
+      Request request = (Request) o;
+      return Objects.equal(name, request.name) && Objects.equal(namespace, request.namespace);
```
brendandburns: Do you need to use `Objects.equal(...)` here? can't you just do `name.equals(request.name)`?

extended/src/main/java/io/kubernetes/client/extended/controller/reconciler/Reconciler.java
```diff
@@ -0,0 +1,203 @@
...
+     *
+     * @param requeue the requeue
+     * @param requeueAfter the requeue after
+     */
+    public Result(boolean requeue, Duration requeueAfter) {
+      this.requeue = requeue;
+      this.requeueAfter = requeueAfter;
+    }
+
+    private boolean requeue;
```
brendandburns: instances before constructor.

brendandburns:
A bunch of comments, mostly minor.

I would delete all the redundant javadoc, I think it just clutters the code.

Thanks!

extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
yue9944882: > No, but isn't there a 100% equivalent check in the log.debug(...)

that depends on the underlying logging library. slf4j defines a unified interface for logger and other libraries like log4j actually works under it.

extended/src/main/java/io/kubernetes/client/extended/controller/ControllerBuilder.java
yue9944882: updated

extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
yue9944882: it `BlockingQueue#get` throwing `InterruptedException` and there we have a testing verifying the behavior when a potential throw happens from Reconciler, see `testControllerStopsWorkingWhenReconcilerAbortsWithRuntimeException `

extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
yue9944882: deleted all the trival java-doc on getter-setter

extended/src/main/java/io/kubernetes/client/extended/controller/WorkQueueResourceEventHandler.java
yue9944882: done

extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
yue9944882: deleted all the trival java-doc on getter-setter

extended/src/main/java/io/kubernetes/client/extended/controller/WorkQueueResourceEventHandler.java
yue9944882: yours looks clearer. updated.

extended/src/main/java/io/kubernetes/client/extended/controller/WorkQueueResourceEventHandler.java
yue9944882: updated

extended/src/main/java/io/kubernetes/client/extended/controller/reconciler/Reconciler.java
yue9944882: updated

extended/src/main/java/io/kubernetes/client/extended/controller/reconciler/Reconciler.java
yue9944882: it's generated by IDE tooling, the two actually works the same i believe

extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
brendandburns: No, but log.debug(...) shouldn't print if that log level isn't set, so regardless of the implementaton, it has to do the check, right?

extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
brendandburns: Can we just put the catch around that line of code?

extended/src/main/java/io/kubernetes/client/extended/controller/reconciler/Reconciler.java
brendandburns: I think there definitely the same, but `Objects.equal(...)` feels unnecessarily verbose/hard to read....

extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
yue9944882: after roughly googling this, i find it's actually debatable for quite a period of time. some folks insist branching by `log.isDebugEnabled()` is neglectable.. how do you think we consider solve it in a follow-up?

extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
yue9944882: done

extended/src/main/java/io/kubernetes/client/extended/controller/reconciler/Reconciler.java
yue9944882: updated. refactored to old-school `equals` and `hashCode`

extended/src/main/java/io/kubernetes/client/extended/controller/ControllerBuilder.java
```diff
@@ -0,0 +1,260 @@
...
+  private String controllerName;
+  private RateLimitingQueue<Reconciler.Request> workQueue;
+
+  private SharedInformerFactory informerFactory;
+  private LeaderElector leaderElector;
+  private List<Supplier<Boolean>> readyFuncs;
+  private Reconciler reconciler;
+
+  private ControllerBuilder() {
+    // Make constructor private explicitly
```
brendandburns: delete this comment.

extended/src/main/java/io/kubernetes/client/extended/controller/ControllerBuilder.java
```diff
@@ -0,0 +1,260 @@
...
+  public static ControllerBuilder withInformerFactory(SharedInformerFactory informerFactory) {
+    ControllerBuilder builder = new ControllerBuilder();
+    builder.informerFactory = informerFactory;
+    return builder;
+  }
+
+  /**
+   * Starts building watches over resource.
+   *
+   * @param <ApiType> the type parameter
```
brendandburns: 'the type parameter for the singular response'

extended/src/main/java/io/kubernetes/client/extended/controller/ControllerBuilder.java
```diff
@@ -0,0 +1,260 @@
...
+   * @param <ApiListType> the type parameter
```
brendandburns: 'The type for the List of resources'

extended/src/main/java/io/kubernetes/client/extended/controller/ControllerBuilder.java
```diff
@@ -0,0 +1,260 @@
...
+        new DefaultController(
+            this.reconciler, this.workQueue, this.readyFuncs.stream().toArray(Supplier[]::new));
+
+    controller.setName(this.controllerName);
+    controller.setWorkerCount(this.workerCount);
+    controller.setWorkerThreadPool(
+        Executors.newScheduledThreadPool(
+            this.workerCount, Controllers.namedControllerThreadFactory(this.controllerName)));
+
+    if (reconciler == null) {
```
brendandburns: this.reconciler (also move to the beginning of the function)

extended/src/main/java/io/kubernetes/client/extended/controller/ControllerBuilder.java
```diff
@@ -0,0 +1,260 @@
...
+      throw new IllegalStateException("Missing reconciler when building controller.");
+    }
+    controller.setReconciler(reconciler);
+
+    if (leaderElector != null) {
```
brendandburns: this.leaderElector (consistency with the call below)

extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
brendandburns: I'd really love to remove the code, I think it makes it much harder to read with minimal benefit in performance.

extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
brendandburns: As above, don't just catch `Throwable`

extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
```diff
@@ -0,0 +1,235 @@
...
+  }
+
+  private void worker() {
+    // taking tasks from work-queue in a loop
+    while (!workQueue.isShuttingDown()) {
+      Reconciler.Request request = null;
+      try {
+        request = workQueue.get();
+      } catch (InterruptedException e) {
+        // we're reaching here mostly because of forcibly shutting down the controller.
```
brendandburns: I think alignment is messed up?

extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
```diff
@@ -0,0 +1,235 @@
...
+            log.debug(
+                "Controller {} reconciling {} failed, requeuing after {}..",
+                request,
+                result.getRequeueAfter());
+          }
+          workQueue.addAfter(request, result.getRequeueAfter());
+        }
+      }
+
+      if (request != null) {
```
brendandburns: How could request be null here? There seems to be lots of usage of request above that would fail if request is null...

brendandburns: There appears to be a null check on line 150....

extended/src/main/java/io/kubernetes/client/extended/controller/reconciler/Reconciler.java
```diff
@@ -0,0 +1,214 @@
...
+   * @return the result
+   */
+  Result reconcile(Request request);
+
+  /**
+   * The type Request contains the information necessary to reconcile a resource object. This
+   * includes the information to uniquely identify the object - its Name and Namespace. Note that it
+   * does NOT contain information about any specific Event or the object contents itself.
+   */
+  class Request {
```
brendandburns: Break these classes out?

brendandburns:
Even more comments (sorry!)

extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
yue9944882: updated, ptal

extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
yue9944882: removed all the debug condition check

extended/src/main/java/io/kubernetes/client/extended/controller/reconciler/Reconciler.java
yue9944882: done

extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
yue9944882: uh indeed.. the null check is redandunt here

extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
yue9944882: no? i think? the formatter requests it..

extended/src/main/java/io/kubernetes/client/extended/controller/ControllerBuilder.java
yue9944882: done

extended/src/main/java/io/kubernetes/client/extended/controller/ControllerBuilder.java
yue9944882: done

extended/src/main/java/io/kubernetes/client/extended/controller/ControllerBuilder.java
yue9944882: updated

extended/src/main/java/io/kubernetes/client/extended/controller/ControllerBuilder.java
yue9944882: updated

extended/src/main/java/io/kubernetes/client/extended/controller/ControllerBuilder.java
yue9944882: deleted

yue9944882:
@brendandburns i refactored the builder package a bit for a clearer structure, please take another look, thanks

PR #639, opened 2019-07-28
add example mock "kubectl get configmap xx -o json"

ZP-AlwaysWin:
add example mock "kubectl get configmap xx -o json"


examples/src/main/java/io/kubernetes/client/examples/KubectlJsonExample.java
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * A simple example of how to use the Java API
+ *
+ * <p>Easiest way to run this: mvn exec:java
+ * -Dexec.mainClass="io.kubernetes.client.examples.CopyExample"
+ *
+ * <p>From inside $REPO_DIR/examples
+ *
+ * <p>mock "kubectl get configamp configmap-name -o json"
```
yue9944882: what's the point of adding this example? i don't think this example worth a new file..

ZP-AlwaysWin: hi @yue9944882
I use java-client to work for pass platform, I need to get xxx.json to change, mock kubectl get resourcename -o json。but Under the existing examples，I don't konw how to do it. Stand in the user's Perspective ，I add example。

It's up to you to decide.

Best Wishes

brendandburns:
This is  probably a good thing to add to the ParseExample.java instead of a brand new example.

Thanks!

ZP-AlwaysWin:
> This is probably a good thing to add to the ParseExample.java instead of a brand new example.
> 
> Thanks!

hi @brendandburns

I already add the example to the ParseExample.java, I am using the java-client to work for paas , If I find any minor problems or usability problems, I will contribute.

Best Wishes

examples/src/main/java/io/kubernetes/client/examples/ParseExample.java
```diff
@@ -61,4 +66,31 @@ public static void main(String[] args) throws IOException, ApiException, ClassNo
...
+
+    Response response = call.execute();
+
+    if (!response.isSuccessful()) {
+      return;
+    }
+
+    String body = response.body().string();
+
+    JsonObject returnData = new JsonParser().parse(body).getAsJsonObject();
```
yue9944882: will this work with the custom-type adapter? in the existing example, it's parsing by using the nested gson instance where we register the adapters?

```java
    Object obj2 =
        client	        client
            .getJSON()	            .getJSON()
```

brendandburns: Hrm, I think that Gson adapters were injected via annotation so that they would work everywhere, e.g. https://github.com/kubernetes-client/java/blob/master/kubernetes/src/main/java/io/kubernetes/client/custom/Quantity.java#L14

yue9944882: https://github.com/kubernetes-client/java/blob/607a34b2dad0af5ea87ec6e37cf7db4536f545af/kubernetes/src/main/java/io/kubernetes/client/JSON.java#L46-L54

but here we're registering the adapters explicitly

brendandburns: Ah, that's true for those built-ins we can't annotate...

brendandburns:
Closing this since it is old. Please feel free to re-open if there are updates.

PR #649, opened 2019-08-02
Refactor: Add constructor for customizing index funcs

yue9944882:
for now, we are only allowed to use `<namespace>` index funcs, this pull enables users to add customized index func. e.g. (as the test shown), for indexing pods, not only we can index by pod's namespace but also pod's nodeName from the spec:

1. Moves static helpers from `Cache` to `Caches`. note that we're deprecating the former and left a `TODO` noting it's gonna be removed after 7.0.0 release
2. Add more constructors to informer so that we can inject customized cache.

### DIFF
--- a/util/src/main/java/io/kubernetes/client/informer/cache/Cache.java
+++ b/util/src/main/java/io/kubernetes/client/informer/cache/Cache.java
@@ -1,13 +1,14 @@
 package io.kubernetes.client.informer.cache;
 
-import com.google.common.base.Strings;
-import io.kubernetes.client.informer.exception.BadObjectException;
-import io.kubernetes.client.models.V1ObjectMeta;
-import io.kubernetes.client.util.Reflect;
-import io.kubernetes.client.util.exception.ObjectMetaReflectException;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.locks.ReentrantLock;
-import java.util.function.*;
+import java.util.function.Function;
 
 /**
  * Cache is a java port of k/client-go's ThreadSafeStore. It basically saves and indexes all the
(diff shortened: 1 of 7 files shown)


util/src/test/java/io/kubernetes/client/informer/cache/CacheTest.java
```diff
@@ -4,9 +4,8 @@
 
 import io.kubernetes.client.models.V1ObjectMeta;
 import io.kubernetes.client.models.V1Pod;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.List;
+import io.kubernetes.client.models.V1PodSpec;
+import java.util.*;
```
brendandburns: Don't use wildcard import.

brendandburns:
one minor comment.

util/src/main/java/io/kubernetes/client/informer/cache/Cache.java
```diff
@@ -1,10 +1,5 @@
 package io.kubernetes.client.informer.cache;
 
-import com.google.common.base.Strings;
-import io.kubernetes.client.informer.exception.BadObjectException;
-import io.kubernetes.client.models.V1ObjectMeta;
-import io.kubernetes.client.util.Reflect;
-import io.kubernetes.client.util.exception.ObjectMetaReflectException;
 import java.util.*;
```
brendandburns: Remove wildcard import...

util/src/main/java/io/kubernetes/client/informer/cache/Cache.java
```diff
@@ -1,10 +1,5 @@
...
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.*;
```
brendandburns: Here too.

util/src/test/java/io/kubernetes/client/informer/cache/CachesTest.java
```diff
@@ -0,0 +1,28 @@
+package io.kubernetes.client.informer.cache;
+
+import static org.junit.Assert.*;
```
brendandburns: Remote wildcard import

util/src/test/java/io/kubernetes/client/informer/cache/CacheTest.java
yue9944882: removed

util/src/main/java/io/kubernetes/client/informer/cache/Cache.java
yue9944882: removed

util/src/test/java/io/kubernetes/client/informer/cache/CachesTest.java
yue9944882: done

PR #662, opened 2019-08-21
manually set OKHttpClient read timeout zero ensure client doesnot tim…

adohe:
…eout

### DIFF
--- a/util/src/main/java/io/kubernetes/client/informer/SharedInformerFactory.java
+++ b/util/src/main/java/io/kubernetes/client/informer/SharedInformerFactory.java
@@ -14,6 +14,7 @@ import java.util.Map;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
 
 /** SharedInformerFactory class constructs and caches informers for api types. */
 public class SharedInformerFactory {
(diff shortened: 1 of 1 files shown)


yue9944882:
default read-timeout is 10s, so the watch connection is breaking every 10s which leads to bursty memory usage in the controller.

/lgtm
/approve

util/src/main/java/io/kubernetes/client/informer/SharedInformerFactory.java
```diff
@@ -24,9 +25,16 @@
 
   private ExecutorService informerExecutor;
 
+  private ApiClient apiClient;
+
   /** Constructor w/ default thread pool. */
   public SharedInformerFactory() {
-    this(Executors.newCachedThreadPool());
+    this(null, Executors.newCachedThreadPool());
```
yue9944882: this(Configuration.getDefaultApiClient(), Executors.newCachedThreadPool())

util/src/main/java/io/kubernetes/client/informer/SharedInformerFactory.java
```diff
@@ -35,6 +43,17 @@ public SharedInformerFactory() {
...
+  }
+
+  /**
+   * Constructor w/ api client and thread pool specified.
+   *
+   * @param client specific api client
+   * @param threadPool specified thread pool
+   */
+  public SharedInformerFactory(ApiClient client, ExecutorService threadPool) {
+    apiClient = client;
```
yue9944882: >     // set read timeout zero to ensure client doesn't time out
    apiClient.getHttpClient().setReadTimeout(0, TimeUnit.MILLISECONDS);

we don't have to call this everytime creating a lister-watcher, let's move this to the constructor

PR #665, opened 2019-08-21
Fix: should not return from DefaultController.worker() method when re…

cizezsy:
…queue is true and add lacked arguments to format string.

I think we shouldn't return from worker() method when requeue is true: [line 167](https://github.com/kubernetes-client/java/pull/665/files#diff-efac65a7ef4bf5e848b418d22794ee5fL167), and add some lacked arguments to logs :)


yue9944882:
ci not happy, otherwise lgtm

cizezsy:
Fix it :P

yue9944882:
thanks!

/lgtm
/approve

PR #666, opened 2019-08-22
Fixes unexpectedly listener quitting leads to OOM

yue9944882:
/cc @brendandburns 

bugfix: listener thread can quit unexpectedly leaves queue growing to OOM..

### DIFF
--- a/util/src/main/java/io/kubernetes/client/informer/cache/ProcessorListener.java
+++ b/util/src/main/java/io/kubernetes/client/informer/cache/ProcessorListener.java
@@ -90,7 +90,9 @@ public class ProcessorListener<ApiType> implements Runnable {
     if (obj == null) {
       return;
     }
-    this.queue.add(obj);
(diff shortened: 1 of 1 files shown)


util/src/main/java/io/kubernetes/client/informer/cache/ProcessorListener.java
```diff
@@ -64,6 +64,9 @@ public void run() {
       } catch (InterruptedException e) {
         log.error("processor interrupted: {}", e.getMessage());
         return;
+      } catch (Throwable t) {
```
brendandburns: Hrm, I don't love this here, I assume that this is because we're running a custom user handler and it could throw anything. Perhaps we can just wrap the handler call in the catch of Throwable?

yue9944882: exactly. i wrap the throwable-catching block only for handler invocation

PR #669, opened 2019-08-26
Fix: remove the capacity limit of ProcessorListener

cizezsy:


util/src/main/java/io/kubernetes/client/informer/cache/ProcessorListener.java
```diff
@@ -90,9 +87,7 @@ public void add(Notification<ApiType> obj) {
     if (obj == null) {
       return;
     }
-    if (!this.queue.offer(obj)) {
-      log.warn("notification queue full!");
-    }
+    this.queue.add(obj);
```
yue9944882: minimun change. it doesnt really make a difference tho.

util/src/main/java/io/kubernetes/client/informer/cache/ProcessorListener.java
```diff
@@ -33,7 +30,7 @@ public ProcessorListener(ResourceEventHandler<ApiType> handler, long resyncPerio
     this.resyncPeriod = resyncPeriod;
     this.handler = handler;
 
-    this.queue = new ArrayBlockingQueue<>(DEFAULT_QUEUE_CAPACITY);
+    this.queue = new LinkedBlockingQueue<>();
```
yue9944882: this is interesting.. i roughly searched and understood the difference between between `ArrayBlockingQueue` and `LinkedBlockingQueue`. and the latter indeed wins in the unbounded case.. is there a third option for blocking queue?

util/src/main/java/io/kubernetes/client/informer/cache/ProcessorListener.java
cizezsy: Yeah, but the code is less...the log is redundant as the capacity is unlimited.

util/src/main/java/io/kubernetes/client/informer/cache/ProcessorListener.java
cizezsy: I have searched implementations of BlockingQueue in our current libraries and the LinkedBlockingQueue is most suitable for our case. Do you have any suggestion for it?

PR #677, opened 2019-08-28
676 - fix for EC cert loading exception

darinhoward:
Possible fix for https://github.com/kubernetes-client/java/issues/676


darinhoward:
i made a second adjustment as the original change fixed the cert load - but no the private key.

PR #708, opened 2019-09-19
Support "newDeadline" for leader election

### LINKED ISSUE #706: "renewDeadline" support for leader election
https://github.com/kubernetes-client/java/blob/5ae91646b465774c194a1b9299274005c185d5d6/extended/src/main/java/io/kubernetes/client/extended/leaderelection/LeaderElectionConfig.java#L11

The "renewDeadline" field in `LeaderElectionConfig` seems to be useless currently, is there a plan to support renew deadline duration in the future?
Probably related to this https://github.com/kubernetes-client/java/pull/538#discussion_r274320762

wu8685:
Fix #706

### DIFF
--- a/extended/src/main/java/io/kubernetes/client/extended/leaderelection/LeaderElector.java
+++ b/extended/src/main/java/io/kubernetes/client/extended/leaderelection/LeaderElector.java
@@ -15,6 +15,41 @@ public class LeaderElector {
 
   private static final double JITTER_FACTOR = 1.2;
 
+  public static LeaderElector newLeaderElector(LeaderElectionConfig config) {
+    if (config == null) {
+      throw new RuntimeException("Config must be provided.");
+    }
+    if (config.getLock() == null) {
+      throw new RuntimeException("Lock must be provided.");
+    }
+    if (config.getLeaseDuration() == null) {
+      throw new RuntimeException("LeaseDuration must be provided.");
+    }
+    if (config.getRetryPeriod() == null) {
+      throw new RuntimeException("RetryPeriod must be provided.");
+    }
+    if (config.getRenewDeadline() == null) {
+      throw new RuntimeException("RenewDeadline must be provided.");
+    }
+    if (config.getLeaseDuration().compareTo(config.getRenewDeadline()) <= 0) {
+      throw new RuntimeException("LeaseDuration must be greater than renewDeadline.");
+    }
+    if (config.getRenewDeadline().compareTo(config.getRetryPeriod()) <= 0) {
+      throw new RuntimeException("RenewDeadline must be greater than retryPeriod.");
+    }
+    if (config.getLeaseDuration().isZero() || config.getLeaseDuration().isNegative()) {
+      throw new RuntimeException("LeaseDuration must be greater than zero.");
+    }
+    if (config.getRenewDeadline().isZero() || config.getRenewDeadline().isNegative()) {
+      throw new RuntimeException("RenewDeadline must be greater than zero.");
+    }
+    if (config.getRetryPeriod().isZero() || config.getRetryPeriod().isNegative()) {
+      throw new RuntimeException("RetryPeriod must be greater than zero.");
+    }
+
+    return new LeaderElector(config);
+  }
+
   private final LeaderElectionConfig config;
 
   // internal bookkeeping
@@ -25,7 +60,7 @@ public class LeaderElector {
   private ExecutorService leaseWorkers = Executors.newSingleThreadExecutor();
   private ExecutorService hookWorkers = Executors.newSingleThreadExecutor();
 
-  public LeaderElector(LeaderElectionConfig config) {
+  protected LeaderElector(LeaderElectionConfig config) {
     this.config = config;
   }
 
@@ -97,6 +132,7 @@ public class LeaderElector {
     Duration retryPeriod = config.getRetryPeriod();
     long retryPeriodMillis = retryPeriod.toMillis();
     AtomicBoolean renewed = new AtomicBoolean(true);
+    long renewDeadlineMillis = config.getRenewDeadline().toMillis();
 
     ScheduledFuture scheduledFuture =
         scheduledWorkers.scheduleAtFixedRate(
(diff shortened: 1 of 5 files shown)


extended/src/main/java/io/kubernetes/client/extended/leaderelection/LeaderElector.java
```diff
@@ -15,6 +15,41 @@
 
   private static final double JITTER_FACTOR = 1.2;
 
+  public static LeaderElector newLeaderElector(LeaderElectionConfig config) {
+    if (config == null) {
+      throw new RuntimeException("Config must be provided.");
+    }
+    if (config.getLock() == null) {
+      throw new RuntimeException("Lock must be provided.");
```
yue9944882: lang.IllegalArgumentException?

extended/src/main/java/io/kubernetes/client/extended/leaderelection/LeaderElector.java
```diff
@@ -15,6 +15,41 @@
 
   private static final double JITTER_FACTOR = 1.2;
 
+  public static LeaderElector newLeaderElector(LeaderElectionConfig config) {
```
yue9944882: any reason move this out of the constructor?

extended/src/main/java/io/kubernetes/client/extended/leaderelection/LeaderElector.java
```diff
@@ -25,7 +60,7 @@
   private ExecutorService leaseWorkers = Executors.newSingleThreadExecutor();
   private ExecutorService hookWorkers = Executors.newSingleThreadExecutor();
 
-  public LeaderElector(LeaderElectionConfig config) {
+  protected LeaderElector(LeaderElectionConfig config) {
```
yue9944882: we cant tighten that for backward-compatibility.

extended/src/main/java/io/kubernetes/client/extended/leaderelection/LeaderElector.java
```diff
@@ -97,14 +132,15 @@ private void renewLoop() {
...
+    long renewDeadlineMillis = config.getRenewDeadline().toMillis();
 
     ScheduledFuture scheduledFuture =
         scheduledWorkers.scheduleAtFixedRate(
             () -> {
               Future<Boolean> future = leaseWorkers.submit(this::tryAcquireOrRenew);
               try {
                 //  set renew success
-                renewed.set(future.get(retryPeriodMillis, TimeUnit.MILLISECONDS));
+                renewed.set(future.get(renewDeadlineMillis, TimeUnit.MILLISECONDS));
```
moeyui1: I don't think this is what `renewDeadline` means. We should retry `tryAcquireOrRenew` even it return false within `renewDeadline` Duration. Just as what `client-go` does 
```go
wait.Until(func() {
		timeoutCtx, timeoutCancel := context.WithTimeout(ctx, le.config.RenewDeadline)
		defer timeoutCancel()
		err := wait.PollImmediateUntil(le.config.RetryPeriod, func() (bool, error) {
			done := make(chan bool, 1)
			go func() {
				defer close(done)
				done <- le.tryAcquireOrRenew()
			}()

			select {
			case <-timeoutCtx.Done():
				return false, fmt.Errorf("failed to tryAcquireOrRenew %s", timeoutCtx.Err())
			case result := <-done:
				return result, nil
			}
		}, timeoutCtx.Done())

		le.maybeReportTransition()
		desc := le.config.Lock.Describe()
		if err == nil {
			klog.V(5).Infof("successfully renewed lease %v", desc)
			return
		}
		le.config.Lock.RecordEvent("stopped leading")
		le.metrics.leaderOff(le.config.Name)
		klog.Infof("failed to renew lease %v: %v", desc, err)
		cancel()
	}, le.config.RetryPeriod, ctx.Done())
```

extended/src/main/java/io/kubernetes/client/extended/leaderelection/LeaderElector.java
```diff
@@ -15,6 +15,41 @@
 
   private static final double JITTER_FACTOR = 1.2;
 
+  public static LeaderElector newLeaderElector(LeaderElectionConfig config) {
+    if (config == null) {
+      throw new RuntimeException("Config must be provided.");
```
brendandburns: IlleagalArgumentException

extended/src/main/java/io/kubernetes/client/extended/leaderelection/LeaderElector.java
brendandburns: definitely this and everywhere. Please don't use `RuntimeException` anywhere it's not 100% required.

extended/src/main/java/io/kubernetes/client/extended/leaderelection/LeaderElector.java
```diff
@@ -15,6 +15,41 @@
...
+      throw new RuntimeException("RenewDeadline must be greater than retryPeriod.");
+    }
+    if (config.getLeaseDuration().isZero() || config.getLeaseDuration().isNegative()) {
+      throw new RuntimeException("LeaseDuration must be greater than zero.");
+    }
+    if (config.getRenewDeadline().isZero() || config.getRenewDeadline().isNegative()) {
+      throw new RuntimeException("RenewDeadline must be greater than zero.");
+    }
+    if (config.getRetryPeriod().isZero() || config.getRetryPeriod().isNegative()) {
+      throw new RuntimeException("RetryPeriod must be greater than zero.");
```
brendandburns: Rather than throwing a bunch of exceptions linearly, it's much better to accumulate all of the errors in a single list and then throw one exception. Something like:

```java
List<String> errors = new ArrayList();
if (config.getRetryPeriod().isZero()) {
   errors.add("retry can't be zero");
}
.... // other if's here.

if (errors.length() > 0) {
   throw new IllegalArgumentException(errors.join(","));
}
```

That way the user has a much better experience rather than fixing an error, re-running, getting a new error, etc etc.

extended/src/main/java/io/kubernetes/client/extended/leaderelection/LeaderElector.java
brendandburns: +1

brendandburns:
@moeyui1  Is this PR still needed?

moeyui1:
> @moeyui1 Is this PR still needed?

No, it could be closed.

PR #718, opened 2019-09-27
Implement copyFileToPod

elanv:
There are utilities copying files from a pod, but it is missing transferring files to a pod. Added a method transferring a file to a pod.

Issue #353

### DIFF
--- a/util/src/main/java/io/kubernetes/client/Copy.java
+++ b/util/src/main/java/io/kubernetes/client/Copy.java
@@ -16,15 +16,20 @@ import com.google.common.io.ByteStreams;
 import io.kubernetes.client.models.V1Pod;
 import java.io.BufferedInputStream;
 import java.io.File;
+import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.nio.file.Path;
 import org.apache.commons.codec.binary.Base64InputStream;
+import org.apache.commons.codec.binary.Base64OutputStream;
 import org.apache.commons.compress.archivers.ArchiveEntry;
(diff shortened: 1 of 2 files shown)


elanv:
I signed it

elanv:
I signed it

elanv:
I added a unit test and the code changed a bit.

util/src/main/java/io/kubernetes/client/Copy.java
```diff
@@ -161,4 +166,34 @@ public static void copyFileFromPod(String namespace, String pod, String srcPath,
...
+
+  public void copyFileToPod(
+      String namespace, String pod, String container, File srcFile, String destPath)
+      throws ApiException, IOException {
+    // Run decoding and extracting processes
+    final Process proc =
+        this.exec(
+            namespace,
+            pod,
+            new String[] {"sh", "-c", "base64 -d | tar -xf - -C " + destPath},
```
alexandrupetre: How will this work against a windows pod?

elanv: Sorry for the late reply. Currently this function has such a limitation. But basically the same is true for the "kubectl cp" issue, and it doesn't seem to be an easy one.

https://github.com/kubernetes/kubernetes/issues/58512

juyiwang7: My file is 120Mb I tried this code,and it returned at 17Mb. How can I resolve this problem?

auxten: > My file is 120Mb I tried this code,and it returned at 17Mb. How can I resolve this problem?

There may be a hole in your 120Mb file, you can test the file with `du -h your_file` to see the size without a hole.

PR #723, opened 2019-10-04
Set config file when using home directory

### LINKED ISSUE #668: config file is not correctly set when using home dir
Hello,

I think I may have found a minor issue here:
https://github.com/kubernetes-client/java/blob/release-5.0/util/src/main/java/io/kubernetes/client/util/ClientBuilder.java#L100

Shouldn't it be: `kc.setFile(config);` rather than `kc.setFile(kubeConfig);`?

Cheers,
Amo

hpandeycodeit:
- Sets the correct config file when using Home Directory
- Added a test case as well

Fix #668


brendandburns:
/lgtm
/approve

Thanks!

PR #724, opened 2019-10-09
EntrySet usage instread of get(prefix)

denis256:
Small performance improvement to use entrySet values instead of getting keys from the set


brendandburns:
/lgtm
/approve

Thanks for the PR!

PR #727, opened 2019-10-11
Magic exception message update + FileInputStream closing in SSLUtils 

denis256:
Hello,
updated ProtoClient to encode in hex not equal magic value, direct concatenation will lead to values like "I@4fee666" which aren't very helpful.

Noticed that in SSLUtils.loadDefaultStoreFile file input stream is not closed after loading of the file to keystore - added java 8 file closing.

### DIFF
--- a/util/src/main/java/io/kubernetes/client/util/SSLUtils.java
+++ b/util/src/main/java/io/kubernetes/client/util/SSLUtils.java
@@ -316,7 +316,9 @@ public class SSLUtils {
   private static boolean loadDefaultStoreFile(KeyStore keyStore, File fileToLoad, char[] passphrase)
       throws CertificateException, NoSuchAlgorithmException, IOException {
     if (fileToLoad.exists() && fileToLoad.isFile() && fileToLoad.length() > 0) {
-      keyStore.load(new FileInputStream(fileToLoad), passphrase);
+      try (FileInputStream inputStream = new FileInputStream(fileToLoad)) {
+        keyStore.load(inputStream, passphrase);
+      }
       return true;
     }
     return false;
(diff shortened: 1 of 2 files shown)


yue9944882:
/lgtm

@brendandburns

util/src/main/java/io/kubernetes/client/util/SSLUtils.java
```diff
@@ -316,7 +317,13 @@ private static void loadDefaultKeyStoreFile(KeyStore keyStore, char[] keyStorePa
   private static boolean loadDefaultStoreFile(KeyStore keyStore, File fileToLoad, char[] passphrase)
       throws CertificateException, NoSuchAlgorithmException, IOException {
     if (fileToLoad.exists() && fileToLoad.isFile() && fileToLoad.length() > 0) {
-      keyStore.load(new FileInputStream(fileToLoad), passphrase);
+      FileInputStream inputStream = null;
```
brendandburns: Maybe use the Java8 syntax:

```java
try (FileInputStream inputStream = new File InputStream(fileToLoad)) {
    ....
}
```

brendandburns: oh, I guess you're not just calling `close()` nevermind.

brendandburns:
This looks fine to me. Can you squash the commits into a single commit? Then we can merge.

Thanks

util/src/main/java/io/kubernetes/client/util/SSLUtils.java
denis256: Make sense to use Java8 closing.

brendandburns:
/lgtm
/approve

Thanks for the PR!

PR #731, opened 2019-10-15
Bugfix(informer): DeltaFIFO Replace method to prevent SharedIndexInformers from missing notifications

yue9944882:
picking the informer fix from https://github.com/kubernetes/kubernetes/pull/83911


yue9944882:
cc @adohe

PR #735, opened 2019-10-17
Fixed argument list of listPodForAllNamespaces

### LINKED ISSUE #734: listPodForAllNamespaces() argument list differs in KubeConfigFileClientExample.java
```
V1PodList list = api.listPodForAllNamespaces(null, null, null, null, null, null, null, null);
```
it takes 8 nulls in original file, but actually takes 9 nulls.

seongwon-kang:
Fixes #734


brendandburns:
This doesn't work with the current client. I think it changed between `5.0.0` and `6.0.1`

We should perhaps introduce versioned directories for our example code, since it's hard to keep it working for older versions of the library.

Closing this PR since it can't pass CI.

PR #738, opened 2019-10-17
Generic kubernetes api client

yue9944882:
@brendandburns 

NOTE: i will sanitize the pull l8r (the license, formatting, etc..).. 

it's neat and simple. w/ the help of this generic client, what we only need is the generated model. i just drafted initial version of code and test, it already covers all the api paths on the kubernetes apiserver.

the general idea is to make use of CustomObjectApi as best as we can, this will help us access custom resources in kubernetes in an easier way. appreciated if you feel like shed some light on it

### DIFF
--- a/extended/src/main/java/io/kubernetes/client/extended/generic/GenericKubernetesApi.java
+++ b/extended/src/main/java/io/kubernetes/client/extended/generic/GenericKubernetesApi.java
@@ -1,0 +1,686 @@
+package io.kubernetes.client.extended.generic;
+
+import com.google.common.annotations.Beta;
+import com.google.common.base.Strings;
+import com.google.gson.Gson;
+import com.google.gson.JsonElement;
+import com.google.gson.reflect.TypeToken;
+import io.kubernetes.client.custom.V1Patch;
+import io.kubernetes.client.extended.generic.options.CreateOptions;
+import io.kubernetes.client.extended.generic.options.DeleteOptions;
+import io.kubernetes.client.extended.generic.options.GetOptions;
+import io.kubernetes.client.extended.generic.options.ListOptions;
+import io.kubernetes.client.extended.generic.options.PatchOptions;
+import io.kubernetes.client.extended.generic.options.UpdateOptions;
+import io.kubernetes.client.openapi.ApiClient;
+import io.kubernetes.client.openapi.ApiException;
+import io.kubernetes.client.openapi.Configuration;
+import io.kubernetes.client.openapi.apis.CustomObjectsApi;
+import io.kubernetes.client.openapi.models.V1ObjectMeta;
+import io.kubernetes.client.openapi.models.V1Status;
+import io.kubernetes.client.util.PatchUtils;
+import io.kubernetes.client.util.Reflect;
+import io.kubernetes.client.util.Watch;
+import io.kubernetes.client.util.Watchable;
+import io.kubernetes.client.util.exception.ObjectMetaReflectException;
+import java.net.SocketTimeoutException;
+import okhttp3.Call;
+
+/**
+ * The Generic kubernetes api provides a unified client interface for not only the non-core-group
+ * built-in resources from kuberntes but also the custom-resources models meet the following
+ * requirements:
+ *
+ * <p>1. there's a `V1ObjectMeta` field in the model along with its getter/setter. 2. there's a
+ * `V1ListMeta` field in the list model along with its getter/setter. - supports Gson
+ * serialization/deserialization. 3. the generic kubernetes api covers all the basic operations over
+ * the custom resources including {get, list, watch, create, update, patch, delete}.
+ *
+ * @param <ApiType> the api type parameter
+ * @param <ApiListType> the api list type parameter
+ */
+@Beta
+public class GenericKubernetesApi<ApiType, ApiListType> {
+
+  // TODO(yue9944882): supports status operations..
+  // TODO(yue9944882): supports generic sub-resoruce operations..
+  // TODO(yue9944882): supports delete-collections..
+
+  private Class<ApiType> apiTypeClass;
+  private Class<ApiListType> apiListTypeClass;
+  private String apiGroup;
+  private String apiVersion;
+  private String resourcePlural;
+  private CustomObjectsApi customObjectsApi;
+
+  /**
+   * Instantiates a new Generic kubernetes api.
+   *
+   * @param apiTypeClass the api type class
+   * @param apiListTypeClass the api list type class
+   * @param apiGroup the api group
+   * @param apiVersion the api version
+   * @param resourcePlural the resource plural
+   */
+  public GenericKubernetesApi(
+      Class<ApiType> apiTypeClass,
+      Class<ApiListType> apiListTypeClass,
+      String apiGroup,
+      String apiVersion,
+      String resourcePlural) {
+    this(
+        apiTypeClass,
+        apiListTypeClass,
+        apiGroup,
+        apiVersion,
+        resourcePlural,
+        new CustomObjectsApi(Configuration.getDefaultApiClient()));
+  }
+
+  /**
+   * Instantiates a new Generic kubernetes api.
+   *
+   * @param apiTypeClass the api type class, e.g. V1Job.class
+   * @param apiListTypeClass the api list type class e.g V1JobList.class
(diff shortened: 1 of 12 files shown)


yue9944882:
note that #86 is solved in this pull, as is shown in the test..

yue9944882:
/hold 

until #742 in

brendandburns:
@yue9944882 is this ready for review?

yue9944882:
@brendandburns let's freeze the pull until the migration to openapi-generator, the work isn't necessary to get into the next release i think.

yue9944882:
/hold cancel

@brendandburns let's move on reviewing this pull. @runzexia feel free to drop your comments too

yue9944882:
@brendandburns bumping for review

extended/src/main/java/io/kubernetes/client/extended/generic/GenericKubernetesApi.java
```diff
@@ -0,0 +1,680 @@
+package io.kubernetes.client.extended.generic;
+
+import com.google.common.base.Strings;
+import com.google.gson.Gson;
+import com.google.gson.JsonElement;
+import com.google.gson.reflect.TypeToken;
+import io.kubernetes.client.custom.V1Patch;
+import io.kubernetes.client.extended.generic.options.*;
```
brendandburns: Please don't use wildcard imports.

extended/src/main/java/io/kubernetes/client/extended/generic/GenericKubernetesApi.java
```diff
@@ -0,0 +1,680 @@
...
+ */
+public class GenericKubernetesApi<ApiType, ApiListType> {
+
+  // TODO(yue9944882): supports status operations..
+  // TODO(yue9944882): supports generic sub-resoruce operations..
+  // TODO(yue9944882): supports delete-collections..
+
+  private Class<ApiType> apiTypeClass;
+  private Class<ApiListType> apiListTypeClass;
+  private String apiGroup;
```
brendandburns: Does it make sense to use the native Kubernetes object here for this? It has these fields.

extended/src/main/java/io/kubernetes/client/extended/generic/GenericKubernetesApi.java
```diff
@@ -0,0 +1,680 @@
...
+  public GenericKubernetesApi(
+      Class<ApiType> apiTypeClass,
+      Class<ApiListType> apiListTypeClass,
+      String apiGroup,
+      String apiVersion,
+      String resourcePlural,
+      CustomObjectsApi customObjectsApi) {
+    if ("".equals(apiGroup)) {
+      throw new IllegalArgumentException(
+          "core group (\"\") not supported, use CoreV1Api otherwise");
```
brendandburns: should this say `use "v1" instead`

extended/src/main/java/io/kubernetes/client/extended/generic/GenericKubernetesApi.java
```diff
@@ -0,0 +1,680 @@
...
+    V1ObjectMeta objectMeta;
+    try {
+      objectMeta = Reflect.objectMetadata(object);
+    } catch (ObjectMetaReflectException e) {
+      throw new IllegalArgumentException("fail to extract object metadata");
+    }
+    return executeCall(
+        customObjectsApi.getApiClient(),
+        apiTypeClass,
+        () -> {
```
brendandburns: This function looks like it's duplicated with above? If so, pull out with a single implementation.

extended/src/main/java/io/kubernetes/client/extended/generic/options/CreateOptions.java
```diff
@@ -0,0 +1,7 @@
+package io.kubernetes.client.extended.generic.options;
+
+public class CreateOptions {
+  private Boolean dryRun;
+  private Object
+      fieldManager; // TODO: structured field-manager for better server-side apply integeration
```
brendandburns: why is this on a newline?

extended/src/main/java/io/kubernetes/client/extended/generic/options/DeleteOptions.java
```diff
@@ -0,0 +1,5 @@
+package io.kubernetes.client.extended.generic.options;
+
+import io.kubernetes.client.openapi.models.V1DeleteOptions;
+
+public class DeleteOptions extends V1DeleteOptions {}
```
brendandburns: what is the purpose of this class?

extended/src/main/java/io/kubernetes/client/extended/generic/options/ListOptions.java
```diff
@@ -0,0 +1,71 @@
+package io.kubernetes.client.extended.generic.options;
```
brendandburns: Shouldn't we use the generated class?

extended/src/main/java/io/kubernetes/client/extended/generic/GenericKubernetesApi.java
yue9944882: fixed

extended/src/main/java/io/kubernetes/client/extended/generic/GenericKubernetesApi.java
yue9944882: updated to `core group (\"v1\") api not supported, use CoreV1Api instead`, does that sound better?

extended/src/main/java/io/kubernetes/client/extended/generic/GenericKubernetesApi.java
yue9944882: the called function in the lambda has similar name but different:

replaceNamespacedCustomObjectCall vs replaceClusterCustomObjectCall

extended/src/main/java/io/kubernetes/client/extended/generic/options/CreateOptions.java
yue9944882: it's the TODO, i moved it to another line

extended/src/main/java/io/kubernetes/client/extended/generic/options/DeleteOptions.java
yue9944882: i added a class named `ListOptions` in this package and it's probably good to gather those option class into the same package. the `DeleteOptions` here is just a link to the generated delete option class. wdyt?

extended/src/main/java/io/kubernetes/client/extended/generic/options/ListOptions.java
yue9944882: there is a new class named `ListOption` introduced by this pull. the point of defining option classes here is to provide the developers a better programming interface, which looks making more logical sense comparing to those generated long-list parameters calls. in golang, we have similar option structs defined over the actual API parameters, developer works on that logical options and the SDK translates them to api interface options.. does that make sense to you?

brendandburns:
@yue9944882 fwiw, I'm waiting on responses to the other comments before re-reviewing.

Let me know when it's ready for another pass.

Thanks!

yue9944882:
@brendandburns am i missing anything? i seem to replied the comments on the page. please let me know, thx!

brendandburns:
Was wondering about this comment:

https://github.com/kubernetes-client/java/pull/738#discussion_r363888814

extended/src/main/java/io/kubernetes/client/extended/generic/GenericKubernetesApi.java
yue9944882: are you refering to [V1APIGroup](https://github.com/kubernetes-client/java/blob/master/kubernetes/src/main/java/io/kubernetes/client/openapi/models/V1APIGroup.java)? the so-called field "apiGroup" here should be the name of the serving resource api group e.g. `apps`, `networking.k8s.io`, `apiextensions.k8s.io`, etc. while the `V1APIGroup` is used for representing all the fundamental information about the an api group . we need only a tiny bit of that, which is V1APIGroup#name. simply a string value would be sufficient, wdyt?

brendandburns:
I guess it's fine, the plurals and everything else are represented in that object too, afaik.

But this seems like a fine place to start.

/lgtm
/approve

PR #741, opened 2019-10-18
Added examples for create StatefulSet from YAML file

### LINKED ISSUE #740: Added example for create StatefulSet from YAML file
There is a PatchExample.java for create a Deployment, but it does not uses yaml from file.
I am not sure to write a working file on `examples/`, but I think this will be a simple explain.

seongwon-kang:
Resolves #740

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,15 +181,15 @@ public class CreateTest {
 
 		/*Example yaml file can be found in examples/test-sts.yaml*/
 		File file = new File("test-sts.yaml");
-		V1StatefulSet pod;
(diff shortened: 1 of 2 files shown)


brendandburns:
This example is a little esoteric for the main README.md.

Also, I think it is covered by the [YamlExample.java](https://github.com/kubernetes-client/java/blob/master/examples/src/main/java/io/kubernetes/client/examples/YamlExample.java)

I'm going to close this PR, if you would like to reopen it for some reason, please use the `/reopen` command.

PR #742, opened 2019-10-18
Re-generate CustomObjectApi to support pagination

yue9944882:
regenerate by following the instrcutions from kubernetes-client/gen


yue9944882:
ref: https://github.com/kubernetes-client/gen/pull/130
ref: https://github.com/kubernetes-client/gen/pull/119

brendandburns:
This is a breaking change (albeit a small one) how do we feel about things like this?

Also, can you separate the build-fix PR into a different PR

yue9944882:
SGTM, separated

brendandburns:
closing and reopen to trigger travis.

brendandburns:
@yue9944882 what do you think about the fact that this is a breaking change?

yue9944882:
@brendandburns are you suggesting we should merging this together w/ #709? instead of one dedicated pull?

brendandburns:
I think we should move to OpenAPI generator as part of the `7.0.0` (kubernetes `1.16.x`) release...

Or at the very least, if we're going to make this change, I think we should rev. to `6.1.x` to indicate that it's a bigger change than just bug fixes...

yue9944882:
> I think we should move to OpenAPI generator as part of the 7.0.0 (kubernetes 1.16.x) release...

for the record, 7.0.0 aligns w/ 1.15 release of k/k. or are we deciding skipping 1.15?

> Or at the very least, if we're going to make this change, I think we should rev. to 6.1.x to indicate that it's a bigger change than just bug fixes...

+1 on your concern about the compatibility, since this pull is tweaking public interfaces. i think the openapi-generator pull is taking off soon. closing this in favor of the migration pull.

PR #745, opened 2019-10-21
Chore: Fix build flakiness

yue9944882:

PR #748, opened 2019-10-22
UTF-8 support for Config class.

sarveshkaushal:
Looked at another TODO item in Config file. Please take a look at code change.

### DIFF
--- a/util/src/main/java/io/kubernetes/client/util/Config.java
+++ b/util/src/main/java/io/kubernetes/client/util/Config.java
@@ -76,13 +76,13 @@ public class Config {
   }
 
   public static ApiClient fromConfig(String fileName) throws IOException {
-    KubeConfig config =
-        KubeConfig.loadKubeConfig(
-            new BufferedReader(
-                new InputStreamReader(
-                    new FileInputStream(fileName), StandardCharsets.UTF_8.name())));
-    config.setFile(new File(fileName));
-    return fromConfig(config);
+    try (BufferedReader bufferedReader =
+        new BufferedReader(
+            new InputStreamReader(new FileInputStream(fileName), StandardCharsets.UTF_8.name()))) {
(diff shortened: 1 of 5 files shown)


util/src/main/java/io/kubernetes/client/util/Config.java
```diff
@@ -74,13 +76,17 @@ public static ApiClient fromToken(String url, String token, boolean validateSSL)
...
 
   public static ApiClient fromConfig(String fileName) throws IOException {
-    KubeConfig config = KubeConfig.loadKubeConfig(new FileReader(fileName)); // TODO UTF-8
+    KubeConfig config =
+        KubeConfig.loadKubeConfig(
+            new BufferedReader(
+                new InputStreamReader(
+                    new FileInputStream(fileName), StandardCharsets.UTF_8.name())));
     config.setFile(new File(fileName));
     return fromConfig(config);
```
brendandburns: We should really close the `FileInputStream` here.  Perhaps:

```java
try (FileInputStream fi = new FIleInputStream(fileName)) {
   KubeConfig config = KubeConfig.loadKubeConfig(...);
   return  fromConfig(config);
}
```

brendandburns:
Thanks for the PR, one small improvement requested.

util/src/main/java/io/kubernetes/client/util/Config.java
sarveshkaushal: Yes try-with-resources is correct way of doing this here. Thanks for your input. I will change it.

sarveshkaushal:
@brendanburns - I have added the changes as per your input. Thanks for pointing it out. Please take a look.

yue9944882:
the flakiness is connected w/ https://github.com/tomakehurst/wiremock/issues/515. sadly #745 doesnt actually solve that

sarveshkaushal:
Closing this. Will open again with latest changes from master.

PR #749, opened 2019-10-22
Replaced hard-coded string with ApiregistrationV1beta1ServiceReference.class

hpandeycodeit:
Replaced hard-coded string with ApiregistrationV1beta1ServiceReference.class


hpandeycodeit:
I just noticed there are several java classes with the same hard-coded class name. If it's okay I will go ahead and clean them up as a part of this PR.

yue9944882:
can you show us the difference between the two approaches? btw those code are machine generated, you need to edit on the code-template then re-gen.

hpandeycodeit:
No difference actually, just more from the java coding standard point of view. Also, since it's auto-generated this need not be corrected. So feel free to close this PR

brendandburns:
This is generated code (and we're about to switch the generator) if you want to fix this, it should be done in the openapi code generator project https://github.com/OpenAPITools/openapi-generator/tree/master/modules/openapi-generator/src/main/resources/Java.

I'm going to close this, since we shouldn't modify generated code in this project.

PR #754, opened 2019-10-24
Using GzipCompressorInputStream for decompressing .gz files.

sarveshkaushal:
Very small change. This is for another TODO comment.

### DIFF
--- a/util/src/main/java/io/kubernetes/client/Copy.java
+++ b/util/src/main/java/io/kubernetes/client/Copy.java
@@ -117,7 +117,7 @@ public class Copy extends Exec {
         this.exec(
             namespace,
             pod,
-            new String[] {"sh", "-c", "tar cf - " + srcPath + " | base64"},
+            new String[] {"sh", "-c", "tar cz - " + srcPath + " | base64"},
             container,
             false,
             false);
--- a/util/src/main/java/io/kubernetes/client/Copy.java
+++ b/util/src/main/java/io/kubernetes/client/Copy.java
@@ -30,6 +30,7 @@ import org.apache.commons.compress.archivers.ArchiveOutputStream;
 import org.apache.commons.compress.archivers.tar.TarArchiveEntry;
 import org.apache.commons.compress.archivers.tar.TarArchiveInputStream;
 import org.apache.commons.compress.archivers.tar.TarArchiveOutputStream;
+import org.apache.commons.compress.compressors.gzip.GzipCompressorInputStream;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -121,8 +122,7 @@ public class Copy extends Exec {
             false,
             false);
     try (InputStream is = new Base64InputStream(new BufferedInputStream(proc.getInputStream()));
-        ArchiveInputStream archive = new TarArchiveInputStream(is)) {
-      // TODO Use new GzipCompressorInputStream(is))) here {
+        ArchiveInputStream archive = new TarArchiveInputStream(new GzipCompressorInputStream(is))) {
       for (ArchiveEntry entry = archive.getNextEntry();
           entry != null;
           entry = archive.getNextEntry()) {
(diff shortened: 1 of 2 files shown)


util/src/main/java/io/kubernetes/client/Copy.java
```diff
@@ -121,8 +122,7 @@ public void copyDirectoryFromPod(
             false,
             false);
     try (InputStream is = new Base64InputStream(new BufferedInputStream(proc.getInputStream()));
-        ArchiveInputStream archive = new TarArchiveInputStream(is)) {
-      // TODO Use new GzipCompressorInputStream(is))) here {
+        ArchiveInputStream archive = new TarArchiveInputStream(new GzipCompressorInputStream(is))) {
```
yue9944882: i think we need to enable the gzip compression on the remote process too, by applying "z" option? additionally, we need a test for the change.

brendandburns: Yep, we definitely need it on both sides, the current encoding is just 'tar'

sarveshkaushal: I think
final Process proc =
        this.exec(
            namespace,
            pod,
            new String[] {"sh", "-c", "tar cf**z** - " + srcPath + " | base64"},
            container,
            false,
            false);
Am I thinking right?

sarveshkaushal:
I have added a test case for the change. Also enabled zipping on remote process. Please take a look.

util/src/main/java/io/kubernetes/client/Copy.java
yue9944882: yes, but you dropped "f", is that intended?

util/src/test/java/io/kubernetes/client/CopyTest.java
```diff
@@ -127,4 +129,53 @@ public void run() {
...
+
+    Copy copy = new Copy(client);
+
+    stubFor(
+        get(urlPathEqualTo("/api/v1/namespaces/" + namespace + "/pods/" + podName + "/exec"))
+            .willReturn(
+                aResponse()
+                    .withStatus(404)
+                    .withHeader("Content-Type", "application/json")
+                    .withBody("{}")));
```
yue9944882: ahh, so sry for oversight.. the exec call works basing on websocket, there's no way to mock that out via wiremock framework.. i will verify this pull locally. we should remove the test

util/src/main/java/io/kubernetes/client/Copy.java
sarveshkaushal: Yes- we don't need "f". As we are directly getting the stream from process' stdout.

util/src/test/java/io/kubernetes/client/CopyTest.java
sarveshkaushal: Let me remove the test.

sarveshkaushal: Test case is removed.

brendandburns:
This looks fine to me. Is it covered by existing test code?

sarveshkaushal:
I see a test case for method copyFileToPod but don't see any other test case for copyDirectoryFromPod. However currently test for copyFileToPod is ignored because it was causing instability in the build.

util/src/main/java/io/kubernetes/client/Copy.java
yue9944882: iiuc, if you drop "f", you will also have to drop the tailing "-" from the command. in the previous command `tar cf - ...` where "-" is the argument for "f" from the bundled options. i suggest keep the "f" for clarification given that it's a more commonly used form.

brendandburns:
ok, we can merge this as is. We should fix the tests too...

sarveshkaushal:
Thanks @brendandburns .I can take a look at test case for copyFileToPod but need to little more details about build. I will investigate and create a separate PR for it.

yue9944882:
> new String[] {"sh", "-c", "tar cfz - " + srcPath + " | base64"},

https://github.com/kubernetes-client/java/pull/754#discussion_r339897994 i dont think this is a legitimate tar command

Issue #628, opened 2019-07-18
informer ListerWatcher should support pagination


adohe:
for ReflectorRunnable init list, it should use pagination list. Perhaps we could add a PaginationListerWatcher which supports pagination list, then other ListerWatcher could extend this one.

yue9944882:
/kind feature

users can implement their ListerWatcher in a paginated way, let's hear more voice then

PR #757, opened 2019-10-25
Enhance Informer

rjbhewei:
1. informer list request support pagination like client-go(default limit 500).  

    eg: PageInformerExample.java

    fix issue : https://github.com/kubernetes-client/java/issues/628

2. informer local-cache filter resource fields which customer definition，to ease server-side JVM memory usage，avoid oom.

    eg: FieldFilterPageInformerExample.java


extended/pom.xml
```diff
@@ -42,6 +42,12 @@
             <artifactId>bucket4j-core</artifactId>
             <version>${bucket4jVersion}</version>
         </dependency>
+        <dependency>
```
brendandburns: These pom changes seem unrelated. Are they necessary?

extended/src/main/java/io/kubernetes/client/extended/informer/FilterField.java
```diff
@@ -0,0 +1,48 @@
+package io.kubernetes.client.extended.informer;
+
+import org.apache.commons.lang3.StringUtils;
+
+public class FilterField {
+
+  // the name of the field
```
brendandburns: delete redundant comments. (here and below)

extended/src/main/java/io/kubernetes/client/extended/informer/FilterFunction.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.util.List;
+import java.util.Objects;
+
+@FunctionalInterface
+public interface FilterFunction<T> {
+
+  T apply(T t);
+
+  default FilterFunction<T> compose(FilterFunction<T> before) {
```
brendandburns: I think these methods need JavaDoc

extended/src/main/java/io/kubernetes/client/extended/informer/PageSharedInformerFactory.java
```diff
@@ -0,0 +1,86 @@
...
+
+        return pager.getListObjectCurrentPage();
+      }
+
+      @Override
+      public Watch<ApiType> watch(CallGeneratorParams callGeneratorParams) throws ApiException {
+
+        Call call = callGenerator.generate(PagerCallGeneratorParams.to(callGeneratorParams));
+
+        Type watchType = TypeToken.getParameterized(Watch.Response.class, apiTypeClass).getType();
```
brendandburns: Perhaps move this to the constructor? It's never going to change.

extended/src/main/java/io/kubernetes/client/extended/informer/SerializeFilterJSON.java
```diff
@@ -0,0 +1,131 @@
...
+import java.util.Map;
+import java.util.function.BiPredicate;
+import org.apache.commons.collections4.CollectionUtils;
+import org.apache.commons.collections4.ListUtils;
+import org.apache.commons.lang3.StringUtils;
+import org.joda.time.DateTime;
+import org.joda.time.LocalDate;
+
+/**
+ * enhance default JSON，which serialize list/watch objects，SerializeFilterJSON enhance common object
```
brendandburns: which serializes list/watch objects.

brendandburns: Clarify second sentence, it's hard to understand.

extended/src/main/java/io/kubernetes/client/extended/informer/SerializeFilterJSON.java
```diff
@@ -0,0 +1,131 @@
...
+    if (CollectionUtils.isEmpty(filterFields)) {
+      return null;
+    }
+
+    return new ExclusionStrategy() {
+
+      @Override
+      public boolean shouldSkipField(FieldAttributes f) {
+
+        SerializedName serialized = f.getAnnotation(SerializedName.class);
```
brendandburns: This can be moved into the constructor right?

extended/src/main/java/io/kubernetes/client/extended/informer/SerializeFilterJSON.java
```diff
@@ -0,0 +1,131 @@
...
+
+        FilterField filterField =
```
brendandburns: This too?

extended/src/main/java/io/kubernetes/client/extended/informer/SerializeFilterJSON.java
```diff
@@ -0,0 +1,131 @@
...
+      @Override
+      public boolean shouldSkipClass(Class<?> clazz) {
+        return false;
+      }
+    };
+  }
+
+  private static final BiPredicate<FilterField, FilterField> MATCH_PREDICATE =
+      (customer, system) -> {
+        if (!customer.isNotEmpty()) { // double check
```
brendandburns: double negatives are hard to understand. why not `if (customer.isEmpty()) {`

extended/src/main/java/io/kubernetes/client/extended/informer/SerializeFilterJSON.java
```diff
@@ -0,0 +1,131 @@
...
+          return false;
+        }
+
+        boolean isSameName =
+            StringUtils.isEmpty(customer.getName())
```
brendandburns: why is it the same if the customer name is empty? what if the system name is not?

extended/src/main/java/io/kubernetes/client/extended/informer/SerializeFilterJSON.java
```diff
@@ -0,0 +1,131 @@
...
+
+      @Override
+      public boolean shouldSkipClass(Class<?> clazz) {
+        return false;
+      }
+    };
+  }
+
+  private static final BiPredicate<FilterField, FilterField> MATCH_PREDICATE =
+      (customer, system) -> {
```
brendandburns: I'm not sure why one is `customer` and one is `system`

util/src/main/java/io/kubernetes/client/util/CallGenerator.java
```diff
@@ -13,13 +13,13 @@
  * params.resourceVersion, params.timeoutSeconds, params.watch, null, null); },
  */
 @FunctionalInterface
-public interface CallGenerator {
+public interface CallGenerator<T extends CallGeneratorParams> {
```
brendandburns: Why is templating necessary here? It seems that polymorphism would take care of it.

rjbhewei:
I wanted to enhance the new features by changing the original code as little as possible.Without affecting the original function。

original code use io.kubernetes.client.JSON serializes list/watch objects. but I need to ease server-side JVM memory usage，when serialize objects customer can definition filter some filter，so impl SerializeFilterJSON to do this

util/src/main/java/io/kubernetes/client/util/CallGenerator.java
rjbhewei: PageSharedInformerFactory use PagerCallGeneratorParams ,but ListerWatcher  list method hardcode  CallGeneratorParams

extended/src/main/java/io/kubernetes/client/extended/informer/FilterField.java
rjbhewei: delete now

extended/src/main/java/io/kubernetes/client/extended/informer/SerializeFilterJSON.java
rjbhewei: add isEmpty() method now

extended/src/main/java/io/kubernetes/client/extended/informer/SerializeFilterJSON.java
rjbhewei: customer definition filterField  not use the "name" attribute，but use "declaringClass" or "serializedName"

PR #786, opened 2019-11-08
Resetting wiremock to fix flaky test

sarveshkaushal:


util/src/test/java/io/kubernetes/client/CopyTest.java
```diff
@@ -80,16 +78,14 @@ public void testUrl() throws IOException, ApiException, InterruptedException {
...
-  // root cause is located
   public void testCopyFileToPod() throws IOException, ApiException, InterruptedException {
 
     File testFile = File.createTempFile("testfile", null);
     testFile.deleteOnExit();
 
     Copy copy = new Copy(client);
 
-    stubFor(
+    wireMockRule.stubFor(
```
yue9944882: sry to be dense. can you elaborate how this can help resolve the flakiness? thanks

sarveshkaushal: I was able to repro the build failure when i just tried to build util module with OpenJDK 11. Please note i build just Util module. It doesn't repro even with Java 11 if i run mvn test for CopyTest class. It only shows up when entire module is tested. This doesn't repro at all with Java 8 whether you build entire module or just CopyTest. I am not an expert in WireMock but i feel it has something to do with registered stubs in the memory. My change will reset the WireMock results in memory before executing this particular test in CopyTest class. Please let me know your thoughts.

brendandburns: This change seems fine to me, and if it fixes the flaky behavior that's great.  I can see that moving from a static method to a instance method could fix a flake. So it's plausible and I don't see any harm in merging this.

sarveshkaushal: Thanks @brendandburns . If this is acceptable then I would also like to uncomment the other test case in CopyTest class and use similar approach.

brendandburns: Sure, that works for me. Thanks!

sarveshkaushal: I will create an another PR for another test case after this is merged. Thanks.

PR #793, opened 2019-11-12
Chore: Split reflect utils to [Object|List]accessor for clarification

yue9944882:
separating these util functions helps us sorting them by their usages. and also this aligns w/ the naming from upstream client-go so that users can find them more easily.

### DIFF
--- a/util/src/main/java/io/kubernetes/client/util/ListAccessor.java
+++ b/util/src/main/java/io/kubernetes/client/util/ListAccessor.java
@@ -1,0 +1,32 @@
+package io.kubernetes.client.util;
+
+import io.kubernetes.client.openapi.models.V1ListMeta;
+import io.kubernetes.client.util.exception.ObjectMetaReflectException;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.List;
+
+/** ListAccessor is a set of utils which accesses common fields from list objects dynamically */
+public class ListAccessor {
+
+  private static final String METHOD_NAME_GET_ITEMS = "getItems";
+  private static final String METHOD_NAME_GET_LIST_METADATA = "getMetadata";
+
+  public static V1ListMeta listMetadata(Object listObj) throws ObjectMetaReflectException {
+    try {
+      Method mdField = listObj.getClass().getMethod(METHOD_NAME_GET_LIST_METADATA);
+      return (V1ListMeta) mdField.invoke(listObj);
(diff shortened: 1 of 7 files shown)


util/src/main/java/io/kubernetes/client/util/ListAccessor.java
```diff
@@ -0,0 +1,32 @@
...
+import io.kubernetes.client.openapi.models.V1ListMeta;
+import io.kubernetes.client.util.exception.ObjectMetaReflectException;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.List;
+
+/** ListAccessor is a set of utils which accesses common fields from list objects dynamically */
+public class ListAccessor {
+
+  public static final String METHOD_NAME_GET_ITEMS = "getItems";
```
brendandburns: Make these private?

brendandburns: Given that these strings are only used once, are constants needed?

util/src/main/java/io/kubernetes/client/util/ObjectAccessor.java
```diff
@@ -0,0 +1,29 @@
...
+
+import io.kubernetes.client.openapi.models.V1ObjectMeta;
+import io.kubernetes.client.util.exception.ObjectMetaReflectException;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+
+/** ObjectAccessor is a set of utils which accesses common fields from objects dynamically */
+public class ObjectAccessor {
+
+  public static final String METHOD_NAME_GET_METADATA = "getMetadata";
```
brendandburns: same comment wrt to whether this is needed.

util/src/main/java/io/kubernetes/client/util/ListAccessor.java
yue9944882: updated to private

util/src/main/java/io/kubernetes/client/util/ObjectAccessor.java
yue9944882: updated to private

util/src/main/java/io/kubernetes/client/util/ListAccessor.java
yue9944882: separate hardcodes as const is aesthetically pleasing, they don't really make a difference tho

PR #794, opened 2019-11-12
Uncommenting the test and changing wireMockRule to be used in instanc…

sarveshkaushal:
…e method.
This is as per our discussion on #786 .


brendandburns:
Works for me, we can wait.

brendandburns:
I think that this seems to have fixed any flakiness? Shall we merge this?

yue9944882:
agree, let's proceed

PR #799, opened 2019-11-13
Add codegen test to constraint behaviors from generator

yue9944882:

PR #804, opened 2019-11-18
Re-gen from generator release 4.2.1

yue9944882:
This will make #799 work


brendandburns:
Took a glance through, looks sane.

Tests are failing though

yue9944882:
@brendandburns after digging into the CI failure, i found that using reflective deep equal will fail the  equality of other foundamental types e.g. `java.lang.String`. the reflective way is comparing equality of every member fields of a class even though the field is completely internal e.g. `java.lang.String#hash`. i sent https://github.com/kubernetes-client/gen/pull/139 to revert the enabling the reflective. fwiw, we're probably not getting https://github.com/kubernetes-client/java/issues/431 fixed in 7.0.0. wdyt?

PR #807, opened 2019-11-20
Extract thread pool shutdown to a separate method

ahadas:
The current API of SharedInformerFactory supports two cases
in regards to the thread pool it uses inside:
1. Default thread pool - if no thread pool is specified, a new
thread pool is instantiated internally.
2. Provided thread pool - if a thread pool is specified, it
is the one used internally by this class.

Seems that the purpose of (2) is to enable users to provide
their own thread pool, configured the way they want rather than
relying on the thread pool that is instantiated internally.

However, one may think that (2) also enables the caller to pass
a thread pool that is shared among other components in their
application. The issue is that when calling the method
stopAllRegisteredInformers, it shuts down the thread pool and
that may not necessarily be the desired behavior in this case.

This patch introduces a way to signal that the thread pool should
not be shut down when removing all the registered informers.

### DIFF
--- a/util/src/main/java/io/kubernetes/client/informer/SharedInformerFactory.java
+++ b/util/src/main/java/io/kubernetes/client/informer/SharedInformerFactory.java
@@ -178,8 +178,17 @@ public class SharedInformerFactory {
         });
   }
 
-  /** Stop all registered informers. */
+  /** Stop all registered informers and shut down the thread pool. */
   public synchronized void stopAllRegisteredInformers() {
+    stopAllRegisteredInformers(true);
+  }
+
+  /**
+   * Stop all registered informers.
+   *
+   * @param shutdownThreadPool whether or not to shut down the thread pool.
+   */
+  public synchronized void stopAllRegisteredInformers(boolean shutdownThreadPool) {
     if (MapUtils.isEmpty(informers)) {
       return;
     }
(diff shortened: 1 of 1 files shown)


brendandburns:
Hrm, this syntax seems confusing. I'd rather have the user pass in an explicit boolean to shut it down (or not)

yue9944882:
>  I'd rather have the user pass in an explicit boolean to shut it down (or not)

i think so.. it's a bit overkilling to add a new protected method just for skipping shutting-down the thread-pool. or is there anything i missed? @ahadas how about adding a new method `stopAllRegisteredInformers(boolean shutdownInformerExecutors)`?

ahadas:
Thanks for the feedback!
Well, it wouldn't be the way I would choose to do it since the API becomes a bit confusing - as I user I may instantiate this factory without providing a thread-pool, thus letting this class to instantiate a default one, and yet ask not to shut down the thread-pool when stopping all informers. In contrast, with the previous approach, the API of this class keeps assuming a dedicated thread-pool (passed to the instance of this class) and one should "work hard" to change that.
But we can do it that way.

ahadas:
> @ahadas how about adding a new method `stopAllRegisteredInformers(boolean shutdownInformerExecutors)`?

Done
I just named the parameter shutdownThreadPool because in all signatures the executor is named thread pool at the moment

PR #818, opened 2019-12-02
enhance informer

rjbhewei:
1. informer list request support pagination like client-go(default limit 500).  

    eg: PageInformerExample.java

    fix issue : https://github.com/kubernetes-client/java/issues/628

2. informer local-cache filter resource fields which customer definition，to ease server-side JVM memory usage，avoid oom.

    eg: FieldFilterPageInformerExample.java

### DIFF
--- a/extended/pom.xml
+++ b/extended/pom.xml
@@ -42,6 +42,12 @@
             <artifactId>bucket4j-core</artifactId>
             <version>${bucket4jVersion}</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-collections4</artifactId>
+            <version>${commons-collections4.version}</version>
+        </dependency>
+
         <!-- test dependencies -->
         <dependency>
             <groupId>junit</groupId>
@@ -67,15 +73,25 @@
             <version>2.19.0</version>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>org.apache.commons</groupId>
-            <artifactId>commons-collections4</artifactId>
-            <scope>test</scope>
-        </dependency>
+
     </dependencies>
 
     <build>
         <plugins>
+            <plugin>
+                <artifactId>maven-source-plugin</artifactId>
+                <configuration>
+                    <attach>true</attach>
+                </configuration>
+                <executions>
+                    <execution>
+                        <phase>compile</phase>
+                        <goals>
+                            <goal>jar</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
             <plugin>
                 <groupId>org.jacoco</groupId>
                 <artifactId>jacoco-maven-plugin</artifactId>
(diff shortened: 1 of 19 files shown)


extended/pom.xml
```diff
@@ -136,10 +152,14 @@
         </plugins>
     </build>
     <properties>
+        <maven.compiler.target>1.7</maven.compiler.target>
```
brendandburns: Please preserve the existing parts w/ the variables as much as possible.

extended/src/main/java/io/kubernetes/client/extended/informer/FilterField.java
```diff
@@ -0,0 +1,49 @@
+package io.kubernetes.client.extended.informer;
+
+import org.apache.commons.lang3.StringUtils;
+
+public class FilterField {
+
+  private String name;
+
```
brendandburns: Spacing is weird here. Remove extra newlines?

extended/src/main/java/io/kubernetes/client/extended/informer/PageSharedInformerFactory.java
```diff
@@ -0,0 +1,118 @@
...
+  public PageSharedInformerFactory(ApiClient client, ExecutorService threadPool) {
+    super(client, threadPool);
+  }
+
+  protected <ApiType, ApiListType> ListerWatcher<ApiType, ApiListType> listerWatcherFor(
+      final CallGenerator callGenerator,
+      final Class<ApiType> apiTypeClass,
+      final Class<ApiListType> apiListTypeClass) {
+
+    if (apiClient.getHttpClient().readTimeoutMillis() > 0) {
```
brendandburns: Hrm, do we want to do this? Or do we want throw with `IlleagalArgumentException` silently changing the timeout feels weird to me.

extended/src/main/java/io/kubernetes/client/extended/informer/PageSharedInformerFactory.java
```diff
@@ -0,0 +1,118 @@
...
+      // set read timeout zero to ensure client doesn't time out
+      OkHttpClient httpClient =
+          apiClient.getHttpClient().newBuilder().readTimeout(0, TimeUnit.MILLISECONDS).build();
+      apiClient.setHttpClient(httpClient);
+    }
+
+    return new ListerWatcher<ApiType, ApiListType>() {
```
brendandburns: This feels like a pretty big chunk of code to be an anonymous class, perhaps break it out as it's own class? It's pretty hard to test it as is.

brendandburns:
The Field Filter code seems separate from the paging code. Can you split this into 2 PRs?

Thanks

fejta:
/hold
testing

fejta:
/hold cancel
thanks

extended/src/main/java/io/kubernetes/client/extended/informer/PageSharedInformerFactory.java
rjbhewei: same with [SharedInformerFactory](https://github.com/kubernetes-client/java/blob/69ca6fd7cca4f8188ec9b490530e53cc9b0dea09/util/src/main/java/io/kubernetes/client/informer/SharedInformerFactory.java#L131-L136)

extended/src/main/java/io/kubernetes/client/extended/informer/PageSharedInformerFactory.java
rjbhewei: same with [SharedInformerFactory](https://github.com/kubernetes-client/java/blob/69ca6fd7cca4f8188ec9b490530e53cc9b0dea09/util/src/main/java/io/kubernetes/client/informer/SharedInformerFactory.java#L131-L136)

extended/src/main/java/io/kubernetes/client/extended/informer/PageSharedInformerFactory.java
rjbhewei: PageSharedInformerFactory  and PagerCallGeneratorParams use like client-go, add page default limit size (500)，this factory use for scale data。

I wanted to enhance the new features by changing the original code as little as possible.Without affecting the original function。

submit the basic implementation first, and then change other related implementations。

yue9944882:
plz split the pull by their intention

PR #826, opened 2019-12-10
Re-generate from kubernetes 1.16.4

yue9944882:
related code-generator changes: https://github.com/kubernetes-client/gen/pull/140


brendandburns:
Did this require any non-generated code changes? If so, please add them as a separate commit. If not LGTM.

yue9944882:
i splitted another commit for manually removing the previously generated tests

PR #827, opened 2019-12-10
Introduce oneline patch util for overriding patch format from generated client

yue9944882:
/cc @brendandburns


brendandburns:
/lgtm
/approve

Perhaps add an example of how to use this as a follow up PR?

PR #834, opened 2019-12-16
Implement DefaultSharedIndexInformer#addIndexers

shakeelrao:
Implements `addIndexers` by modifying the `Indexer` interface to match `client-go`.  The semantics of `addIndexers` remains consistent with `client-go` as well.

References:
1) https://github.com/kubernetes/client-go/blob/master/tools/cache/shared_informer.go#L362
2) https://github.com/kubernetes/client-go/blob/master/tools/cache/index.go#L48
3) https://github.com/kubernetes/client-go/blob/master/tools/cache/thread_safe_store.go#L228

### DIFF
--- a/util/src/main/java/io/kubernetes/client/informer/cache/Cache.java
+++ b/util/src/main/java/io/kubernetes/client/informer/cache/Cache.java
@@ -320,6 +320,38 @@ public class Cache<ApiType> implements Indexer<ApiType> {
   }
 
   /**
+   * Return the indexers registered with the cache.
+   *
+   * @return registered indexers
+   */
+  @Override
+  public Map<String, Function<ApiType, List<String>>> getIndexers() {
+    return indexers;
+  }
+
+  /**
+   * Add additional indexers to the cache.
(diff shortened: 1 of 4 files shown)


shakeelrao:
Signed.

shakeelrao:
I signed it

shakeelrao:
I have signed it.

shakeelrao:
@yue9944882 Thanks! What are you thoughts on throwing unchecked exceptions here? I wasn't entirely sure how best to port go error handling to Java.

yue9944882:
> throw new IllegalStateException("cannot add indexers to a running informer");

unchecked exception can bring bad UX to those who try adding indexers dynamically to a running informer. but on the other side, i suppose declaring a checked exception here can probably bring more confusion for the starter developers trying the informer. given the fact that customizing indexers is completely optional and expecting developers to have a basic understanding of underlying mechnism. i'd vote on the uncheck.

yue9944882:
getting this in

/approve

PR #838, opened 2019-12-17
Re-gen to make custom object delete body optional

yue9944882:
xref: https://github.com/kubernetes-client/gen/pull/142


kubernetes/src/main/java/io/kubernetes/client/openapi/apis/CustomObjectsApi.java
```diff
@@ -456,13 +456,8 @@ public Object createNamespacedCustomObject(String group, String version, String
             throw new ApiException("Missing the required parameter 'name' when calling deleteClusterCustomObject(Async)");
         }
         
-        // verify the required parameter 'body' is set
-        if (body == null) {
-            throw new ApiException("Missing the required parameter 'body' when calling deleteClusterCustomObject(Async)");
```
yue9944882: highlighting the non-trival changes

brendandburns:
This is a breaking change, so this will only go out as part of `8.0.0`

PR #848, opened 2020-01-07
Builder-modes for attach() and exec()

### LINKED ISSUE #847: Attach standardOutputStream not working
Several issues have been opened about this, to later be auto-closed.

I investigated this, and found the culprit to be "stdout=true" needs to be added to URL to activate stdout streaming. I suspect it was once the default, server-side, but now needs to be explicit.

For me, I fixed it temporarily by copying Attach.java to local file and applying this;

```
--- Attach.java.old	2020-01-07 14:19:53.819452472 +0100
+++ Attach.java	2020-01-07 14:20:28.055836186 +0100
@@ -59,6 +59,7 @@
 
   private String makePath(
       String namespace, String name, String container, boolean stdin, boolean tty) {
+    boolean stdout = true;
     return "/api/v1/namespaces/"
         + namespace
         + "/pods/"
@@ -66,6 +67,8 @@
         + "/attach?"
         + "stdin="
         + stdin
+	+ "&stdout="
+	+ stdout
         + "&tty="
         + tty
         + (container != null ? "&container=" + container : "");

```

In general I think maybe with the selection of three booleans for different streams, setting up Attach deserves a builder. If there is interest and agreement, I can make a PR.

rawler:
Resolves #847.

The basic idea is to express options for Attach and Exec like:

```
attach
  .prepare(namespace, name)
  .setStdout(true)
  .setTty(true)
  .connect();
```

### DIFF
--- a/util/src/main/java/io/kubernetes/client/Attach.java
+++ b/util/src/main/java/io/kubernetes/client/Attach.java
@@ -57,18 +57,14 @@ public class Attach {
     this.apiClient = apiClient;
   }
 
-  private String makePath(
-      String namespace, String name, String container, boolean stdin, boolean tty) {
-    return "/api/v1/namespaces/"
-        + namespace
-        + "/pods/"
-        + name
-        + "/attach?"
-        + "stdin="
-        + stdin
-        + "&tty="
-        + tty
-        + (container != null ? "&container=" + container : "");
+  /**
+   * Setup a Builder for the given namespace and name
+   *
+   * @param namespace The namespace of the Pod
+   * @param name The name of the Pod
+   */
+  public ConnectionBuilder newConnectionBuilder(String namespace, String name) {
+    return new ConnectionBuilder(namespace, name);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


util/src/main/java/io/kubernetes/client/Attach.java
```diff
@@ -57,18 +57,14 @@ public void setApiClient(ApiClient apiClient) {
...
-        + "&tty="
-        + tty
-        + (container != null ? "&container=" + container : "");
+  /**
+   * Setup a Builder for the given namespace and name
+   *
+   * @param namespace The namespace of the Pod
+   * @param name The name of the Pod
+   */
+  public Builder prepare(String namespace, String name) {
```
brendandburns: Perhaps `newBuilder()` or `getBuilder()` or just `builder()` instead of `prepare()`?

util/src/main/java/io/kubernetes/client/Exec.java
```diff
@@ -79,36 +79,15 @@ public void setApiClient(ApiClient apiClient) {
...
-            + StringUtils.join(command, "&command=");
-    return path;
+  /**
+   * Setup a Builder for the given namespace, name and command
+   *
+   * @param namespace The namespace of the Pod
+   * @param name The name of the Pod
+   * @param command The command to run
+   */
+  public Builder prepare(String namespace, String name, String[] command) {
```
brendandburns: same comment wrt naming

brendandburns:
Thanks for the PR!

This looks reasonable, I had a couple of small comments. 

However unit tests are failing, you can run them with `mvn test` in the `util` directory. Let me know if you run into any problems.

util/src/main/java/io/kubernetes/client/Attach.java
rawler: I spent a long time pondering the naming here. It's not really a "builder" per-se, it more of a configured connector of sorts. But I guess `builder()` is good enough.

rawler:
There. I'm now happy with this and ready for review and merge.

util/src/main/java/io/kubernetes/client/Attach.java
yue9944882: `Attach` is more of a client class so the name `newBuilder()` sounds like it's meant to build another deep copy of the client, which is not the case. i'd vote on `attach.newConnectionBuilder()...build()`.

brendandburns: or maybe even make it a static on the `Builder` class itself? (does that work in Java?) so it would be:

`attach.Builder.newBuilder()...build()`

brendandburns:
Let's get the naming for the connection builder right and then we can merge.

Thanks for the patience and prompt changes!

util/src/main/java/io/kubernetes/client/Attach.java
rawler: I don't think you can use "new" for a method name (reserved keyword). I've changed the code to "newConnectionBuilder()" vs. "newExecutionBuilder".

brendandburns:
@rawler names look good now, but CI/CD is failing.

Once that is corrected we can merge this.

brendandburns:
This looks good to me. @yue9944882 what do you think?

brendandburns:
/approve
/lgtm

Thanks for the patience!

PR #863, opened 2020-02-17
Fix javadoc for picky javascript detector.

brendandburns:


brendandburns:
self lgtm to get the release working.

Issue #855, opened 2020-01-30
Support for OpenID Connect


mlbiam:
I want to use the java api for a kubectl plugin but it looks like openid connect clients aren't supported:

```
[main] ERROR io.kubernetes.client.util.KubeConfig - Unknown auth provider: oidc
----- Scale Deployment Start -----
[main] WARN com.tremolosecurity.App - Exception had occured 
io.kubernetes.client.openapi.ApiException: Unauthorized
        at io.kubernetes.client.openapi.ApiClient.handleResponse(ApiClient.java:979)
        at io.kubernetes.client.openapi.ApiClient.execute(ApiClient.java:895)
        at io.kubernetes.client.openapi.apis.CoreV1Api.listNamespaceWithHttpInfo(CoreV1Api.java:16069)
        at io.kubernetes.client.openapi.apis.CoreV1Api.listNamespace(CoreV1Api.java:16041)
        at com.tremolosecurity.App.getAllNameSpaces(App.java:91)
        at com.tremolosecurity.App.main(App.java:53)
```

I'd like to write an oidc implementation.  Which branch should I write it against?  master doesn't compile.

brendandburns:
Master should compile. What errors are you seeing?

mlbiam:
environment:
```
➜  java git:(master) java -version
java version "1.8.0_181"
Java(TM) SE Runtime Environment (build 1.8.0_181-b13)
Java HotSpot(TM) 64-Bit Server VM (build 25.181-b13, mixed mode)
➜  java git:(master) mvn -version
executing mvnw instead of mvn
Apache Maven 3.5.0 (ff8f5e7444045639af65f6095c62210b5713f426; 2017-04-03T15:39:06-04:00)
Maven home: /Users/mlb/.m2/wrapper/dists/apache-maven-3.5.0-bin/6ps54u5pnnbbpr6ds9rppcc7iv/apache-maven-3.5.0
Java version: 1.8.0_181, vendor: Oracle Corporation
Java home: /Library/Java/JavaVirtualMachines/jdk1.8.0_181.jdk/Contents/Home/jre
Default locale: en_US, platform encoding: UTF-8
OS name: "mac os x", version: "10.15.1", arch: "x86_64", family: "mac"
```

build error:
```
0%: io.kubernetes.client.openapi.models.V1ClusterRoleBindingListFluentImpl
Generating: io.kubernetes.client.openapi.models.V1ClusterRoleBindingListFluentImplFluent
[...]

PR #871, opened 2020-02-23
Add openid connect support

mlbiam:
Added oidc support for #855 tested with both commercially signed certs and with self signed certs


util/src/main/java/io/kubernetes/client/util/authenticators/OpenIDConnectAuthenticator.java
```diff
@@ -0,0 +1,246 @@
+/*
+Copyright 2017 The Kubernetes Authors.
```
brendandburns: year is now 2020

util/src/main/java/io/kubernetes/client/util/authenticators/OpenIDConnectAuthenticator.java
```diff
@@ -0,0 +1,246 @@
...
+  public boolean isExpired(Map<String, Object> config) {
+    String idToken = (String) config.get(OIDC_ID_TOKEN);
+
+    if (idToken == null) {
+      return true;
+    } else {
+      JsonWebSignature jws = new JsonWebSignature();
+      try {
+        jws.setCompactSerialization(idToken);
+        // we don't care if its valid or not cryptographicly
```
brendandburns: why not?

util/src/main/java/io/kubernetes/client/util/authenticators/OpenIDConnectAuthenticator.java
```diff
@@ -0,0 +1,246 @@
...
+    String idpCert = (String) config.get(OIDC_IDP_CERT_DATA);
+
+    SSLContext sslContext = null;
+
+    if (idpCert != null) {
+      // fist, lets get the pem
+      String pemCert = new String(Base64.getDecoder().decode(idpCert));
+
+      // next lets get a cert object
+      String alias = "doenotmatter";
```
brendandburns: Can you explain this a little more in the comment? What does this string do?

brendandburns: Also, is it used anywhere?

util/src/main/java/io/kubernetes/client/util/authenticators/OpenIDConnectAuthenticator.java
```diff
@@ -0,0 +1,246 @@
...
+      KeyStore ks;
+
+      try {
+        ks = java.security.KeyStore.getInstance("PKCS12");
+        ks.load(null, "doesnotmatter".toCharArray());
+        ByteArrayInputStream bais = new ByteArrayInputStream(pemCert.getBytes("UTF-8"));
+        CertificateFactory cf = CertificateFactory.getInstance("X.509");
+        Collection<? extends java.security.cert.Certificate> c = cf.generateCertificates(bais);
+
+        if (c.size() > 1) {
```
brendandburns: Is there a reason for this if condition? Why not use a loop for all cases?

util/src/main/java/io/kubernetes/client/util/authenticators/OpenIDConnectAuthenticator.java
```diff
@@ -0,0 +1,246 @@
...
+            ks.setCertificateEntry(alias + "-" + j, certificate);
+          }
+        } else {
+          ks.setCertificateEntry(alias, c.iterator().next());
+        }
+
+        TrustManagerFactory tmf = TrustManagerFactory.getInstance("PKIX");
+        tmf.init(ks);
+
+        sslContext = SSLContext.getInstance("TLSv1.2");
```
brendandburns: Perhaps make `TLSv1.2` customizable? Can be a TODO comment.

util/src/main/java/io/kubernetes/client/util/authenticators/OpenIDConnectAuthenticator.java
```diff
@@ -0,0 +1,246 @@
...
+      } catch (KeyStoreException
+          | NoSuchAlgorithmException
+          | CertificateException
+          | IOException
+          | KeyManagementException e) {
+        throw new RuntimeException("Could not import idp certificate", e);
+      }
+    }
+
+    if (clientSecret == null) {
```
brendandburns: Use `Map.getOrDefault(...)` above instead of this if condition.

util/src/main/java/io/kubernetes/client/util/authenticators/OpenIDConnectAuthenticator.java
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    StringBuilder sb = new StringBuilder();
+    sb.append(issuer);
+    if (!issuer.endsWith("/")) {
+      sb.append("/");
+    }
+    sb.append(".well-known/openid-configuration");
+
+    String wellKnownURL = sb.toString();
```
brendandburns: delete this and just use sb.toString() below in the URL constructor.

util/src/main/java/io/kubernetes/client/util/authenticators/OpenIDConnectAuthenticator.java
```diff
@@ -0,0 +1,246 @@
...
+    try {
+      URL wellKnown = new URL(wellKnownURL);
+      HttpsURLConnection https = (HttpsURLConnection) wellKnown.openConnection();
+      https.setRequestMethod("GET");
+      if (sslContext != null) {
+        https.setSSLSocketFactory(sslContext.getSocketFactory());
+      }
+      https.setUseCaches(false);
+      int code = https.getResponseCode();
+      if (code == 200) {
```
brendandburns: Use the constants in `HttpURLConnection`

brendandburns: Rather than 

```java
if (code == 200) {
   //long block of code
} else {
   // throw
}
```

It is preferable to do:

```java
if (code != 200) {
   // throw exception
}
// long block of code here
```

That keeps the flow of the code easier to read because it reduces indentation.

util/src/main/java/io/kubernetes/client/util/authenticators/OpenIDConnectAuthenticator.java
```diff
@@ -0,0 +1,246 @@
...
+                .append(URLEncoder.encode(refreshToken, "UTF-8"))
+                .toString();
+        OutputStream ou = https.getOutputStream();
+        ou.write(urlData.getBytes("UTF-8"));
+        ou.flush();
+        ou.close();
+
+        code = https.getResponseCode();
+
+        if (code == 200) {
```
brendandburns: as above wrt to `if (code == 200)` vs `if (code != 200)`

util/src/main/java/io/kubernetes/client/util/authenticators/OpenIDConnectAuthenticator.java
```diff
@@ -0,0 +1,246 @@
...
+          scanner = new Scanner(https.getInputStream(), StandardCharsets.UTF_8.name());
```
brendandburns: a bunch of this code is duplicated from above, and this method is pretty long anyway, refactor the HTTP object loading logic out into a helper method.

brendandburns:
Thanks for the PR, I added a bunch of comments.

This also needs a unit test before it can merge.

kobi26c:
@mlbiam 
Hi, 
I've tried to use this class (OpenIDConnectAuthenticator) and it didn't work for me.
Got an exception of: Invalid response code for token retrieval  
When I looked at the message it returned:
```
 {
    "error_description": "grant type not supported",
    "error": "unsupported_grant_type"
}
```
After changing (in refresh method) the creation of url data to this: 
```
                String urlData =
                        new StringBuilder()
                                .append("grant_type=refresh_token&")
                                .append("client_id=").append(clientId).append("&")
                                .append("client_secret=").append(clientSecret).append("&")
                                .append("refresh_token=")
                                .append(URLEncoder.encode(refreshToken, "UTF-8"))
                                .toString();
```
(And removing the Authorization request property) 
We are using google accounts issuer. 
Hope it can help someone..

mlbiam:
@kobi26c Good catch.  The fix is to use `Basic` instead of `Bearer` in the header and include the `grant_type` in the request.  Per https://tools.ietf.org/html/rfc6749#section-2.3 including the `client_id` and `client_secret` in the body is discouraged.

mlbiam:
@brendandburns happy to make the updates.  Is there a style guide that can be linked to by https://github.com/kubernetes-client/java/blob/master/CONTRIBUTING.md?  When I checked before implementing most of your comments weren't covered.

util/src/main/java/io/kubernetes/client/util/authenticators/OpenIDConnectAuthenticator.java
mlbiam: updated comment.  since the cert used to verify the token and the token are both coming over the same channel a malicious proxy can alter both without the client having any way to know.  verifying the token serves no security benefit.

brendandburns:
@mlbiam there's no formal style guide, though we do use `mvn fmt:fmt` to enforce some basic style, but most of those comments are more about code structure than style, so it's hard to put that in a guide.

I think this is ready to merge, but the CI/CD is failing. Once that is green, I think we're good to merge.

Thanks again for the PR.

util/src/main/java/io/kubernetes/client/util/authenticators/OpenIDConnectAuthenticator.java
```diff
@@ -0,0 +1,288 @@
...
+    return (String) config.get(OIDC_ID_TOKEN);
+  }
+
+  @Override
+  public boolean isExpired(Map<String, Object> config) {
+    String idToken = (String) config.get(OIDC_ID_TOKEN);
+
+    if (idToken == null) {
+      return true;
+    } else {
```
brendandburns: Sorry I missed this. Please drop the `else` it's unnecessary given the `return` in the if clause.

brendandburns:
/lgtm
/approve

I had one comment, but it can be addressed in a follow-up PR.

Thanks for the PR!

PR #882, opened 2020-03-11
Release 8.0

brendandburns:


examples/pom.xml
```diff
@@ -8,7 +8,7 @@
 	<parent>
 		<groupId>io.kubernetes</groupId>
 		<artifactId>client-java-parent</artifactId>
-		<version>8.0.0-SNAPSHOT</version>
```
yue9944882: we should also trim the `SNAPSHOT` suffix to indicate it's a release version

yue9944882:
> [ERROR]   KubernetesInformerCreatorTest.testInformerInjection:148 No requests exactly matched. Most similar request was:  expected:<
GET
/api/v1/pods?watch=true
watch: true
> but was:<
GET
/api/v1/pods?watch=true
watch: true
>

interesting.. CI seems failing on something identical to the expectation..

PR #898, opened 2020-03-26
Adding example for spring-based controller construction

yue9944882:
/cc @brendandburns


brendandburns:
Looks good to me, but needs `mvn fmt:fmt` to fit project style guide.

PR #900, opened 2020-03-27
Feat: Implementing resizable/error exec stream

yue9944882:
Kubelet is actually exposing a set of channels in the exec webconnection:

0. stdin
1. stdout
2. stderr
3. kubelet connection error
4. resizing 

this pull adds access for the last two channel

/cc @brendandburns

### DIFF
--- a/util/src/main/java/io/kubernetes/client/Exec.java
+++ b/util/src/main/java/io/kubernetes/client/Exec.java
@@ -349,7 +349,7 @@ public class Exec {
     return -1;
   }
 
-  protected static class ExecProcess extends Process {
+  public static class ExecProcess extends Process {
     private final WebSocketStreamHandler streamHandler;
     private int statusCode = -1;
     private boolean isAlive = true;
@@ -423,6 +423,14 @@ public class Exec {
     @Override
     public InputStream getErrorStream() {
       return getInputStream(2);
+    }
+
+    public InputStream getConnectionErrorStream() {
+      return getInputStream(3);
+    }
+
+    public OutputStream getResizeStream() {
+      return streamHandler.getOutputStream(4);
     }
 
     private synchronized InputStream getInputStream(int stream) {


util/src/main/java/io/kubernetes/client/Exec.java
```diff
@@ -349,7 +349,16 @@ static int parseExitCode(ApiClient client, InputStream inputStream) {
     return -1;
   }
 
-  protected static class ExecProcess extends Process {
+  public interface ResizableStream {
+    OutputStream getResizeStream();
```
brendandburns: Add documentation as to what this is so that people can implement the interface.

util/src/main/java/io/kubernetes/client/Exec.java
```diff
@@ -349,7 +349,16 @@ static int parseExitCode(ApiClient client, InputStream inputStream) {
...
+  }
+
+  public interface ConnectionErrorStream {
+    InputStream getConnectionErrorStream();
```
brendandburns: Here too.

util/src/main/java/io/kubernetes/client/Exec.java
```diff
@@ -349,7 +349,16 @@ static int parseExitCode(ApiClient client, InputStream inputStream) {
     return -1;
   }
 
-  protected static class ExecProcess extends Process {
+  public interface ResizableStream {
```
brendandburns: What is the purpose of adding these interfaces? Why not just add the methods to `ExecProcess`?

util/src/main/java/io/kubernetes/client/Exec.java
yue9944882: added documents

util/src/main/java/io/kubernetes/client/Exec.java
yue9944882: added here too

util/src/main/java/io/kubernetes/client/Exec.java
yue9944882: > What is the purpose of adding these interfaces?

the standard interface `java.lang.Process` only exposes access to three streams: stdin/stdout/stderr. while for a complete implementation of "web-terminal", we will have to leave the developers public access to another stream named "resize" which tells the kubernetes exec stream to probably adjust the terminals width&height.. 

> Why not just add the methods to ExecProcess?

it  is protected. otherwise how do you think about promote it to a public inner class?

brendandburns: I guess my feeling is you're going to have to cast it one way or the other:

```java
Process p = Exec.exec(...);
ResizableExecStream res = (ResizableExecStream)p;
res.getResizeStream();
...
```

vs

```java
ExecProcess p = (ExecProcess) Exec.exec(...);
p.getResizeStream();
```

I don't see the win from the interfaces, but I could be wrong.

yue9944882: i reverted the interfaces and made the class `ExecProcess` public w/ the new getters, wdyt?

myoss:
@yue9944882 hello, how to usage this ? resize web terminal

```java
ExecProcess p = (ExecProcess) Exec.exec(...);
p.getResizeStream();
```

PR #901, opened 2020-03-27
Require Kubernetes API type for event-filter annotation

yue9944882:
/cc @brendandburns 

this pull requires developers to set apiType explicitly for event filters.


spring/src/main/java/io/kubernetes/client/spring/extended/controller/KubernetesReconcilerProcessor.java
```diff
@@ -85,13 +85,22 @@ private Controller buildController(SharedInformerFactory sharedInformerFactory,
         Function<?, Request> workQueueKeyFunc = watch.workQueueKeyFunc().newInstance();
         for (Method method : r.getClass().getMethods()) {
           if (method.isAnnotationPresent(AddWatchEventFilter.class)) {
-            addFilter = new AddFilterAdaptor(r, method);
+            AddWatchEventFilter annotation = method.getAnnotation(AddWatchEventFilter.class);
+            if (watch.apiTypeClass().equals(annotation.apiTypeClass())) {
+              addFilter = new AddFilterAdaptor(r, method);
+            }
```
brendandburns: This will silently drop non `apiType` types, right? I feel like that might confuse users, perhaps throw an exception? Or log an error?

yue9944882: added warning logging if any dangling filter annotation were found

spring/src/main/java/io/kubernetes/client/spring/extended/controller/KubernetesReconcilerProcessor.java
```diff
@@ -13,14 +13,14 @@
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.time.Duration;
-import java.util.ArrayList;
-import java.util.List;
+import java.util.*;
```
brendandburns, line 16: No wildcard imports, please.

brendandburns:
One minor style comment.

spring/src/main/java/io/kubernetes/client/spring/extended/controller/KubernetesReconcilerProcessor.java
yue9944882: thx, great catch

PR #905, opened 2020-03-31
Create job from cron job

gobozov:
Creating job from cronjob. Based on kubectl source code https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/create/create_job.go
Original issue https://github.com/kubernetes/kubernetes/issues/47538


gobozov:
I signed it

kubernetes/src/main/java/io/kubernetes/client/openapi/ModelUtil.java
```diff
@@ -0,0 +1,70 @@
+package io.kubernetes.client.openapi;
```
brendandburns, line 1: This should be in the `util/src/main/java/...` directory, this directory is all generated code.

kubernetes/src/main/java/io/kubernetes/client/openapi/apis/BatchV1Api.java
```diff
@@ -13,26 +13,21 @@
 
 package io.kubernetes.client.openapi.apis;
 
+import com.google.gson.reflect.TypeToken;
```
brendandburns, line 16: This code is all generated, we don't edit it in this project.

brendandburns:
Most of the code in this library is code generated. We can't modify it because it will prevent future code generation runs.

If you want to add this, you need to add it as a separate class in the `util` directory, similar to how `Exec` and `Attach` etc were added.

kubernetes/src/main/java/io/kubernetes/client/openapi/ModelUtil.java
yue9944882: would rename the package path to `io.kubernetes.client.util.conversion.Jobs`

gobozov: ok, I didn't figure out project structure. I can move into util

kubernetes/src/main/java/io/kubernetes/client/openapi/apis/BatchV1Api.java
gobozov: ok, where all api methods gets defined?

gobozov: Just trying to understand... You mean we can not change api methods for separate project aka adding BatchV1Api.createNamespacedJobFromCronJob, we can add util method for conversion, call it by passing cronJob object and pass result to BatchV1Api.createNamespacedJob to run job. Correct?

brendandburns: Yes, that's correct. Basically, we don't edit things in the `kubernetes` directory (mostly, there is a `custom` dir in there...)

gobozov: Examples self generated as well? Wonder if I can add example class to show how this work.

brendandburns:
CI/CD looks like it got disabled. We need to switch to github actions anyway, I need to hold off merging this until we get CI/CD fixed.

brendandburns:
close/reopen to trigger CI/CD

yue9944882:
> [ERROR] Non complying file: /home/runner/work/java/java/util/src/main/java/io/kubernetes/client/util/conversion/Jobs.java
> [ERROR] Failed to execute goal com.coveo:fmt-maven-plugin:2.5.0:check (default) on project client-java: Found 1 non-complying files, failing build -> [Help 1]

@gobozov the CI job is failing on format-check, plz run the plugin to fix it

gobozov:
Fixed formatting.

gobozov:
How to get this merged? I see that approved label is required.

brendandburns:
/approve

Sorry for the delay

tozka:
Hi, 

How do I see in which k8s release is this change going to be part of ?

Issue #908, opened 2020-04-05
copy.copyFileFromPod() always throws java.io.IOException: Pipe broken


bhardwahnitish19:
Hi,

I am using 'io.kubernetes.client-java:8.0.0' with java 11.0.6 and k8s v1.15. I am facing a weird issue in copying some files from pod. I am using minikube on ubuntu and container has a base image of linux OS as well.

     try{
        InputStream dataStream = copy.copyFileFromPod("default", "myPod", "/etc/server/tls-cert.pem");
        final byte[] bytes = IOUtils.toByteArray(dataStream);
        String content = new String(bytes, StandardCharsets.UTF_8);
         log.info("TLS cert of CA:: {}", content);
       } catch (Exception e){
                   log.error("could not read tlsFile", e);
          }

Despite of the file being present at that location, this always throws: 

    java.io.IOException: Pipe broken
	at java.base/java.io.PipedInputStream.read(PipedInputStream.java:321) ~[na:na]
	at java.base/java.io.PipedInputStream.read(PipedInputStream.java:377) ~[na:na]
	at java.base/java.io.InputStream.read(InputStream.java:205) ~[na:na]
	at org.apache.commons.codec.binary.BaseNCodecInputStream.read(BaseNCodecInputStream.java:159) ~[commons-codec-1.13.jar:1.13]
	at java.base/java.io.FilterInputStream.read(FilterInputStream.java:107) ~[na:na]
	at org.apache.commons.compress.utils.IOUtils.copy(IOUtils.java:85) ~[commons-compress-1.19.jar:1.19]
	at org.apache.commons.compress.utils.IOUtils.copy(IOUtils.java:60) ~[commons-compress-1.19.jar:1.19]
	at org.apache.commons.compress.utils.IOUtils.toByteArray(IOUtils.java:226) ~[commons-compress-1.19.jar:1.19]
	at com.hlf.admin.console.apiserver.controller.FabricCAController.addKubeConfig(FabricCAController.java:71) ~[main/:na]
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[na:na]
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[na:na]
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[na:na]
[...]

Issue #861, opened 2020-02-11
Copy.copyFileFromPod() does not close stream


ChangSurrey:
API version: 7.0.0
I'm using the following code to copy a file from a specified pod's `sourcePath` to `destinationPath`.

```
Copy copy = new Copy();
try (InputStream is = copy.copyFileFromPod(namespace, podName, containerName, sourcePath);
FileOutputStream fos = new FileOutputStream(destinationPath)) {
            is.transferTo(fos);
            fos.flush();
}
catch (...)
```

While the file gets successfully copied over to `destinationPath` quickly (within 1-2s), and I've verified that the copied file's content is intact, the stream is never closed, which causes `is.transferTo(fos)` to never return. 

I've tried using `ByteStreams.copy(is, fos)` and it's the same outcome. 

As a result, after around 20-30s, I get an `IOException` that says `Pipe broken`: 
```
java.io.IOException: Pipe broken

	at java.base/java.io.PipedInputStream.read(PipedInputStream.java:321)
	at java.base/java.io.PipedInputStream.read(PipedInputStream.java:377)
	at java.base/java.io.InputStream.read(InputStream.java:205)
	at org.apache.commons.codec.binary.BaseNCodecInputStream.read(BaseNCodecInputStream.java:158)
	at java.base/java.io.InputStream.transferTo(InputStream.java:704)
(rest of stack trace)
```

In `Copy.java` the source code of the method I'm calling is: 
```
  public InputStream copyFileFromPod(String namespace, String pod, String container, String srcPath)
      throws ApiException, IOException {
    Process proc =
        this.exec(
            namespace,
            pod,
            new String[] {"sh", "-c", "cat " + srcPath + " | base64"},
            container,
            false,
            false);
    return new Base64InputStream(proc.getInputStream());
  }
```

This might be a bug?

ChangSurrey:
I have reproduced the issue using the example at https://github.com/kubernetes-client/java/blob/master/examples/src/main/java/io/kubernetes/client/examples/CopyExample.java

ChangSurrey:
I can also reproduce it using the `Exec` example at https://github.com/kubernetes-client/java/blob/master/examples/src/main/java/io/kubernetes/client/examples/ExecExample.java, which is thrown at line 93: `ByteStreams.copy(proc.getInputStream(), System.out);`

```
java.io.IOException: Pipe broken
	at java.base/java.io.PipedInputStream.read(PipedInputStream.java:321)
[...]

PR #920, opened 2020-04-17
Fix for non closure of input stream for Copy utility

sagarbhavsar4328:
Fix for Issues #861 and #908
Closing the exec process (or WebSocket) post receiving the exit message (i.e. when stream == 3).

### DIFF
--- a/examples/src/main/java/io/kubernetes/client/examples/ExecExample.java
+++ b/examples/src/main/java/io/kubernetes/client/examples/ExecExample.java
@@ -98,6 +98,19 @@ public class ExecExample {
             });
     out.start();
 
+    Thread error =
+        new Thread(
+            new Runnable() {
+              public void run() {
+                try {
+                  ByteStreams.copy(proc.getErrorStream(), System.err);
+                } catch (IOException ex) {
+                  ex.printStackTrace();
+                }
+              }
+            });
+    error.start();
+
     proc.waitFor();
 
     // wait for any last output; no need to wait for input thread
--- a/util/src/main/java/io/kubernetes/client/Copy.java
+++ b/util/src/main/java/io/kubernetes/client/Copy.java
@@ -150,14 +150,6 @@ public class Copy extends Exec {
         }
       }
     }
-    try {
-      int status = proc.waitFor();
-      if (status != 0) {
-        throw new IOException("Copy command failed with status " + status);
-      }
-    } catch (InterruptedException ex) {
-      throw new IOException(ex);
-    }
   }
 
   public static void copyFileFromPod(String namespace, String pod, String srcPath, Path dest)
--- a/util/src/main/java/io/kubernetes/client/Exec.java
+++ b/util/src/main/java/io/kubernetes/client/Exec.java
@@ -371,6 +371,8 @@ public class Exec {
                   }
                 }
                 inStream.close();
+                // The websocket needs to be closed.
+                this.close();
               } else super.handleMessage(stream, inStream);
             }
 
--- a/util/src/test/java/io/kubernetes/client/ExecTest.java
+++ b/util/src/test/java/io/kubernetes/client/ExecTest.java
@@ -109,13 +109,14 @@ public class ExecTest {
 
     process.getHandler().bytesMessage(makeStream(1, msgData.getBytes(StandardCharsets.UTF_8)));
     process.getHandler().bytesMessage(makeStream(2, errData.getBytes(StandardCharsets.UTF_8)));
-    process.getHandler().bytesMessage(makeStream(3, OUTPUT_EXIT0.getBytes(StandardCharsets.UTF_8)));
 
     final ByteArrayOutputStream stdout = new ByteArrayOutputStream();
     final ByteArrayOutputStream stderr = new ByteArrayOutputStream();
 
     Thread t1 = asyncCopy(process.getInputStream(), stdout);
     Thread t2 = asyncCopy(process.getErrorStream(), stderr);
+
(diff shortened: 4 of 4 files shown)


util/src/main/java/io/kubernetes/client/Copy.java
```diff
@@ -150,14 +150,15 @@ public void copyDirectoryFromPod(
         }
       }
     }
-    try {
+    // No need to call waitFor() as proc is closed on exit(stream == 3).
+    /*try {
```
brendandburns, line 154: Please just delete code instead of commenting it out.

brendandburns:
One minor comment, otherwise LGTM.

util/src/main/java/io/kubernetes/client/Copy.java
sagarbhavsar4328: review comments are incorporated.

util/src/main/java/io/kubernetes/client/Exec.java
```diff
@@ -371,6 +371,8 @@ protected void handleMessage(int stream, InputStream inStream) throws IOExceptio
                   }
                 }
                 inStream.close();
+                // The websocket needs to be closed.
```
yue9944882, line 374:
```suggestion
                // Stream ID of `3` delivers exec errors from kubelet and it means the exec is somehow working abnormally, the websocket connection shall be closed by itself upon these errors.
```

examples/src/main/java/io/kubernetes/client/examples/ExecExample.java
```diff
@@ -98,6 +98,19 @@ public void run() {
             });
     out.start();
 
+    Thread error =
```
yue9944882, line 101: this looks irrelevant w/ the purpose of this pull? but the change is good

util/src/main/java/io/kubernetes/client/Copy.java
```diff
@@ -150,14 +150,6 @@ public void copyDirectoryFromPod(
         }
       }
     }
-    try {
-      int status = proc.waitFor();
```
yue9944882, line 154: what's the point of deleting this? the method is always supposed to be hanging until the copy completes, isn't it?

util/src/main/java/io/kubernetes/client/Exec.java
```diff
@@ -371,6 +371,8 @@ protected void handleMessage(int stream, InputStream inStream) throws IOExceptio
...
+                this.close();
```
yue9944882, line 375: and we should call `ExecProcess.this.notifyAll();` in the end, are we?

util/src/main/java/io/kubernetes/client/Copy.java
sagarbhavsar4328: When the response from k8s API (exit call) is lightning fast, `ExecProcess.this.notifyAll();` gets called before `proc.waitFor();` and this method gets hanged for an infinite duration as no one is left to call `notify()` or `notifyAll()` method on this `proc` object.

util/src/main/java/io/kubernetes/client/Exec.java
sagarbhavsar4328: yes, we are but that will just release the locks and resume the blocked threads (if any) but it won't close the WebSocket.
I believe the WebSocket connection is not getting close by itself nor the input stream.
And eventually, the code gets hanged at `ByteStreams.copy(dataStream, System.out);` in `CopyExample`.

examples/src/main/java/io/kubernetes/client/examples/ExecExample.java
sagarbhavsar4328: I will move this to a different PR.

util/src/main/java/io/kubernetes/client/Exec.java
yue9944882: > yes, we are but that will just release the locks and resume the blocked threads (if any)

and it will improve the robustness. how about releasing a CountDownLatch here?

> I believe the WebSocket connection is not getting close by itself nor the input stream.
And eventually, the code gets hanged at ByteStreams.copy(dataStream, System.out); in CopyExample.

if i'm understanding correctly, as the websocket is a full-duplex communication, there's supposed to be two "close" action for completely shutting down the connection:
- one is at the server-side, triggerred by the k8s apiserver.
- the other is at the client-side, by invoking `socket.close()` somewhere else

so the websocket isn't "closed" as long as either side isn't yet closed. after the k8s apiserver signals an exit code by the pipe id-3 (reaching this line), it should do the server-side close, which on the other hand triggers the client-side close. so theoretically we don't need to close it again upon exit codes. but as the close call looks idempotent, closing it twice wouldn't be risky..

sagarbhavsar4328: > after the k8s apiserver signals an exit code by the pipe id-3 (reaching this line), it should do the server-side close,

Unfortunately, at this point, k8s apiserver is NOT closing the connection from there end but on the other side, if the client itself initiates the WebSocket close, post that k8s apiserver also invokes `close()` method of `WebSocketStreamHandler`.

util/src/main/java/io/kubernetes/client/Copy.java
sagarbhavsar4328: > m how do you think about replacing the `wait/notify` w/ `CountDownLatch`? 

`CountDownLatch` seems to be a good idea, let me try it out.

brendandburns:
Closing this in favor of #933

PR #950, opened 2020-05-13
Re-generated source/models for kubernetes 1.17 branch

yue9944882:


yue9944882:
hold until https://github.com/kubernetes-client/gen/pull/157

PR #956, opened 2020-05-20
Bring spring integration pom inline with other poms.

### LINKED ISSUE #934: client-java-spring-integration maven artifacts
I don't seem to be able to locate the maven artifacts for Spring Integration. Is this an oversight, or is it that its not yet supported?

brendandburns:
Fixes #934 (I think)

cc @yue9944882


yue9944882:
https://mvnrepository.com/artifact/io.kubernetes/client-java-spring-integration

the spring module is released days ago. so #934 is already fixed, is it? but the change looks good i think

PR #967, opened 2020-06-02
reload service account token after expiry

### LINKED ISSUE #652: client should periodically reload InClusterConfig token
See https://github.com/kubernetes/kubernetes/issues/70679

zshihang:
fixes #652. by default, this will set an expiration period of 1 minutes on service account token. as we are graduating the Beta feature Projected Service Account Volume to GA, we expect clients to reload token before it expires.


zshihang:
/assign @brendandburns 

we can use the same approach here to fix #290.

util/src/main/java/io/kubernetes/client/util/credentials/TokenFileAuthentication.java
```diff
@@ -0,0 +1,59 @@
...
+import java.nio.charset.Charset;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.time.Instant;
+import okhttp3.Interceptor;
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.Response;
+
+public class TokenFileAuthentication implements Authentication, Interceptor {
```
yue9944882, line 14: i presume using `Interceptor` to hijack the header is a bit overkilling.. the previous approach by `AccessTokenAuthentication` is setting tokens by `((ApiKeyAuth) auth).setApiKey(apiKey);` in the `provide` call, wdyt if we do the token-reloading similarly?

util/src/main/java/io/kubernetes/client/util/credentials/TokenFileAuthentication.java
```diff
@@ -0,0 +1,59 @@
...
+    this.expiry = expiry;
+  }
+
+  public void setFile(String file) {
+    this.file = file;
+  }
+
+  @Override
+  public void provide(ApiClient client) {
+    OkHttpClient withInterceptor = client.getHttpClient().newBuilder().addInterceptor(this).build();
```
yue9944882, line 48: the # of interceptors seems to be growing unboundly upon every `provide` call, isn't it?

zshihang: yes, the assumption is made that the `provide` function is called only once for each in-cluster client. i am not sure when a user will call this twice on the same client instance.

util/src/main/java/io/kubernetes/client/util/credentials/TokenFileAuthentication.java
zshihang: the current openapi doesn't have reloading logic in `ApiKeyAuth` which will only be set once in this java client.

we need something similar to this change https://github.com/OpenAPITools/openapi-generator/pull/6036 (it might take much longer time) to implement reloading.

util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
```diff
@@ -209,11 +208,8 @@ public static ClientBuilder cluster() throws IOException {
     final String port = System.getenv(ENV_SERVICE_PORT);
     builder.setBasePath(host, port);
 
-    final String token =
-        new String(
-            Files.readAllBytes(Paths.get(SERVICEACCOUNT_TOKEN_PATH)), Charset.defaultCharset());
     builder.setCertificateAuthority(Files.readAllBytes(Paths.get(SERVICEACCOUNT_CA_PATH)));
-    builder.setAuthentication(new AccessTokenAuthentication(token));
+    builder.setAuthentication(new TokenFileAuthentication(SERVICEACCOUNT_TOKEN_PATH));
```
yue9944882, line 212: we need to keep backward compatibility and users are expected to opt-in to the change. wdyt add a constructor w/ parameter `reloadInterval` and factor the original `cluster()` to adopt an infinite (or zero) interval?

util/src/main/java/io/kubernetes/client/util/credentials/TokenFileAuthentication.java
yue9944882: >  the assumption is made that the provide function is called only once for each in-cluster client.

yea, that's true

util/src/main/java/io/kubernetes/client/util/credentials/TokenFileAuthentication.java
yue9944882: > the current openapi doesn't have reloading logic in ApiKeyAuth which will only be set once in the initialization of in-cluster client.

convinced, can you leave a TODO comment to reference OpenAPITools/openapi-generator#6036 so that we can converge in the future?

util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
zshihang: `cluster()` with reloading should be the default so that users dont have to do anything to migrate and add a `oldCluster` as the fail safe. 

update the InClusterClusterExample to something like:

```
// new style
ApiClient client = ClientBuilder.cluster().build();

// old style
ApiClient client = ClientBuilder.oldCluster().build();
```

WDYT?

yue9944882: > cluster() with reloading should be the default so that users dont have to do anything to migrate and add a oldCluster as the fail safe.

yea we should also reflect how to opt-in SA reloading in the example. note that there's two aspects of change the pull is involving:

1. introducing capability for periodical SA reloading
2. moving default behavior the in-cluster client to 1-min reloading (preserving the old-style method or not)

the first is good while i'm leaning on doing the second in another release. i will defer to @brendandburns for thoughts

brendandburns: Let's just do the whole thing. We haven't cut a `9.0.0` release yet so I think it's ok to make this change in the `9.0.0` release.

util/src/main/java/io/kubernetes/client/util/credentials/TokenFileAuthentication.java
zshihang: yes, done

zshihang:
thanks for the review. need /approval label to merge.

PR #974, opened 2020-06-10
Adding a check to test if Tar utility is present in the target container

sarveshkaushal:
Right now copy method throws runtime error if Tar is not present in the target container. Added a check for it before proceeding with copying process.
Note: This will only work for the containers which has POSIX based underlying image

### DIFF
--- a/util/src/main/java/io/kubernetes/client/Copy.java
+++ b/util/src/main/java/io/kubernetes/client/Copy.java
@@ -115,48 +115,51 @@ public class Copy extends Exec {
   public void copyDirectoryFromPod(
       String namespace, String pod, String container, String srcPath, Path destination)
       throws ApiException, IOException {
-    // TODO: Test that 'tar' is present in the container?
-    final Process proc =
-        this.exec(
-            namespace,
-            pod,
-            new String[] {"sh", "-c", "tar cz - " + srcPath + " | base64"},
-            container,
-            false,
-            false);
-    try (InputStream is = new Base64InputStream(new BufferedInputStream(proc.getInputStream()));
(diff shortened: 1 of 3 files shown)


util/src/main/java/io/kubernetes/client/Copy.java
```diff
@@ -115,48 +115,51 @@ public void copyDirectoryFromPod(String namespace, String pod, String srcPath, P
...
-        if (!archive.canReadEntryData(entry)) {
-          log.error("Can't read: " + entry);
-          continue;
-        }
-        File f = new File(destination.toFile(), entry.getName());
-        if (entry.isDirectory()) {
-          if (!f.isDirectory() && !f.mkdirs()) {
-            throw new IOException("create directory failed: " + f);
+    // Test that 'tar' is present in the container?
+    if (isTarPresentInContainer(namespace, pod, container)) {
```
yue9944882, line 119:
```suggestion
    if (!isTarPresentInContainer(namespace, pod, container)) {
        // do something
    }
```

nit: plz reverse the test for a cleaner indent?

sarveshkaushal: I was originally thinking of exception but wasn't sure of the name. I have created an exception CopyNotSupportedException in io.kubernetes.client.util.exception package. Hope that is right place for it. I have also flipped the if check. Also added the CopyNotSupportedException in respective calling methods.

brendandburns:
Thanks for the PR

Ultimately, we may want to use [ephemeral containers](https://kubernetes.io/docs/concepts/workloads/pods/ephemeral-containers/) when they become more mature.

/lgtm
/approve

PR #1012, opened 2020-06-27
Add relative path loading for certificates.

### LINKED ISSUE #1009: KubeConfig should read certs files relative to config file
The kube client detects the location from the KUBECONFIG/default directory, but it resolves the files relative to the running application. This behavior differs from `kubectl`s behavior.

````txt
Exception in thread "main" java.nio.file.NoSuchFileException: certs\kube.internal.pem
	at java.base/sun.nio.fs.WindowsException.translateToIOException(WindowsException.java:85)
	at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:103)
	at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:108)
	at java.base/sun.nio.fs.WindowsFileSystemProvider.newByteChannel(WindowsFileSystemProvider.java:235)
	at java.base/java.nio.file.Files.newByteChannel(Files.java:370)
	at java.base/java.nio.file.Files.newByteChannel(Files.java:421)
	at java.base/java.nio.file.Files.readAllBytes(Files.java:3205)
	at io.kubernetes.client.util.KubeConfig.getDataOrFile(KubeConfig.java:403)
	at io.kubernetes.client.util.ClientBuilder.kubeconfig(ClientBuilder.java:254)
	at io.kubernetes.client.util.ClientBuilder.getClientBuilder(ClientBuilder.java:114)
	at io.kubernetes.client.util.ClientBuilder.standard(ClientBuilder.java:90)
	at io.kubernetes.client.util.ClientBuilder.standard(ClientBuilder.java:85)
	at io.kubernetes.client.util.Config.defaultClient(Config.java:113)
	at de.example.test.LeaderElectionTest.main(LeaderElectionTest.java:15)
````

**Expected behavior**

The KubeConfig should load the additional files relative to the config file that specified them.

**Versions**

- kubernetes `client-java-extended` v8.0.2
- Java 11 (64bit)
- Windows 10

### PR DESCRIPTION
**Pull request by brendandburns**
Fixes #1009

### DIFF
--- a/util/src/main/java/io/kubernetes/client/util/KubeConfig.java
+++ b/util/src/main/java/io/kubernetes/client/util/KubeConfig.java
@@ -395,7 +395,15 @@ public class KubeConfig {
     return null;
   }
 
-  public static byte[] getDataOrFile(final String data, final String file) throws IOException {
+  public byte[] getDataOrFileRelative(final String data, final String path) throws IOException {
+    String resolvedPath = path;
+    if (resolvedPath != null && this.file != null) {
+      resolvedPath = this.file.toPath().getParent().resolve(path).normalize().toString();
+    }
+    return getDataOrFile(data, resolvedPath);
+  }
+
+  private static byte[] getDataOrFile(final String data, final String file) throws IOException {
     if (data != null) {
       return Base64.decodeBase64(data);
     }
--- a/pom.xml
+++ b/pom.xml
@@ -45,7 +45,7 @@
     <apache.commons.lang3.version>3.10</apache.commons.lang3.version>
     <apache.commons.collections4.version>4.4</apache.commons.collections4.version>
     <apache.commons.compress>1.19</apache.commons.compress>
-    <common.codec.version>1.11</common.codec.version>
+    <common.codec.version>1.14</common.codec.version>
     <spring.boot.version>2.3.1.RELEASE</spring.boot.version>
     <spring.version>5.1.2.RELEASE</spring.version>
   </properties>
--- a/util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
+++ b/util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
@@ -265,7 +265,7 @@ public class ClientBuilder {
     }
 
     final byte[] caBytes =
-        KubeConfig.getDataOrFile(
+        config.getDataOrFileRelative(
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

File: util/src/main/java/io/kubernetes/client/util/KubeConfig.java [hunk 1]
```diff
@@ -395,7 +395,15 @@ private static String getData(Map<String, Object> obj, String key) {
...
-  public static byte[] getDataOrFile(final String data, final String file) throws IOException {
+  public byte[] getDataOrFileRelative(final String data, final String path) throws IOException {
+    String resolvedPath = path;
+    if (resolvedPath != null && this.file != null) {
+      resolvedPath = this.file.toPath().getParent().resolve(path).normalize().toString();
+    }
+    return getDataOrFile(data, resolvedPath);
+  }
+
+  private static byte[] getDataOrFile(final String data, final String file) throws IOException {
```
yue9944882, line 406: how about keeping this `public static`for compatibility?

util/src/main/java/io/kubernetes/client/util/KubeConfig.java
```diff
@@ -395,7 +395,15 @@ private static String getData(Map<String, Object> obj, String key) {
     return null;
   }
 
-  public static byte[] getDataOrFile(final String data, final String file) throws IOException {
+  public byte[] getDataOrFileRelative(final String data, final String path) throws IOException {
```
yue9944882, line 398: if this is an alternative for the old `getDataOrFile`, wdyt we also make this `public static`?

util/src/main/java/io/kubernetes/client/util/KubeConfig.java
brendandburns: I don't think anyone should really be using this function. If you feel strongly I'll switch it back to public, but I think people should really use the new function.

util/src/main/java/io/kubernetes/client/util/KubeConfig.java
brendandburns: It can't be static because it relies on the file location that the KubeConfig was loaded from.  I suppose we could make it static and pass in the file, but that seems unnecessarily complex.

brendandburns:
Fixed formatting, replied to comments, please take another look.

PR #1015, opened 2020-06-28
Switch to kubernetes object common interfaces for informers

yue9944882:
this pull refactors the informer framework to adopt newly introduced `KubernetesObject` and `KubernetesListObject` interface


util/src/main/java/io/kubernetes/client/informer/cache/Cache.java
```diff
@@ -23,13 +23,6 @@
   /** keyFunc defines how to map objects into indices */
   private Function<ApiType, String> keyFunc;
 
-  /**
-   * DEPRECATE: use Caches#NAMESPACE_INDEX instead. TODO: remove after 7.0.0
```
yue9944882, line 27: these functions/variables are removed as been deprecated for releases

PR #1037, opened 2020-07-03
Chore: Github action testing against multiple JDK versions

yue9944882:


brendandburns:
I think we can condense this into a build matrix:

https://docs.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow#configuring-a-build-matrix

yue9944882:
@brendandburns thanks, looks much better now

PR #1045, opened 2020-07-07
Prune "skip" configuration for deploying proto module

yue9944882:
@brendandburns is this module skipped intentionally?


brendandburns:
Yeah, it needs a major update, it was generated a looong time ago.

PR #1051, opened 2020-07-09
Removing duplicate code

sarveshkaushal:
Found another code fragment which can be extracted as method.


sarveshkaushal:
Not sure why it is failing. It's building fine in my local.
/assign @yue9944882

brendandburns:
I think the test might have been flaky. I retriggered the test.

/lgtm
/approve

PR #1063, opened 2020-07-13
Removing unused imports and adding a log parameter for placeholder

sarveshkaushal:
I have added 2 changes:
1. Removed unused imports
2. Added missing log parameter for placeholder

### DIFF
--- a/util/src/main/java/io/kubernetes/client/informer/impl/DefaultSharedIndexInformer.java
+++ b/util/src/main/java/io/kubernetes/client/informer/impl/DefaultSharedIndexInformer.java
@@ -125,6 +125,7 @@ public class DefaultSharedIndexInformer<
           log.warn(
               "DefaultSharedIndexInformer#resyncPeriod {} is smaller than resyncCheckPeriod {} and the informer has already started. Changing it to {}",
               resyncPeriodMillis,
+              resyncCheckPeriodMillis,
               resyncCheckPeriodMillis);
           resyncPeriodMillis = resyncCheckPeriodMillis;
         } else {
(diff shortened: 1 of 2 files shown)


util/src/main/java/io/kubernetes/client/informer/impl/DefaultSharedIndexInformer.java
```diff
@@ -125,6 +125,7 @@ public void addEventHandlerWithResyncPeriod(
           log.warn(
               "DefaultSharedIndexInformer#resyncPeriod {} is smaller than resyncCheckPeriod {} and the informer has already started. Changing it to {}",
               resyncPeriodMillis,
+              resyncCheckPeriodMillis,
```
brendandburns, line 128: why is this change needed?

sarveshkaushal: If you look at the log message- _"DefaultSharedIndexInformer#resyncPeriod {} is smaller than resyncCheckPeriod {} and the informer has already started. Changing it to {}"_. It has 3 placeholders but we are supplying only 2 variables to log.

sarveshkaushal: I can remove this if this looks redundant.

brendandburns: No that's fine, I was just confused about how adding a parameter was necessary if it was compiling before, but now I realize that it's a varargs function.

PR #1067, opened 2020-07-15
Code clean up.

sarveshkaushal:
Summary of changes:
1. Removing redundant return statement
2. Replacing numeric check with Math.max
3. Ending long variable with 'L' for better readability


yue9944882:
thanks! @sarveshkaushal 

/lgtm

PR #1071, opened 2020-07-15
Removing the deprecated method of Google Gson

sarveshkaushal:


yue9944882:
/hold

okhttp 3 and 4 is not fully compatible

PR #1078, opened 2020-07-16
Kubectl label equivalence

yue9944882:
this pull initiates a new set of helper classes under extended module providing similar capabilities w/ kubectl commands:

```
    V1Pod labelledPod =
        Kubectl.label(apiClient, V1Pod.class)
            .apiGroup("")
            .apiVersion("v1")
            .resourceNamePlural("pods")
            .namespace("default")
            .name("foo")
            .addLabel("k1", "v1")
            .addLabel("k2", "v2")
            .execute();
```

### DIFF
--- a/extended/src/main/java/io/kubernetes/client/extended/kubectl/Kubectl.java
+++ b/extended/src/main/java/io/kubernetes/client/extended/kubectl/Kubectl.java
@@ -1,0 +1,66 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package io.kubernetes.client.extended.kubectl;
+
+import io.kubernetes.client.common.KubernetesObject;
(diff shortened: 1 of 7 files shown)


brendandburns:
cc @sarveshkaushal 

Code looks fine to me.

Do we want this in `extended` this has seemed like a pretty common request from folks, maybe just in the `util` package?

yue9944882:
> Do we want this in extended this has seemed like a pretty common request from folks, maybe just in the util package?

yes, we can keep the implementation in the `util` package and the new helper `Kubectl` added by this pull under `extended` is merely a more user-friendly wrapper for those who are already familiar w/ kubectl. wydt?

sarveshkaushal:
@yue9944882 @brendandburns - Can we also create a package in utils where we keep helper utilities? I see a lot of classes directly under io.kubernetes.client.util package. Is this intentional?

yue9944882:
> @yue9944882 @brendandburns - Can we also create a package in utils where we keep helper utilities? I see a lot of classes directly under io.kubernetes.client.util package. Is this intentional?

yea ideally. but that prevents us from referencing other extended utilities b/c of dependency inversion. e.g. in this pull we're referencing `GenericKubernetesApi` which is under `extended` to manipulate dynamic resources.

brendandburns:
@yue9944882 I like the idea of having static utility functions in the `util` package and then a `Kubectl` in extended.

Alternately, we could move the GenericAPI into the `util` package.

yue9944882:
> @yue9944882 I like the idea of having static utility functions in the util package and then a Kubectl in extended.

+1 on having utilities under `util`. however the labels utility this pull adding doesn't actually have a dedicated implementation. it is simply a wrapper for generic api. am i missing anything else?

> Alternately, we could move the GenericAPI into the util package.

sgtm, see #1087

yue9944882:
i rebased onto latest master

Issue #1075, opened 2020-07-15
Update PatchExample to not use Deployment from extensions/v1beta1


php-coder:
The `PatchExample` work with Deployment from deprecated `extensions/v1beta1` API group that [was removed in Kubernetes 1.16](https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/):

https://github.com/kubernetes-client/java/blob/bd8c85771008c266673e2163e4bebd7f7b29f646/examples/src/main/java/io/kubernetes/client/examples/PatchExample.java#L48-L51

This (and possibly others) example should be updated to use Deployment from `apps/v1` instead.

brendandburns:
Would be happy to take a PR to fix this.

sarveshkaushal:
Do we want to change the `ExtensionV1beta1Deployment` to `V1Deployment` by using `AppsV1Api`?

PR #1084, opened 2020-07-17
Switiching to apps/v1 deployment api in the examples

yue9944882:
Fixes: #1075

### DIFF
--- a/examples/src/main/java/io/kubernetes/client/examples/PatchExample.java
+++ b/examples/src/main/java/io/kubernetes/client/examples/PatchExample.java
@@ -15,8 +15,8 @@ package io.kubernetes.client.examples;
 import io.kubernetes.client.custom.V1Patch;
 import io.kubernetes.client.openapi.ApiException;
 import io.kubernetes.client.openapi.Configuration;
-import io.kubernetes.client.openapi.apis.ExtensionsV1beta1Api;
-import io.kubernetes.client.openapi.models.ExtensionsV1beta1Deployment;
+import io.kubernetes.client.openapi.apis.AppsV1Api;
+import io.kubernetes.client.openapi.models.V1Deployment;
 import io.kubernetes.client.util.ClientBuilder;
 import io.kubernetes.client.util.PatchUtils;
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


examples/src/main/java/io/kubernetes/client/examples/PatchExample.java
```diff
@@ -102,17 +101,17 @@ public static void main(String[] args) throws IOException {
       // apply-yaml a deployment, server side apply is alpha in kubernetes v1.14,
       // You need to actively enable the Server Side Apply alpha feature
       // https://kubernetes.io/docs/reference/using-api/api-concepts/#server-side-apply
```
php-coder, line 103: Is this comment is still needed given in the lite of 1.18?

examples/src/main/java/io/kubernetes/client/examples/PatchExample.java
```diff
@@ -102,17 +101,17 @@ public static void main(String[] args) throws IOException {
...
-      ExtensionsV1beta1Deployment deploy4 =
+      V1Deployment deploy4 =
           PatchUtils.patch(
```
innulic, line 105: @yue9944882 I run your version of example and for yaml apply I received an error:

{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"failed to create typed patch object: .spec.template.spec.containers[name=\"hello-node\"].ports: element 0: associative list with keys has an element that omits key field \"protocol\"","code":500}

examples/src/main/java/io/kubernetes/client/examples/PatchExample.java
yue9944882: good catch, i updated the comment to clarify that server-side-apply is enabled by default after v1.16

examples/src/main/java/io/kubernetes/client/examples/PatchExample.java
yue9944882: can you retry on my latest update? i just re'produced the error on a v1.18 cluster and fixed it.

innulic: thank you, now it works

brendandburns:
/lgtm

Thanks!

PR #1090, opened 2020-07-21
Gracefully shutdown listener pool

dddddai:
Use the shutdown pattern in https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ExecutorService.html

I wonder how to determine the timeout

### DIFF
--- a/util/src/main/java/io/kubernetes/client/informer/cache/SharedProcessor.java
+++ b/util/src/main/java/io/kubernetes/client/informer/cache/SharedProcessor.java
@@ -13,10 +13,12 @@ limitations under the License.
 package io.kubernetes.client.informer.cache;
 
 import io.kubernetes.client.common.KubernetesObject;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import org.apache.commons.collections4.CollectionUtils;
@@ -34,15 +36,22 @@ public class SharedProcessor<ApiType extends KubernetesObject> {
 
   private ExecutorService executorService;
 
+  private final Duration timeout;
+
   public SharedProcessor() {
     this(Executors.newCachedThreadPool());
   }
 
   public SharedProcessor(ExecutorService threadPool) {
+    this(threadPool, Duration.ofMinutes(1));
+  }
+
+  public SharedProcessor(ExecutorService threadPool, Duration timeout) {
     this.listeners = new ArrayList<>();
     this.syncingListeners = new ArrayList<>();
 
     this.executorService = threadPool;
+    this.timeout = timeout;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


dddddai:
Perhaps it could create a new thread to shutdown so that `stop()` would not be blocked by
`awaitTermination(long timeout, TimeUnit unit)`

util/src/main/java/io/kubernetes/client/informer/cache/SharedProcessor.java
```diff
@@ -146,7 +147,17 @@ public void stop() {
...
+    executorService.shutdown();
+    try {
+      // Wait a while for existing tasks to terminate
+      if (!executorService.awaitTermination(1, TimeUnit.MINUTES)) {
+        // Cancel currently executing tasks
+        executorService.shutdownNow();
+      }
+    } catch (InterruptedException e) {
+      executorService.shutdownNow();
+      Thread.currentThread().interrupt();
```
yue9944882, line 160: can you elaborate the point of this line? interrupting the current thread doesnt make sense to me

yue9944882:
> Perhaps it could create a new thread to shutdown so that stop() would not be blocked by
awaitTermination(long timeout, TimeUnit unit)

1 min-ish timeout sounds reasonable to me.. in terms of observability, it's good to have more logging for longer timeout i suppose..

util/src/main/java/io/kubernetes/client/informer/cache/SharedProcessor.java
dddddai: Yes, there is no need to interrupt the thread here. (There is no higher level interrupt handlers so it doesn't need to reraise the InterruptedException)

I have removed it.

dddddai:
> thanks for this, overall looks good. can you add test cases verifies that:
> 
> * the stopping workers (within the timeout) should properly block the processor's `stop` call
> * the slow controller-workers are interrupted upon timeout

I add a new test func, please take a look : )

dddddai:
> /lgtm
> 
> 60s can be too long for running a single test, we can consider a passive clock for robustness in the future.. and plz squash

Thanks for your review, I have squashed the commits.

dddddai:
> /lgtm
> 
> 60s can be too long for running a single test, we can consider a passive clock for robustness in the future.. and plz squash

Thanks for your review, I add a func to give the timeout, so that the test would not be too long.

util/src/main/java/io/kubernetes/client/informer/cache/SharedProcessor.java
```diff
@@ -34,6 +35,9 @@
 
   private ExecutorService executorService;
 
+  private long timeout = 1;
+  private TimeUnit unit = TimeUnit.MINUTES;
```
yue9944882, line 39: private final Duration timeout;

util/src/main/java/io/kubernetes/client/informer/cache/SharedProcessor.java
```diff
@@ -146,7 +150,22 @@ public void stop() {
     } finally {
       lock.writeLock().unlock();
     }
-    // TODO(yue9944882): gracefully shutdown listener pool
-    executorService.shutdownNow();
+    // Disable new tasks from being submitted
+    executorService.shutdown();
+    try {
+      // Wait a while for existing tasks to terminate
+      if (!executorService.awaitTermination(timeout, unit)) {
```
yue9944882, line 157: ...(timeof.OfMillis(),  TimeUnit.MilliSecond)

util/src/main/java/io/kubernetes/client/informer/cache/SharedProcessor.java
```diff
@@ -146,7 +150,22 @@ public void stop() {
...
+        // Cancel currently executing tasks
+        executorService.shutdownNow();
+      }
+    } catch (InterruptedException e) {
+      executorService.shutdownNow();
+    }
+  }
+
+  /** set the time to wait for existing tasks to terminate. */
+  public void setTimeout(long timeout, TimeUnit unit) {
```
yue9944882, line 167: instead of setters, let's make a parameter in the constructor

PR #1096, opened 2020-07-23
Gracefully shutdown listener pool

dddddai:
Use the shutdown pattern in https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ExecutorService.html


dddddai:
@yue9944882 Thanks for your suggestion, I have modified as you addressed

Issue #984, opened 2020-06-20
how to get kubernetes version by client api 


zyq1994:
just like kubectl version

brendandburns:
The URL is:

```
https://<master-ip>:443/version?timeout=32s
```

But as far as I know this client doesn't have code to access it currently. It wouldn't be hard to add, but I don't think it's in the swagger/openapi document that Kubernetes currently publishes.

I'll keep this issue open so that someone can implement it.

rjeberhard:
io.kubernetes.client.openapi.apis.VersionApi

zyq1994:
![image](https://user-images.githubusercontent.com/23446769/86590324-f3513400-bfc1-11ea-9801-1c4d3b695042.png)
thank you

brendandburns:
Closing this since apparently it's possible (and `KubectlVersion` was also added)

PR #1098, opened 2020-07-23
Adding utility to get K8 version

sarveshkaushal:
Related to #984


util/src/main/java/io/kubernetes/client/util/version/VersionUtil.java
```diff
@@ -0,0 +1,59 @@
...
+import io.kubernetes.client.openapi.Configuration;
+import io.kubernetes.client.openapi.apis.VersionApi;
+import io.kubernetes.client.openapi.models.VersionInfo;
+import java.io.IOException;
+import okhttp3.Call;
+import okhttp3.Response;
+
+public class VersionUtil {
+
+  private ApiClient apiClient;
```
yue9944882, line 27: it looks that we dont actually need the field?

util/src/main/java/io/kubernetes/client/util/version/VersionUtil.java
```diff
@@ -0,0 +1,59 @@
...
+  private VersionApi versionApi;
```
yue9944882, line 28: let's make it final

util/src/main/java/io/kubernetes/client/util/version/VersionUtil.java
```diff
@@ -0,0 +1,59 @@
...
+      response = call.execute();
+    } catch (IOException ex) {
+      throw new ApiException("Version request failed");
+    }
+
+    if (!response.isSuccessful()) {
+      throw new ApiException(response.code(), "Version request failed: " + response.code());
+    }
+    Gson gson = new Gson();
+    VersionInfo versionInfo = gson.fromJson(response.body().string(), VersionInfo.class);
```
yue9944882, line 56: `this.versionApi.getApiClient().getJSON().deserialize(...)` instead of instantiating a new gson instance

util/src/main/java/io/kubernetes/client/util/version/VersionUtil.java
sarveshkaushal: I have removed it.

util/src/main/java/io/kubernetes/client/util/version/VersionUtil.java
sarveshkaushal: Done.

util/src/main/java/io/kubernetes/client/util/version/VersionUtil.java
sarveshkaushal: I have changed the code to deserialize using above method.

sarveshkaushal:
Thanks for reviewing @yue9944882 .

util/src/main/java/io/kubernetes/client/util/version/VersionUtil.java
```diff
@@ -0,0 +1,53 @@
...
+import io.kubernetes.client.openapi.ApiClient;
+import io.kubernetes.client.openapi.ApiException;
+import io.kubernetes.client.openapi.Configuration;
+import io.kubernetes.client.openapi.apis.VersionApi;
+import io.kubernetes.client.openapi.models.VersionInfo;
+import java.io.IOException;
+import okhttp3.Call;
+import okhttp3.Response;
+
+public class VersionUtil {
```
brendandburns, line 24: suggest `Version` since `util` is redundant with the package.

util/src/main/java/io/kubernetes/client/util/version/VersionUtil.java
```diff
@@ -0,0 +1,53 @@
...
+  public VersionUtil(ApiClient apiClient) {
+    this.versionApi = new VersionApi(apiClient);
+  }
+
+  public VersionInfo getVersion() throws ApiException, IOException {
+    Call call = versionApi.getCodeCall(null);
+    Response response = null;
+    try {
+      response = call.execute();
+    } catch (IOException ex) {
```
brendandburns, line 41: Given that this method can throw `IOException` I don't think there's any benefit to catching this here and re-throwing a different exception.

Also, by throwing `ApiException` without including the root cause `IOException` it will be difficult to understand what the error actually was.

brendandburns:
Some minor comments.

Thanks for the PR!

util/src/main/java/io/kubernetes/client/util/version/VersionUtil.java
sarveshkaushal: I have changed the name.

util/src/main/java/io/kubernetes/client/util/version/VersionUtil.java
sarveshkaushal: Without caching IOException my test is not working. Is it ok if i throw new APIException(ex). Where ex is IOException?

yue9944882:
@sarveshkaushal thanks! plz squash the commits

PR #1101, opened 2020-07-24
Regenerate for 1.18

brendandburns:
cc @yue9944882


yue9944882:
@brendandburns after we switching the formatter, we need to run `mvn spotless:apply` after code generation. i appended one commit to conform the formatter.

brendandburns:
I used `scripts/update-client.sh` to regenerate, we should update that script to do the necessary work (e.g. run spotless, set flags to do the inheritance right)

I'm happy to do that PR if you can point me at the right settings, we should probably update the README:
https://github.com/kubernetes-client/java#update-the-generated-code

too.

yue9944882:
> I used scripts/update-client.sh to regenerate, we should update that script to do the necessary work (e.g. run spotless, set flags to do the inheritance right)

sure, i will update the script in a follow-up

> I'm happy to do that PR if you can point me at the right settings, we should probably update the README:
https://github.com/kubernetes-client/java#update-the-generated-code

the guide is valid and it looks that we don't need to update it. it's just that we need to get https://github.com/kubernetes-client/gen/pull/167 merged so that the generated models can properly implements the common interfaces otherwise the compiling will fail

sarveshkaushal:
@brendandburns @yue9944882 - Probably not related to above discussion but do you think we should add instructions to run maven spotless in our contributing guidelines? This is something which people (including me) forget before sending a PR.

brendandburns:
@yue9944882 I merged that PR and I will attempt to regenerate the code.

yue9944882:
> @yue9944882 I merged that PR and I will attempt to regenerate the code.

sure plz feel free to drop my commits (and force-push)

yue9944882:
> @brendandburns @yue9944882 - Probably not related to above discussion but do you think we should add instructions to run maven spotless in our contributing guidelines? This is something which people (including me) forget before sending a PR.

opened https://github.com/kubernetes-client/java/pull/1110 to document more about the formatter, PTAL. thanks

brendandburns:
For some reason when I regenerate it has `V1Status` implementing `KubernetesListObject`

Any ideas?

yue9944882:
https://github.com/kubernetes-client/gen/blob/48d96a3a3c6531277e026f6d85ef0c50f087fc54/openapi/preprocess_spec.py#L393-L396

sounds odd.. v1status should be excluded when attaching `x-implement` annotations in the generation script..

yue9944882:
@brendandburns can you retry w/ https://github.com/kubernetes-client/gen/pull/169 merged?

brendandburns:
@yue9944882 Thanks for the patch, it worked. This PR has been regenerated and I think it's now ready to go in.

Issue #1099, opened 2020-07-23
How to get raw response from GenericKubernetesClient when error response is not V1Status


avmohan:
In case of response body that is not deserializable into `V1Status`, `GenericKubernetesClient` throws a `RuntimeException` but there is no way to access the actual response body.
https://github.com/kubernetes-client/java/blob/e8de0e6a30/util/src/main/java/io/kubernetes/client/util/generic/GenericKubernetesApi.java#L713
Any good way to log the raw response from server in this case? 

eg: if an api group is not available on the server, the error response is just:
`404 Not found`

Alternately, wouldn't it be better if deserialization to the proper `DataType` or `V1Status` happened lazily only when `getObject/getStatus` is invoked on the `KubernetesResponse`. That way, client could check `isSuccess` before calling `getObject`?

avmohan:
or maybe wrap the ApiException directly - `throw new RuntimeException(e)` instead of `throw new RuntimeException(e.getCause())` so client could read raw response body if needed.

yue9944882:
> or maybe wrap the ApiException directly - throw new RuntimeException(e) instead of throw new RuntimeException(e.getCause()) so client could read raw response body if needed.

@avmohan this approach can be a prompt fix.. please take a look at https://github.com/kubernetes-client/java/pull/1104. i agree that current way of handling exceptions in generic api is ugly --- we're focusing on generalize the kubernetes api access in the first stage. as far as i'm concerned, `ApiException` is too vague for describing kubernetes api failures and i'm looking for a way to break it down to more detailed exceptions, e.g. `ResourceNotFoundException`, `AlreadyExistException`, `ObjectValidationException` etc ideally

avmohan:
Sorry that wouldn't work cause deserialize itself throws `JsonSyntaxException`. 
Added a comment in #1104

avmohan:
oops nvm, i was on 8.0.2

PR #1104, opened 2020-07-24
Throw the original exception if failed deserializing status

yue9944882:
resolves https://github.com/kubernetes-client/java/issues/1099


util/src/main/java/io/kubernetes/client/util/generic/GenericKubernetesApi.java
```diff
@@ -711,7 +711,7 @@ public GenericKubernetesApi(
       }
       V1Status status = apiClient.getJSON().deserialize(e.getResponseBody(), V1Status.class);
       if (null == status) { // the response body can be something unexpected sometimes..
-        throw new RuntimeException(e.getCause());
+        throw new RuntimeException(e);
```
avmohan, line 714: @yue9944882 i don't think this works in the case when the body cannot be deserialized to V1Status.
the `deserialize` call itself throws a `JsonSyntaxException`

brendandburns:
Looks fine to me, but it's failing formatting.

PR #1105, opened 2020-07-24
Adding utility to get K8 version

sarveshkaushal:
My local workspace got corrupted. So created new PR.

### DIFF
--- a/util/src/main/java/io/kubernetes/client/util/version/Version.java
+++ b/util/src/main/java/io/kubernetes/client/util/version/Version.java
@@ -38,8 +38,8 @@ public class Version {
     Response response = null;
     try {
       response = call.execute();
-    } catch (IOException ex) {
-      throw new ApiException(ex);
+    } catch (IOException e) {
+      throw new ApiException(e);
     }
 
     if (!response.isSuccessful()) {
--- a/util/src/main/java/io/kubernetes/client/util/version/Version.java
+++ b/util/src/main/java/io/kubernetes/client/util/version/Version.java
@@ -1,0 +1,53 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


util/src/main/java/io/kubernetes/client/util/version/Version.java
```diff
@@ -0,0 +1,53 @@
...
+
+  public Version() {
+    this(Configuration.getDefaultApiClient());
+  }
+
+  public Version(ApiClient apiClient) {
+    this.versionApi = new VersionApi(apiClient);
+  }
+
+  public VersionInfo getVersion() throws ApiException, IOException {
```
brendandburns, line 36: Does this actually ever throw IOException?

util/src/test/java/io/kubernetes/client/util/version/VersionTest.java
```diff
@@ -0,0 +1,66 @@
...
+  private static final int PORT = 8091;
+  @Rule public WireMockRule wireMockRule = new WireMockRule(PORT);
+
+  @Before
+  public void setup() throws IOException {
+    client = new ClientBuilder().setBasePath("http://localhost:" + PORT).build();
+  }
+
+  @Test
+  public void testUrl() throws InterruptedException, IOException, ApiException {
```
brendandburns, line 45: Add a test for a non-200 return code.

brendandburns:
Looks good, two more comments.

util/src/main/java/io/kubernetes/client/util/version/Version.java
sarveshkaushal: It actually does but it's not able to reach K8 cluster for some reason. But I will need to handle this.

sarveshkaushal: Sorry i meant, the code to deserialize response will throw IOException.

util/src/test/java/io/kubernetes/client/util/version/VersionTest.java
sarveshkaushal: Done

util/src/main/java/io/kubernetes/client/util/version/Version.java
brendandburns: ah, ok.

Issue #1091, opened 2020-07-21
Could Copy avoid tar dependency?


takeshi-yoshimura:
I am testing `io.kubernetes.client.Copy`, but it requires `tar` is installed in the target container.
I workaround this by adding it to my image, but it would be nice if we removed that requirement.

yue9944882:
i believe we need an upstream support to remove tar requirement, see https://github.com/kubernetes/kubernetes/issues/58512. in terms of docker runtime, it already supports runtime-level copy for a period of time but sadly it's not supported in the unified container-runtime interface.

brendandburns:
Some utility is required (e.g. it will never work with a `scratch` container without runtime support)

I'm open to alternative implementations (e.g. base64, bzip2, etc) if something seems better but `tar` is pretty universal.

yue9944882:
i suppose `cat $file | base64` can be an alternative, but it can take much longer time than the tar approach..

sarveshkaushal:
We keep `tar` utility but if tar fails then may be try to do `cat'??

sarveshkaushal:
Should we check for both cpio and tar?

takeshi-yoshimura:
I understand we need runtime supports for running Copy on all images. But, fall back to cat seems enough for my tested containers. I finally run it with multiple file copies instead of directory copy.

brendandburns:
I'm ok with coding a fallback to `cat` we can use `which tar` to figure out if `tar` is present.

brendandburns:
I wonder if we could also use ephemeral containers:

sarveshkaushal:
> I'm ok with coding a fallback to `cat` we can use `which tar` to figure out if `tar` is present.

I will take a look at this.

brendandburns:
I'm closing this since #1107 merged.

Ilhicas:
I don't follow, we went from, no tar, let's fallback to having two other binaries? which and cat?
I'm using a distroless image, and people with from scratch and binaries won't have those as well.
Is there any impediment preventing the use of "binaryless" containers to make a cp from those containers such as in docker?

PR #1107, opened 2020-07-24
Supporting copy operation when tar is not present

sarveshkaushal:
This is related to #1091 .If `tar` is not present in target container, then the code will fall back to copying individual files using `cat`. There are few TODO comments for me to address in more PRs after this.

### DIFF
--- a/util/src/main/java/io/kubernetes/client/Copy.java
+++ b/util/src/main/java/io/kubernetes/client/Copy.java
@@ -19,14 +19,17 @@ import io.kubernetes.client.openapi.Configuration;
 import io.kubernetes.client.openapi.models.V1Pod;
 import io.kubernetes.client.util.exception.CopyNotSupportedException;
 import java.io.BufferedInputStream;
+import java.io.BufferedReader;
 import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.io.OutputStream;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import org.apache.commons.codec.binary.Base64InputStream;
 import org.apache.commons.codec.binary.Base64OutputStream;
 import org.apache.commons.compress.archivers.ArchiveEntry;
@@ -121,6 +124,35 @@ public class Copy extends Exec {
     if (!isTarPresentInContainer(namespace, pod, container)) {
       throw new CopyNotSupportedException("Tar is not present in the target container");
     }
+    copyDirectoryFromPod(namespace, pod, container, srcPath, destination, true);
+  }
+
+  /**
+   * Copy directory from a pod to local.
+   *
+   * @param namespace
+   * @param pod
+   * @param container
+   * @param srcPath
+   * @param destination
+   * @param enableTarCompressing: false if tar is not present in target container
+   * @throws IOException
+   * @throws ApiException
+   */
+  public void copyDirectoryFromPod(
+      String namespace,
+      String pod,
+      String container,
+      String srcPath,
+      Path destination,
+      boolean enableTarCompressing)
+      throws IOException, ApiException {
+    if (!enableTarCompressing) {
+      TreeNode tree = new TreeNode(true, srcPath, true);
+      createDirectoryTree(tree, namespace, pod, container, srcPath);
+      createDirectoryStructureFromTree(tree, namespace, pod, container, srcPath, destination);
+      return;
+    }
     final Process proc =
         this.exec(
             namespace,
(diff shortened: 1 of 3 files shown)


util/src/main/java/io/kubernetes/client/Copy.java
```diff
@@ -164,6 +172,120 @@ public void copyDirectoryFromPod(
...
+        try (InputStream is = copyFileFromPod(namespace, pod, modifiedSrcPath);
+            OutputStream fs = new FileOutputStream(f)) {
+          ByteStreams.copy(is, fs);
+          fs.flush();
+        }
+      } else {
+        String newSrcPath = genericPathBuilder(srcPath, childNode.name);
+        // TODO: Change this once the method genericPathBuilder is changed to varargs
+        Path newDestination =
+            Paths.get(destination.toString() + "/" + node.name + "/" + childNode.name);
```
brendandburns, line 219: `Paths.get(destination.toString(), node.name, childNode.name);` to ensure it works on Windows.

util/src/main/java/io/kubernetes/client/Copy.java
```diff
@@ -164,6 +172,120 @@ public void copyDirectoryFromPod(
...
+    }
+  }
+
+  /*
+  Generic method to create path.
+  TODO: Change this to varargs
+  */
+  private String genericPathBuilder(String first, String second) {
+    StringBuilder pathBuilder = new StringBuilder(first);
+    pathBuilder.append("/");
```
brendandburns, line 284: Should this use `File.separator`

brendandburns:
This code is semi-complex, I think it needs some unit tests for all of the path stuff.

util/src/main/java/io/kubernetes/client/Copy.java
sarveshkaushal: I have changed it.

util/src/main/java/io/kubernetes/client/Copy.java
sarveshkaushal: Done.

sarveshkaushal:
@brendandburns - I agree.  I don't have Windows machine so it would be difficult for me to test. I need help for testing this on Windows. 
Also in order to test this fully using JUnit, we need to mock the websocket. This requires dedicated changes in testing framework. 
BTW code implementation is as follows:

1. If `tar` is not present in target container, get the list of files and directories in the target container by executing `ls`.
2. Create a directory tree by looping over the response of `ls` command on target container. Model it into `TreeNode` class.
3. Parse resultant `TreeNode` and create directories in the destination folder.
4. Parse resultant `TreeNode` and create files in the destination folder. This uses existing `copyFileFromPod` method from `Copy` class.

Also I am not sure if the current location of `TreeNode` class is correct. I think it can be moved to a separate package.
Comments/ thoughts?

util/src/main/java/io/kubernetes/client/Copy.java
```diff
@@ -116,9 +119,14 @@ public void copyDirectoryFromPod(String namespace, String pod, String srcPath, P
   public void copyDirectoryFromPod(
       String namespace, String pod, String container, String srcPath, Path destination)
```
yue9944882, line 120: > public void copyDirectoryFromPod(String namespace, String pod, String container, String srcPath, Path destination, boolean fallbackToNonTarCopy)

how about adding a new static method w/ argument `boolean fallbackToNonTarCopy` and make the old method reuse the new one? so that users can opt-in this change on their need?

sarveshkaushal: Ok, so do we want to make the new method public??

sarveshkaushal:
> thanks! and i will download the patch and test locally on my machine.

Thank you so much for helping @yue9944882 !!

util/src/main/java/io/kubernetes/client/Copy.java
yue9944882: i think so

sarveshkaushal: Ok- please take a look at my latest commit.

yue9944882: sry for being unclear. i was proposing something like:

```java

public void copyDirectoryFromPod(String namespace, String pod, String container, String srcPath, Path destination) 
        throws ApiException, IOException, CopyNotSupportedException {
    if (!isTarPresentInContainer(namespace, pod, container)) {
      throw new CopyNotSupportedException("Tar is not present in the target container"); // If tar is not present then fallback to slow process of copying
    }
    copyDirectoryFromPod(namespace, pod, container, srcPath, destination, false)
}

public void copyDirectoryFromPod(String namespace, String pod, String container, String srcPath, Path destination, boolean enableTarCompressing) 
        throws ApiException, IOException {
    if (!enableTarCompressing) {
        copyDirectoryFromPod(namespace, pod, container, srcPath, destination, true);
        return;
    }
    // the old implemention
}
```

what do you think?

sarveshkaushal: SGTM. I will make the change.

yue9944882:
@sarveshkaushal thanks, the code looks good, and i will test it locally and get back w/ more feedbacks!

yue9944882:
@sarveshkaushal when i was copying a dir from a pod in my local cluster, my code is:

```java
  copy.copyDirectoryFromPod(namespace, podName, null, "/root", Paths.get("/tmp/etc"), false);
```

and the file structure under `/root` in the pod is:

```
└── root
    ├── a.sh
    └── b.sh
```

after running the code above, i got the following structure in `/tmp/etc/`:

```
├── a.sh
├── b.sh
└── root
```

which suggests something wrong w/ the current implementation, am i missing anything?

sarveshkaushal:
> @sarveshkaushal when i was copying a dir from a pod in my local cluster, my code is:
> 
> ```java
>   copy.copyDirectoryFromPod(namespace, podName, null, "/root", Paths.get("/tmp/etc"), false);
> ```
> 
> and the file structure under `/root` in the pod is:
> 
> ```
> └── root
>     ├── a.sh
>     └── b.sh
> ```
> 
> after running the code above, i got the following structure in `/tmp/etc/`:
> 
> ```
> ├── a.sh
> ├── b.sh
> └── root
> ```
> 
> which suggests something wrong w/ the current implementation, am i missing anything?

No this seems to be a bug in the implementation. I will take a look at this. I guess the handling for `/start/dirpath/` and `/start/dirpath` is different.

sarveshkaushal:
@yue9944882 - I have fixed the issue with directory creation. Can you please test this? Also can you please test on windows?

brendandburns:
Sorry, I will try to get this tested on Windows soon.

brendandburns:
I tested this and it looks good to me.

@yue9944882 any more comments before we merge?

sarveshkaushal:
I have rebased it with latest master.

yue9944882 (commented):
@sarveshkaushal thank you for contributing this! let's get this in!

/lgtm
/approve

charcharan:
Hello, Sorry to post in this closed merge request, I have created pod and am trying to copy the file to pod and I am getting , error. Please refer and suggest any workaround, 
https://github.com/kubernetes-client/java/issues/2044

PR #1110, opened 2020-07-27
Documenting guidance on running formatter 

yue9944882:


sarveshkaushal:
Thank you @yue9944882 . I was also thinking if it's possible to format the files during the build when using `mvn clean install`. User doesn't have to run the formatter explicitly.

yue9944882:
@sarveshkaushal can you elaborate on how that works? not really sure about these maven setups ..

PR #1113, opened 2020-07-27
Add an implementation of kubectl scale.

brendandburns:
cc @yue9944882

### DIFF
--- a/extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlScale.java
+++ b/extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlScale.java
@@ -1,0 +1,109 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package io.kubernetes.client.extended.kubectl;
+
+import io.kubernetes.client.common.KubernetesObject;
+import io.kubernetes.client.custom.V1Patch;
+import io.kubernetes.client.extended.kubectl.exception.KubectlException;
+import io.kubernetes.client.openapi.ApiClient;
+import io.kubernetes.client.openapi.ApiException;
+import io.kubernetes.client.openapi.apis.AppsV1Api;
+import io.kubernetes.client.openapi.models.V1Deployment;
+import io.kubernetes.client.openapi.models.V1ReplicaSet;
+import io.kubernetes.client.util.PatchUtils;
+
+public class KubectlScale<ApiType extends KubernetesObject> implements Kubectl.Executable<ApiType> {
+  private final ApiClient apiClient;
+  private final Class<ApiType> apiTypeClass;
+  private final AppsV1Api api;
+  private String namespace;
+  private String name;
+  private int replicas;
+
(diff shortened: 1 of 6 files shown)


extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlScale.java
```diff
@@ -0,0 +1,109 @@
...
+import io.kubernetes.client.custom.V1Patch;
+import io.kubernetes.client.extended.kubectl.exception.KubectlException;
+import io.kubernetes.client.openapi.ApiClient;
+import io.kubernetes.client.openapi.ApiException;
+import io.kubernetes.client.openapi.apis.AppsV1Api;
+import io.kubernetes.client.openapi.models.V1Deployment;
+import io.kubernetes.client.openapi.models.V1ReplicaSet;
+import io.kubernetes.client.util.PatchUtils;
+
+public class KubectlScale<ApiType extends KubernetesObject> implements Kubectl.Executable<ApiType> {
```
yue9944882, line 25: > class KubectlScale<ApiType extends KubernetesObject> implements Kubectl.Executable<ApiType> {

let's degrade public access to package access to converge entrypoint to `Kubectl`'s static method

brendandburns: ok, but this requires changing the return type of the method in `Kubectl.scale(...)` and `Kubectl.label(...)` to be `Kubectl.Executable` rather than `KubectlScale` or `KubectlLabel` is that ok w/ you?

yue9944882: SGTM!

yue9944882:
and this pull needs rebase @brendandburns

extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlScale.java
brendandburns: Oh, I realized that this doesn't work. The reason is that if you try do do:

`Kubectl.scale(...).name(...).namespace(...).replicas(...).execute()`

You need access to the `KubectlScale` class since the `Kubectl.Executable` only has `execute()`

As it stands if `KubectlScale`/`KubectlVersion`/`KubectlLabel` aren't public they can't be used outside of that package.

I'm open to other ideas, but I think if we want to maintain this builder style pattern, we need to make those classes (or alternate interfaces) public.

Thoughts?

brendandburns:
@yue9944882 

Rebased, but please see my reply above wrt to the `public` vs `protected` interfaces.

extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlScale.java
yue9944882: how about we keep the `KubectlScale/KubectlVersion/KubectlLabel` public but degrade the constructor access to `package` (or `private`)?

extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlScale.java
```diff
@@ -0,0 +1,109 @@
...
+  private final ApiClient apiClient;
+  private final Class<ApiType> apiTypeClass;
+  private final AppsV1Api api;
+  private String namespace;
+  private String name;
+  private int replicas;
+
+  public KubectlScale(ApiClient client, Class<ApiType> apiTypeClass) {
```
yue9944882, line 33:
```suggestion
  KubectlScale(ApiClient client, Class<ApiType> apiTypeClass) {
```

extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlScale.java
brendandburns: Works for me, I will do that.

brendandburns:
Comments addressed, made the classes public with package protected constructors. ptal.

Thanks!

PR #1115, opened 2020-07-28
Provide standard method to cleanup schedulers

nurkiewicz:
Once created, `LeaderElector` will never release thread pools. In some environments this prevents JVM shutdown and causes memory leak.


brendandburns:
/approve

Should we update the examples to use the AutoClose functionality?

Issue #1108, opened 2020-07-24
enqueue_mapped functionality


tony-clarke-amdocs:
The controller runtime has way to register secondary resources and have the reconciler be called via the main resource. See [here](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/handler/enqueue_mapped.go). Without this ability it looks like you have to implement two separate controllers, one for the primary resource and one for the secondary resource. Is there a plan to support this enqueue mapped functionality?

yue9944882:
IIUC, watching multiple resources (primary & secondary) is already supported. you can call `watch` multiple times to spawn multiple watches for single controller.

https://github.com/kubernetes-client/java/blob/95d25f9c8200d863c4d06687f1ac961c6105d58f/examples/src/main/java/io/kubernetes/client/examples/ControllerExample.java#L87

tony-clarke-amdocs:
But this is not the same thing. You can add multiple watches that have the same reconciler but then the problem is that the reconciler cannot know which of the resources it is being called for. What the enqueue_mapped provides is a mapping from the secondary resource to the primary resource so that the reconciler always gets called with the mapped primary resource. The client code looks something like:
```
	err = c.Watch(&source.Kind{Type: &corev1.ConfigMap{}}, &handler.EnqueueRequestsFromMapFunc{
		ToRequests: handler.ToRequestsFunc(func(bindingObject handler.MapObject) []reconcile.Request {
			return getAllReconcileRequestList(bindingObject.Meta.GetName(), bindingObject.Meta.GetNamespace())
		}),
	})

        func getAllReconcileRequestList(configmapName string, namespace string) []reconcile.Request {
	requests := []reconcile.Request{}

	prs := model.GetModel().GetPrimaryReferencingConfigMap(configmapName)
	for _, v := range prs {
		requests = append(requests, reconcile.Request{
			NamespacedName: types.NamespacedName{
				Namespace: namespace,
				Name:      v,
			},
		})
	}
	return requests
}

```

Another thing that you can do with the go controller runtime is requested to be notified about secondary resources owned by the primary resource. Again, when the reconciler is called it is called with the primary resource. The client code looks something like:
```
        subresources := []runtime.Object{
		&corev1.ConfigMap{},
	}

	// watch for all instances of these subresources that are owned by the Primary
	for _, subresource := range subresources {
		err = c.Watch(&source.Kind{Type: subresource}, &handler.EnqueueRequestForOwner{
			IsController: true,
[...]

PR #1118, opened 2020-07-29
Support custom workqueue-keyfunc to directly manipulate work-queue instance

yue9944882:
resolves https://github.com/kubernetes-client/java/issues/1108

in order to support mapping one resource event to multiple controller-requests, we made the following adjustment to the controller framework:

- support returning a `null` request from the work-queue-key-func so that it can optionally be silent/no-op at some time
- (for spring integration) support receiving `WorkQueue` instance in the constructor of the work-queue-key-func so it can now manually push multiple controller-requests to the work-queue upon resource events (for sample, see the tests in this pull)

as for [non-spring construction](https://github.com/kubernetes-client/java/blob/894b1631e83d23a65440060b341faeba58c66760/examples/src/main/java/io/kubernetes/client/examples/ControllerExample.java#L90-L92) of controllers, you can pass a key-func instance which holds work-queue instance and manually pushes events to the queue by its customized logic. however, for the versions before 9.0.0, you will have to push a dummy controller-request to the queue and handle it explicitly in your reconciler b/c `null` return value is not supported..

@brendandburns @tony-clarke-amdocs

### DIFF
--- a/spring/src/test/java/io/kubernetes/client/spring/extended/controller/KubernetesReconcilerCreatorTest.java
+++ b/spring/src/test/java/io/kubernetes/client/spring/extended/controller/KubernetesReconcilerCreatorTest.java
@@ -12,24 +12,41 @@ limitations under the License.
 */
 package io.kubernetes.client.spring.extended.controller;
 
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.fail;
 
 import com.github.tomakehurst.wiremock.junit.WireMockRule;
+import io.kubernetes.client.common.KubernetesObject;
 import io.kubernetes.client.extended.controller.Controller;
+import io.kubernetes.client.extended.controller.DefaultController;
 import io.kubernetes.client.extended.controller.reconciler.Reconciler;
 import io.kubernetes.client.extended.controller.reconciler.Request;
 import io.kubernetes.client.extended.controller.reconciler.Result;
+import io.kubernetes.client.extended.workqueue.WorkQueue;
 import io.kubernetes.client.informer.SharedInformer;
+import io.kubernetes.client.informer.SharedInformerFactory;
+import io.kubernetes.client.informer.cache.DeltaFIFO;
 import io.kubernetes.client.informer.cache.Lister;
+import io.kubernetes.client.informer.impl.DefaultSharedIndexInformer;
+import io.kubernetes.client.openapi.models.V1ConfigMap;
+import io.kubernetes.client.openapi.models.V1ObjectMeta;
 import io.kubernetes.client.openapi.models.V1Pod;
-import io.kubernetes.client.spring.extended.controller.annotation.*;
+import io.kubernetes.client.openapi.models.V1PodList;
 import io.kubernetes.client.spring.extended.controller.annotation.AddWatchEventFilter;
 import io.kubernetes.client.spring.extended.controller.annotation.DeleteWatchEventFilter;
+import io.kubernetes.client.spring.extended.controller.annotation.KubernetesReconciler;
+import io.kubernetes.client.spring.extended.controller.annotation.KubernetesReconcilerReadyFunc;
+import io.kubernetes.client.spring.extended.controller.annotation.KubernetesReconcilerWatch;
+import io.kubernetes.client.spring.extended.controller.annotation.KubernetesReconcilerWatches;
 import io.kubernetes.client.spring.extended.controller.annotation.UpdateWatchEventFilter;
+import java.util.LinkedList;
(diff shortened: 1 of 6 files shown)


spring/src/test/java/io/kubernetes/client/spring/extended/controller/KubernetesReconcilerCreatorTest.java
```diff
@@ -96,10 +127,53 @@ private boolean podInformerCacheReady() {
...
+
+    public CustomWorkQueueKeyFunc(WorkQueue<Request> workQueue) {
+      this.workQueue = workQueue;
+    }
+
+    private final WorkQueue<Request> workQueue;
+
+    @Override
+    public Request apply(KubernetesObject item) {
+      workQueue.add(new Request("foo"));
```
yue9944882, line 147: @tony-clarke-amdocs here is an example of customized logic for pushing requests to the queue

tony-clarke-amdocs: So we just need to implement WorkQueueKeyFunc that maps back to the name of primary resource?

yue9944882: there's two separate cases you mentioned in the issue:

- __mapping 1 secondary resource to 1 primary resource__: e.g. referencing by owner-reference. i think this is already supported and you can simply make that work by implementing `WorkQueueKeyFunc` which for example craft a controller-request by unpacking a pod's owner-reference.

- __mapping 1 secondary resource to multiple primary resources__: this case is what this pull trying to address. for now the `WorkQueueKeyFunc` only supports mapping one resource to one request b/c it is declared to be `Function<KubernetesObject, Request>`. in order to achieve that, this pull allows the `WorkQueueKeyFunc` to return a `null` request to indicate the framework not to push anything to the queue, and instead the `WorkQueueKeyFunc` craft/pushes the requests to the queue directly. does that make sense to you?

tony-clarke-amdocs: Thanks! I didn't realize I could already do use case one with the WorkQueueKeyFunc...I will try that out. I understand what I need to do to support the second scenario...maybe we need to add some javadoc to explain how a user is expected to return null and add their own entries to the queue. What version might this be available in?

util/src/main/java/io/kubernetes/client/informer/impl/DefaultSharedIndexInformer.java
```diff
@@ -216,7 +216,7 @@ public boolean hasSynced() {
    *
    * @param deltas deltas
    */
-  private void handleDeltas(Deque<MutablePair<DeltaFIFO.DeltaType, KubernetesObject>> deltas) {
+  public void handleDeltas(Deque<MutablePair<DeltaFIFO.DeltaType, KubernetesObject>> deltas) {
```
brendandburns, line 219: why did this become public?

spring/src/test/java/io/kubernetes/client/spring/extended/controller/KubernetesReconcilerCreatorTest.java
```diff
@@ -12,24 +12,39 @@
 */
 package io.kubernetes.client.spring.extended.controller;
 
-import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.*;
```
brendandburns, line 15: Can you remove this wildcard import in favor of the specific functions?

util/src/main/java/io/kubernetes/client/informer/impl/DefaultSharedIndexInformer.java
yue9944882: this is for testing, so that we can directly push events to the queue. otherwise we will have to mock out watch api-endpoint which is far more difficult.

spring/src/test/java/io/kubernetes/client/spring/extended/controller/KubernetesReconcilerCreatorTest.java
yue9944882: updated

util/src/main/java/io/kubernetes/client/informer/impl/DefaultSharedIndexInformer.java
brendandburns: Can you add the Guava `@VisibleForTesting` annotation

https://guava.dev/releases/19.0/api/docs/com/google/common/annotations/VisibleForTesting.html

PR #1130, opened 2020-07-31
Little improvements to the update client script

leogtzr:
- errors should go to stderr
- exit with no arguments takes the previous command exit code, so it wasking the exit code from the echo command.
- Using $(...) instead of `` which is way better.


leogtzr:
> LGTM, can you sign the CLA so that the pull can proceed?

Hi, @yue9944882 I just did it :) I can send you a copy if you need it.

PR #1136, opened 2020-08-05
Refactor(kubectl): get rid of force api type class casting

yue9944882:
Get rid of force type casting by adding type parameter to ResourceBuilder.

follow-up of #1135


brendandburns:
Cool! I struggled w/ getting all the generic stuff to line up right, thanks for cleaning this up!

/lgtm
/approve

PR #1137, opened 2020-08-06
Add devcontainer for github codespaces.

brendandburns:
See details here:

https://docs.github.com/en/github/developing-online-with-codespaces/configuring-codespaces-for-your-project


yue9944882:
great! i was actually investigating [katacoda](https://www.katacoda.com/learn) weeks ago, so that we can provide a smooth demo of our example codes on cloud-IDE. will be getting back reviewing this pull after i learn more about codespaces!

yue9944882:
@brendandburns can you rebase so that the tests can pass?

PR #1139, opened 2020-08-06
Kubectl exec follow-up & refactors

yue9944882:


examples/src/main/java/io/kubernetes/client/examples/KubectlExample.java
```diff
@@ -110,11 +110,14 @@ public static void main(String[] args)
         System.exit(0);
       case "exec":
         name = args[1];
-        String[] command = Arrays.copyOfRange(args, 1, args.length);
+        String[] command = Arrays.copyOfRange(args, 2, args.length);
```
yue9944882, line 113: @brendandburns we need to copy from 2 otherwise the pod name will be included

extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlExec.java
```diff
@@ -23,15 +24,14 @@
 import java.io.InputStream;
 import java.io.OutputStream;
 
-public class KubectlExec extends Kubectl.ResourceAndContainerBuilder<V1Pod, KubectlExec>
-    implements Kubectl.Executable<V1Pod> {
+public class KubectlExec extends Kubectl.ResourceAndContainerBuilder<KubernetesObject, KubectlExec>
+    implements Kubectl.Executable<Integer> {
```
yue9944882, line 27: use `Integer` as return value type --- remote process exit code

extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlExec.java
```diff
@@ -24,11 +24,10 @@
 import java.io.OutputStream;
 
 public class KubectlExec extends Kubectl.ResourceAndContainerBuilder<V1Pod, KubectlExec>
-    implements Kubectl.Executable<V1Pod> {
+    implements Kubectl.Executable<Integer> {
```
yue9944882, line 27: use Integer as return value type, so `exitCode()` can be removed --- remote process exit code

PR #1140, opened 2020-08-06
Graceful shutdown sharedprocessor

### LINKED ISSUE #1134: Shutdown of SharedProcessor
I believe the shutdown of the shared processor is designed to be graceful. It will wait one minute for each informer. This means that for each informer it will wait one minute. These waits are not done in parallel but in sequence. This means that if you have say 10 informers it will wait 10 minutes. It doesn't seem correct? Furthermore, it would be nice if there was a way to say "don't wait, just shutdown now" for testing.

yue9944882:
Fixes #1134

### DIFF
--- a/util/src/main/java/io/kubernetes/client/informer/cache/SharedProcessor.java
+++ b/util/src/main/java/io/kubernetes/client/informer/cache/SharedProcessor.java
@@ -23,11 +23,15 @@ import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import org.apache.commons.collections4.CollectionUtils;
 import org.joda.time.DateTime;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /*
  * SharedProcessor class manages all the registered ProcessorListener and distributes notifications.
  */
 public class SharedProcessor<ApiType extends KubernetesObject> {
+
+  private static final Logger log = LoggerFactory.getLogger(SharedProcessor.class);
 
   private ReadWriteLock lock = new ReentrantReadWriteLock();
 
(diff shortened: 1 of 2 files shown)


dddddai:
The following test should be removed, because `stop()` would interrupt running listeners immediately

https://github.com/kubernetes-client/java/blob/master/util/src/test/java/io/kubernetes/client/informer/cache/SharedProcessorTest.java#L70-L83

yue9944882:
> The following test should be removed, because stop() would interrupt running listeners immediately

thanks for reminding, removed the related tests

tony-clarke-amdocs:
Which branch will this fix make its way into?

yue9944882:
> Which branch will this fix make its way into?

it's targeting master

tony-clarke-amdocs:
> > Which branch will this fix make its way into?
> 
> it's targeting master

Will there be a 9.0.1 release with this? If not, what would be the next release that would include this change? Thanks.

yue9944882:
> Will there be a 9.0.1 release with this?

@tony-clarke-amdocs we can get this fix into master then cherry-pick to release branches then publish a 9.0.1 @brendandburns wdyt?

brendandburns:
/lgtm
/approve

Yes, I'm happy to cut a `9.0.1` release.

tony-clarke-amdocs:
Thanks, will pick this up from 9.0.1

brendandburns:
@yue9944882 as long as we're cherry-picking, I want to get #1145 merged and pick that too.

tony-clarke-amdocs:
@brendanburns I don't see this in the 9.0.1 release?

brendandburns:
@tony-clarke-amdocs I see it here in the branch:

https://github.com/kubernetes-client/java/commits/release-9.0.0

Do you mean that the fix isn't in the Maven package? What symptoms are you seeing?

I believe that everything should be there, but it's possible that something went wrong.

brendandburns:
@tony-clarke-amdocs 

Also, if there is a problem that you are seeing, can you file a new issue, it's tricky to track discussions on pull requests that have been closed, it will be easier if there is an open issue.

Thanks!

PR #1142, opened 2020-08-07
Promote apiGroup/apiVersion/resourceName to resourceBuilder

yue9944882:
share more codes in the resource builder

### DIFF
--- a/extended/src/main/java/io/kubernetes/client/extended/kubectl/Kubectl.java
+++ b/extended/src/main/java/io/kubernetes/client/extended/kubectl/Kubectl.java
@@ -177,6 +177,9 @@ public class Kubectl {
     final Class<ApiType> apiTypeClass;
     String namespace;
     String name;
+    String apiGroup;
+    String apiVersion;
+    String resourceNamePlural;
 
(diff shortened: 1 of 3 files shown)


extended/src/main/java/io/kubernetes/client/extended/kubectl/Kubectl.java
```diff
@@ -192,6 +196,25 @@ public T namespace(String namespace) {
...
+      this.apiVersion = apiVersion;
+      return (T) this;
+    }
+
+    public T resourceNamePlural(String resourceNamePlural) {
+      this.resourceNamePlural = resourceNamePlural;
+      return (T) this;
+    }
+
+    public boolean isNamespaced() {
```
brendandburns, line 215: I'm still not sure that this is the right test, since `kubectl` allows "no namespace" to be "current namespace"

brendandburns:
LGTM, but one comment about `isNamespaced()`

extended/src/main/java/io/kubernetes/client/extended/kubectl/Kubectl.java
yue9944882: agree, i moved `isNamespaced` out of resource builder

PR #1147, opened 2020-08-11
kubectl(api-resources): api discovery implementation

yue9944882:
this pull implements client-side api-discovery and `kubectl api-resources` equivalence

### DIFF
--- a/util/src/main/java/io/kubernetes/client/Discovery.java
+++ b/util/src/main/java/io/kubernetes/client/Discovery.java
@@ -28,6 +28,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 import okhttp3.Call;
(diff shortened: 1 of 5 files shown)


util/src/main/java/io/kubernetes/client/Discovery.java
```diff
@@ -0,0 +1,237 @@
...
+  }
+
+  public Set<APIResource> findAll(
+      String group, List<String> versions, String preferredVersion, String path)
+      throws ApiException {
+    V1APIResourceList resourceList = resourceDiscovery(path);
+    return sort(group, versions, preferredVersion, resourceList);
+  }
+
+  private Set<APIResource> sort(
```
brendandburns, line 74: it feels like `sort` isn't quite the right name for this method as it doesn't seem to be sorting anything. Perhaps a more descriptive name?

brendandburns, line 74: This method is complex, I think having a specific unit test for the method would make sense.

util/src/main/java/io/kubernetes/client/Discovery.java
```diff
@@ -0,0 +1,237 @@
...
+      String group, List<String> versions, String preferredVersion, V1APIResourceList resourceList)
+      throws ApiException {
+    Map<String, Set<String>> subResources = new HashMap<>();
+    Set<APIResource> resources =
+        resourceList.getResources().stream()
+            .filter(
+                r -> {
+                  boolean isSubResource = r.getName().contains("/");
```
brendandburns, line 82: extract out as a helper method to improve readability?

brendandburns:
Few comments and a request for more tests.

util/src/main/java/io/kubernetes/client/Discovery.java
yue9944882: renamed to `groupResourcesByName` and added descriptive comments

yue9944882: splitted into multiple methods, PTAL

util/src/main/java/io/kubernetes/client/Discovery.java
yue9944882: done

util/src/main/java/io/kubernetes/client/Discovery.java
```diff
@@ -21,14 +21,7 @@
...
 import io.kubernetes.client.openapi.models.V1APIVersions;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
brendandburns, line 24: remove wildcard imports.

brendandburns:
Looks good, let's remove that wildcard import and then it's good to merge.

util/src/main/java/io/kubernetes/client/Discovery.java
yue9944882: ah oversight

Issue #1129, opened 2020-07-31
Utility code to create configmap from multiple files in a directory?


pravindamodaran:
I'm looking to create configmap from a directory (basically add all files in the directory to a single configmap), same as this:
https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#create-configmaps-from-directories

Is there a way to do it using java api's?

brendandburns:
You will have to do it yourself. Create a `V1ConfigMap` object and then read the files in and add them to the `binaryData` field in the `ConfigMap` resource.

There's no utility to do this currently (though we'd gladly take a PR to add such utility)

sarveshkaushal:
@brendandburns - does this still require work?

PR #1152, opened 2020-08-13
A utility to create Configmaps #1129

sarveshkaushal:
This is related to #1129 . Please take a look. It only implements utility method to read a .properties file and create configmap. If it looks good, then we can enhance it to have similar capabilities as **_kubectl create configmap_**

### DIFF
--- a/util/src/main/java/io/kubernetes/client/Configmap.java
+++ b/util/src/main/java/io/kubernetes/client/Configmap.java
@@ -1,0 +1,203 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package io.kubernetes.client;
+
+import com.google.common.io.CharStreams;
+import io.kubernetes.client.openapi.ApiClient;
+import io.kubernetes.client.openapi.ApiException;
+import io.kubernetes.client.openapi.Configuration;
+import io.kubernetes.client.openapi.apis.CoreV1Api;
+import io.kubernetes.client.openapi.models.V1ConfigMap;
+import io.kubernetes.client.openapi.models.V1ObjectMeta;
+import io.kubernetes.client.util.exception.CreateConfigmapException;
+import java.io.BufferedReader;
+import java.io.FileReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.Reader;
+import java.io.StringReader;
+import java.nio.ByteBuffer;
+import java.nio.charset.CharacterCodingException;
+import java.nio.charset.CharsetDecoder;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Base64;
+import java.util.HashMap;
+import java.util.Map;
+import okhttp3.Call;
+import okhttp3.Response;
+
+/** Utility class to create Configmaps */
+public class Configmap {
+
+  private ApiClient apiClient;
+  private CoreV1Api coreClient;
+
+  public Configmap() {
+    this(Configuration.getDefaultApiClient());
+  }
+
+  public Configmap(ApiClient apiClient) {
+    this.apiClient = apiClient;
+    this.coreClient = new CoreV1Api(apiClient);
+  }
+
+  public ApiClient getApiClient() {
+    return apiClient;
+  }
+
+  public void setApiClient(ApiClient apiClient) {
+    this.apiClient = apiClient;
+  }
+
+  /**
+   * Method to create a Configmap from a file. The file should contain key-value pairs
+   *
+   * @param namespace
+   * @param name
+   * @param file
+   * @throws IOException
+   * @throws CreateConfigmapException
+   * @throws ApiException
+   */
+  public void createFromFile(String namespace, String name, Path file)
+      throws IOException, CreateConfigmapException, ApiException {
+    if (Files.isRegularFile(file)) {
+      try (BufferedReader bufferedReader =
+          new BufferedReader(new FileReader(String.valueOf(file)))) {
+        createFromReader(namespace, name, bufferedReader);
+      }
+    } else {
+      throw new CreateConfigmapException("Target path is a directory");
+    }
+  }
+
+  /**
+   * To create Configmap by passing InputStream of a file which has key value pairs
+   *
+   * @param namespace
+   * @param name
+   * @param stream
+   * @throws IOException
+   * @throws CreateConfigmapException
+   * @throws ApiException
+   */
+  public void createFromStream(String namespace, String name, InputStream stream)
+      throws IOException, CreateConfigmapException, ApiException {
+    createFromReader(namespace, name, new InputStreamReader(stream));
+  }
+
+  /**
+   * To creat Configmap by passing string literal. Value should be a string in the form "key=value"
+   *
+   * @param namespace
+   * @param name
+   * @param value
+   * @throws IOException
+   * @throws CreateConfigmapException
+   * @throws ApiException
+   */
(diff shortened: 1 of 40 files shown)


yue9944882:
i think we should generalize this to all kinds of resources (instead of configmap-only)

util/src/main/java/io/kubernetes/client/Configmap.java
```diff
@@ -0,0 +1,143 @@
...
+   * @return
+   * @throws IOException
+   * @throws ApiException
+   */
+  public String createConfigmapFromFile(String configmapName, Path filePath)
+      throws IOException, ApiException, CreateConfigmapException {
+    return createConfigmapFromFile("default", configmapName, filePath);
+  }
+
+  public String createConfigmapFromFile(String namespace, String configmapName, Path filePath)
```
brendandburns, line 74: I think that we should either make this capable of loading any file, or rename this method `createConfigMapFromPropertiesFile(...)`

My preference would be to make this generic so that it works with any type of file.

brendandburns:
Thanks for the PR. I think it would be good to make this generic so that it can be used to create a config map containing any kind of file, not just a properties file.

Also, as you write it, I would suggest that you do the following:

```java
public class ConfigMap {
   ...

   void createFromStream(String namespace, String name, InputStream stream) {
        createFromReader(namespace, name, new InputStreamReader(stream));
   }

   void createFromFile(String namespace, String name, Path file) {
       createFromReader(namespace, name, new FileReader(file));
   }

   void createFromString(String namespace, String name, String value) {
       createFromReader(namespace, name, new StringReader(value));
   }

   void createFromReader(String namespace, String name, Reader reader) {
        ...
    }
}
```

Thanks!

brendandburns:
@yue9944882 I'm not sure I follow, since ConfigMaps are pretty specific in that they can hold data from other files.

I do think that if we implement `kubectl create ...` we should make it generic to any kind of resource, which is maybe what you meant.

sarveshkaushal:
> Thanks for the PR. I think it would be good to make this generic so that it can be used to create a config map containing any kind of file, not just a properties file.
> 
> Also, as you write it, I would suggest that you do the following:
> 
> ```java
> public class ConfigMap {
>    ...
> 
>    void createFromStream(String namespace, String name, InputStream stream) {
>         createFromReader(namespace, name, new InputStreamReader(stream));
>    }
> 
>    void createFromFile(String namespace, String name, Path file) {
>        createFromReader(namespace, name, new FileReader(file));
>    }
> 
>    void createFromString(String namespace, String name, String value) {
>        createFromReader(namespace, name, new StringReader(value));
>    }
> 
>    void createFromReader(String namespace, String name, Reader reader) {
>         ...
>     }
> }
> ```
> 
> Thanks!

Thanks for reviewing. I will add these methods. BTW- I just kept single method in it to keep my PR small in size. :-) I thought I would implement others in subsequent PRs.

sarveshkaushal:
@brendandburns - I have made the changes as per your comment. Please take a look.

util/src/main/java/io/kubernetes/client/Configmap.java
```diff
@@ -0,0 +1,197 @@
...
+   * @throws IOException
+   * @throws CreateConfigmapException
+   * @throws ApiException
+   */
+  public void createFromFile(String namespace, String name, Path file)
+      throws IOException, CreateConfigmapException, ApiException {
+    if (Files.isRegularFile(file)) {
+      try (BufferedReader bufferedReader =
+          new BufferedReader(
+              new InputStreamReader(Files.newInputStream(file), StandardCharsets.UTF_8.name()))) {
```
brendandburns, line 76: Just use `new FileReader(...)` instead of `InputStreamReader(....)`

util/src/main/java/io/kubernetes/client/Configmap.java
```diff
@@ -0,0 +1,197 @@
...
+   * @param reader
+   * @throws IOException
+   * @throws CreateConfigmapException
+   * @throws ApiException
+   */
+  public void createFromReader(String namespace, String name, Reader reader)
+      throws IOException, CreateConfigmapException, ApiException {
+    if (reader != null) {
+      if (name != null && !name.isEmpty()) {
+        Map<String, String> map = getPropertyMap(reader);
```
brendandburns, line 128: I think we need to extend this to support any kind of file, not just `.properties` files.

sarveshkaushal: Do you mean that we identify the type of file from extension or metadata and then proceed to parse it? There is bindayData field  in V1ConfigMap. Do you think we should just pass the contents of file (as byte array) into binaryData?

sarveshkaushal: When i try to pass the binary array inside binaryData map it creates the configmap. But if you inspect the created configmap, the data is not present in it. Do you think it could be some kind of bug? There is a big chance that I might be missing something here.

brendandburns: The data in the configmap is Base64 encoded. You need to Base64 encode the binary data.

The relevant Golang code is starts here:

https://github.com/kubernetes/kubectl/blob/master/pkg/generate/versioned/configmap.go#L256

util/src/main/java/io/kubernetes/client/Configmap.java
sarveshkaushal: I have changed it.

util/src/main/java/io/kubernetes/client/Configmap.java
sarveshkaushal: Thanks for helping me with this @brendandburns . I looked at the code of Kubectl. It helped me. I have made few changes as per your review comments. Please take a look.

util/src/main/java/io/kubernetes/client/Configmap.java
```diff
@@ -0,0 +1,203 @@
...
+   * @throws CreateConfigmapException
+   * @throws ApiException
+   */
+  public void createFromReader(String namespace, String name, Reader reader)
+      throws IOException, CreateConfigmapException, ApiException {
+    if (reader != null) {
+      if (name != null && !name.isEmpty()) {
+        V1ConfigMap v1ConfigMap = createConfigmapObject(name);
+        byte[] data = CharStreams.toString(reader).getBytes();
+        if (isValidUtf8(data)) {
```
brendandburns, line 133: Does `kubectl` do this? I thought everything was base64 encoded, even if it was already valid utf-8.

brendandburns:
Looking a lot better! Thanks for the patience. This is quite close, just one question about UTF-8 files.

util/src/main/java/io/kubernetes/client/Configmap.java
sarveshkaushal: This was as i understood from the Kubectl code.  BTW i am a noob in Golang.

sarveshkaushal:
I read the code in the link that you shared.

util/src/main/java/io/kubernetes/client/Configmap.java
sarveshkaushal: Line number 262 at https://github.com/kubernetes/kubectl/blob/master/pkg/generate/versioned/configmap.go#L262

sarveshkaushal:
@brendandburns - I am learning a lot from these reviews. Just wanted to thank you.

util/src/main/java/io/kubernetes/client/Configmap.java
```diff
@@ -0,0 +1,203 @@
...
+   * To create Configmap by passing Reader of a file which has key value pairs
+   *
+   * @param namespace
+   * @param name
+   * @param reader
+   * @throws IOException
+   * @throws CreateConfigmapException
+   * @throws ApiException
+   */
+  public void createFromReader(String namespace, String name, Reader reader)
```
yue9944882, line 127: consider rename `createFromFile/createFromString/createFromReader` to `create`. the former function names are in golang-fashion, in java the parameters are overridable

yue9944882:
> I do think that if we implement kubectl create ... we should make it generic to any kind of resource, which is maybe what you meant.

yeah that's exactly what i mean but am fine w/ starting w/ a configmap-specific implementation

util/src/main/java/io/kubernetes/client/Configmap.java
brendandburns: :) I'm definitely too influenced by Go, this approach sgtm

sarveshkaushal:
I corrupted my local workspace again. I will open a fresh one with review comments incorporated.

PR #1159, opened 2020-08-14
add overloading method to copy in memory binary data to pod

ffazzini:
An overloading copy to pod method which would allow us to perform a copy using an in memory src byte array resource.

The existing copy to pod method expects the file to exist physically in the src pod as it uses `FileInputStream` internally,
the second method added in this PR does not.

### DIFF
--- a/util/src/main/java/io/kubernetes/client/Copy.java
+++ b/util/src/main/java/io/kubernetes/client/Copy.java
@@ -19,6 +19,7 @@ import io.kubernetes.client.openapi.Configuration;
 import io.kubernetes.client.openapi.models.V1Pod;
 import io.kubernetes.client.util.exception.CopyNotSupportedException;
 import java.io.BufferedInputStream;
+import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
@@ -179,15 +180,7 @@ public class Copy extends Exec {
       throws ApiException, IOException {
 
     // Run decoding and extracting processes
-    String parentPath = destPath.getParent() != null ? destPath.getParent().toString() : ".";
-    final Process proc =
-        this.exec(
-            namespace,
-            pod,
-            new String[] {"sh", "-c", "base64 -d | tar -xmf - -C " + parentPath},
-            container,
-            true,
-            false);
+    final Process proc = execCopyToPod(namespace, pod, container, destPath);
 
     // Send encoded archive output stream
     File srcFile = new File(srcPath.toUri());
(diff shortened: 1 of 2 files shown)


ffazzini:
I signed it

ffazzini:
I signed it

brendandburns:
@ffazzini thanks for the PR. I think a better refactor might be to have a method that takes an InputStream instead of a File, then users can pass in a ByteArrayInputStream (or any other type of InputStream), it will be more flexible.

(and if you want to add a method that creates the bytearray stream for someone, that could be another addition)

Let me know if there are questions about that.

Thanks!

ffazzini:
@brendanburns thanks for your reply, i am not sure i understood correctly as i did not use `File` in the new method but `byte[]` , can you comment on the code directly maybe so that would clarify me? Btw are you suggesting to combine the 2 copyToPod methods in 1 method having `inputStream` as input param of the method? Thanks.

brendandburns:
Sorry I wasn't clear, what I was suggesting was instead of:
```java
  public void copyFileToPod(
      String namespace,
      String pod,
      String container,
      byte[] src,
      String targetFileName,
      Path destPath)
```

instead
```java
  public void copyFileToPod(
      String namespace,
      String pod,
      String container,
      InputStream src,
      String targetFileName,
      Path destPath)
```

Then if you want to, you can also add:

```java
  public void copyFileToPod(
      String namespace,
      String pod,
      String container,
      byte[] src,
      String targetFileName,
      Path destPath) {
   copyFileToPod(namespace, pod, container, new ByteArrayInputStream(src), targetFileName, destPath);
}
```

ffazzini:
@brendanburns I tested the method below passing in input either a `FileInputStream` and `ByteArrayInputStream` to the `srcInputStream` and both work fine.

```java
public void copyFileToPod(
      String namespace,
      String pod,
      String container,
      InputStream srcInputStream,
      String targetFileName,
      Path destPath)
      throws ApiException, IOException {

    // Run decoding and extracting processes
    final Process proc = execCopyToPod(namespace, pod, container, destPath);

    try (ArchiveOutputStream archiveOutputStream =
        new TarArchiveOutputStream(new Base64OutputStream(proc.getOutputStream(), true, 0, null))) {

      ArchiveEntry tarEntry = new TarArchiveEntry(new File(targetFileName));
      byte[] src = IOUtils.toByteArray(srcInputStream);
      ((TarArchiveEntry) tarEntry).setSize(src.length);
      
      archiveOutputStream.putArchiveEntry(tarEntry);
      ByteStreams.copy(new ByteArrayInputStream(src), archiveOutputStream);
      archiveOutputStream.closeArchiveEntry();
    } finally {
      proc.destroy();
    }
  }
```

I still need to convert the `InputStream` in `byte[]` inside the method though as it needs to set the size to tar entry
`((TarArchiveEntry) tarEntry).setSize(src.length);`
otherwise it will just create an empty file to the dest pod.

What do you think?

I think your suggestion makes sense as using a generic `InputStream` we might not need this method anymore `public void copyFileToPod(String namespace, String pod, String container, Path srcPath, Path destPath)` in the future ?

brendandburns:
Is it true about needing to call setSize? The current code doesn't do that and I believe that it works ok.

What I'm trying to avoid is reading an entire large file into memory. If you already have an array of bytes its one thing, but if you're streaming from one source (eg. disk, network) to the Pod, I don't want you to have to load a gigabyte of data into memory in order to copy it.

ffazzini:
yes i did try with and without setting the size,
I see the size and other metadata
are set in this constructor https://github.com/kubernetes-client/java/blob/master/util/src/main/java/io/kubernetes/client/Copy.java#L198 to the dest file in the current code, may that make the difference?
The current code expects the file to exist in the src pod so the srcFile has all the needed metadata.

Maybe the code in the current PR using the byte[] is preferable at this point?

If so, let me know i think i can improve it renaming `targetFileName` in favour of `destPath`

brendandburns:
Ok, I'm fine to merge this I guess. We can refactor it more in the future if we need to.

/lgtm
/approve

Issue #1163, opened 2020-08-17
Shared Pod Informer error: io.kubernetes.client.informer.cache.ProcessorListener.java- failed invoking ADD event handler: null


prashantcsutar:
Java client version: 6.0.1

Error:
io.kubernetes.client.informer.cache.ProcessorListener.java- failed invoking ADD event handler: null
io.kubernetes.client.informer.cache.ProcessorListener.java- failed invoking UPDATE event handler: null

We are getting this error on one of our system where AKS version is 1.16.7.

Need your help resolving this issue.

Our code:

    public void startPodInformer() {
        try {
            ApiClient client = Config.defaultClient();
            client.getHttpClient().setReadTimeout(0, TimeUnit.SECONDS);
            Configuration.setDefaultApiClient(client);
            CoreV1Api coreV1Api = new CoreV1Api();

            SharedInformerFactory factory = new SharedInformerFactory();

            // Pod informer
            SharedIndexInformer<V1Pod> podInformer =
                factory.sharedIndexInformerFor(
                    (CallGeneratorParams params) -> coreV1Api.listPodForAllNamespacesCall(
                          null,
                          null,
                          null,
                          null,
                          null,
                          null,
                          params.timeoutSeconds,
                          params.watch,
                          null,
                          null),
                    V1Pod.class,
                    V1PodList.class);

            podInformer.addEventHandler(
                new ResourceEventHandler<V1Pod>() {
                    @Override
                    public void onAdd(V1Pod pod) {
                        // Code to handle add event
                    }

                    @Override
                    public void onUpdate(V1Pod oldPod, V1Pod newPod) {
                        // Code to handle update event
                    }

                    @Override
                    public void onDelete(V1Pod pod, boolean deletedFinalStateUnknown) {
                        // Code to handle delete event
                    }
                });

            factory.startAllRegisteredInformers();
        }
        catch (Exception exception)
        {
            logger.error("Kubenetes POD informer not stated due to execetion {}.", exception.getMessage(), exception);
        }
    }

yue9944882:
> io.kubernetes.client.informer.cache.ProcessorListener.java- failed invoking ADD event handler: null
io.kubernetes.client.informer.cache.ProcessorListener.java- failed invoking UPDATE event handler: null

i suppose your error message shows that the implementation of your event handler is throwing a null pointer exception

brendandburns:
Closing since #1164 merged.

PR #1164, opened 2020-08-18
Chore: Better error messages - save the exception tracestack context

yue9944882:


yue9944882:
this pull will help troubleshooting issues like #1163

PR #1169, opened 2020-08-18
generate model classes from CRDs

simingweng:
- create docker image for generating model classes from arbitrary CRDs
- generation for cert-manager CRDs
- `cert-manager/update.sh` can be a template for other CRDs in a separate module

### DIFF
--- a/client-java-contrib/cert-manager/pom.xml
+++ b/client-java-contrib/cert-manager/pom.xml
@@ -1,0 +1,23 @@
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>cert-manager</artifactId>
+    <name>cert-manager-models</name>
+    <url>https://github.com/kubernetes-client/java</url>
+    <parent>
+        <groupId>io.kubernetes</groupId>
+        <artifactId>client-java-parent</artifactId>
+        <version>10.0.0-SNAPSHOT</version>
+        <relativePath>../../pom.xml</relativePath>
+    </parent>
+    <!-- the version tracks the release version of the CRDs in the upstream cert-manager project -->
+    <version>0.16.1-SNAPSHOT</version>
+    <dependencies>
+        <dependency>
+            <groupId>io.kubernetes</groupId>
+            <artifactId>client-java</artifactId>
+            <version>${project.parent.version}</version>
+            <scope>provided</scope>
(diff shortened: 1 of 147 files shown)


client-java-contrib/generate.sh
```diff
@@ -0,0 +1,61 @@
...
+  esac
+done
+
+# create a KinD cluster on the host
+kind create cluster
+
+# install CRDs to the KinD cluster and dump the swagger spec
+kubectl apply -f "${CRD_URL}"
+sleep 5
+kubectl get --raw="/openapi/v2" > /tmp/swagger
```
yue9944882, line 50:
```suggestion
kubectl get --raw="/openapi/v2" > /tmp/swagger

echo "Verifying CRD installation.."
kubectl get crd -o name \
  | while read L
    do
      if [[ $(kubectl get $L -o jsonpath='{.status.conditions[?(@.type=="NonStructuralSchema")].status}') == "True" ]]; then
        echo "$L failed publishing openapi schema because it's attached non-structral-schema condition."
        exit 1
      fi
      if [[ $(kubectl get $L -o jsonpath='{.preserveUnknownFields}') == "true" ]]; then
        echo "$L failed publishing openapi schema because it explicitly disabled unknown fields pruning."
        exit 1
      fi
      echo "$L successfully installed"
    done
```

adding a post crd installation check in case generating empty models.

client-java-contrib/cert-manager/pom.xml
```diff
@@ -0,0 +1,23 @@
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>cert-manager</artifactId>
```
yue9944882, line 4: how about `model-contrib-cert-manager` or `client-java-contrib-cert-manager`?

yue9944882 (commented):
thanks a bunch for starting this! overall LGTM

client-java-contrib/cert-manager/pom.xml
simingweng: yes, totally agree, I guess `client-java-cert-manager-models` is specific and intuitive enough.

client-java-contrib/generate.sh
simingweng: Thanks for contributing this validation logic in shell script, will incorporate it.

yue9944882 (commented):
bunch of thanks! @brendandburns has approval

/lgtm

yue9944882:
@brendandburns politely bump, do you feel like another round of review before it gets in?

brendandburns:
/approve

apologies, I'm mostly out of office this week on vacation :)

PR #1179, opened 2020-08-20
[Code clean-up]Removing unused imports

sarveshkaushal:
Un-used imports.


extended/src/test/java/io/kubernetes/client/extended/kubectl/KubectlLabelTest.java
```diff
@@ -13,7 +13,6 @@
 package io.kubernetes.client.extended.kubectl;
 
 import static com.github.tomakehurst.wiremock.client.WireMock.*;
```
brendandburns, line 15: Can you remove this wildcard import too and replace it with specific imports (do this in all files)

brendandburns:
Thanks for the clean-up. While you are doing this can you remove an wildcard imports (e.g. `foo.*)` and replace with fully qualified imports (`foo.Bar`)

Thanks!

extended/src/test/java/io/kubernetes/client/extended/kubectl/KubectlLabelTest.java
sarveshkaushal: Done.

sarveshkaushal:
I have expanded the wildcard imports.

PR #1184, opened 2020-08-23
Amend Compatibility table to document n-2 version guarantees.

Capstan:
I don't have a link at the moment, but IIRC, kubernetes masters only promise backwards compatibility of two versions of the API client, so running a version older than that may run into potholes of broken behavior as the master no longer supports a deprecated/removed API.

/cc @lavalamp

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,9 +181,9 @@ All APIs and Models' documentation can be found at the [Generated client's docs]
 
 |  client version  | 1.12      | 1.13     | 1.14     |  1.15     |  1.16    |  1.17    |
 |------------------|-----------|----------|----------|-----------|----------|----------|
-|  4.0.0           |  ✓        |  -       |  -       | -         | -        | -        |
-|  5.0.0           |  +        |  ✓       |  -       | -         | -        | -        |
-|  6.0.1           |  +        |  +       |  ✓       | -         | -        | -        |
+|  4.0.0           |  ✓        |  -       |  -       | x         | x        | x        |
+|  5.0.0           |  +        |  ✓       |  -       | -         | x        | x        |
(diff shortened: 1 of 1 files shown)


yue9944882 (commented):
i believe you're referring to [this page](https://kubernetes.io/docs/setup/release/version-skew-policy/#supported-versions) where it says:

> The Kubernetes project maintains release branches for the most recent three minor releases (1.18, 1.17, 1.16).

/lgtm

lavalamp:
The clients should really continue to work for longer than that. It's really feature deprecation that happens.

Our existing go client libraries mix in stuff that doesn't stop being correct (machinery) with stuff that does change over time (the api types), which is unfortunate.

README.md
```diff
@@ -196,6 +196,9 @@ Key:
   Kubernetes cluster, but everything they have in common will work.
 * `-` The Kubernetes cluster has features the java-client library can't use
   (additional API objects, etc).
+* `x` The Kubernetes cluster has no guarantees to support the API client of
```
brendandburns, line 199: Can we change this to "is not tested" rather than "may simply break"

brendandburns:
I'm not necessarily opposed to this, but our APIs have really rarely changed, I'd like the language to be a little weaker so that we don't scare people away.

brendandburns:
@Capstan friendly ping, thanks!

README.md
Capstan: IIUC, it's more than just "not tested". If an API version is deprecated, and two versions later it is removed, then that functionality will simply not work.

Capstan: Tried to keep clarity while softening the statement. PTAL.

brendandburns:
Thanks!

PR #1187, opened 2020-08-24
Equivalent  for kubectl create

yue9944882:


extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlCreate.java
```diff
@@ -0,0 +1,82 @@
...
+import io.kubernetes.client.util.generic.GenericKubernetesApi;
+import io.kubernetes.client.util.generic.KubernetesApiResponse;
+import io.kubernetes.client.util.generic.options.CreateOptions;
+import java.util.Optional;
+import java.util.Set;
+
+public class KubectlCreate extends Kubectl.ApiClientBuilder<KubectlCreate>
+    implements Kubectl.Executable<KubernetesObject> {
+
+  KubectlCreate() {}
```
brendandburns, line 31: Delete since the default constructor is both no-arg and package private (afaik)

extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlCreate.java
```diff
@@ -0,0 +1,82 @@
...
+
+  private KubernetesObject targetObj;
+
+  public KubectlCreate load(KubernetesObject obj) {
```
brendandburns, line 35: `load` is kind of a confusing method name, I think. Perhaps `resource(...)`?

extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlCreate.java
```diff
@@ -0,0 +1,82 @@
...
+    this.targetObj = obj;
+    return this;
+  }
+
+  @Override
+  public KubernetesObject execute() throws KubectlException {
+    Discovery discovery = new Discovery(apiClient);
+    try {
+      Set<Discovery.APIResource> apiResources = discovery.findAll();
```
brendandburns, line 44: Do we want to do this every time? this could be pretty costly in the common case (e.g. Pods, Deployments, etc)

Perhaps only do this if the model mapper returns null?

extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlCreate.java
```diff
@@ -0,0 +1,82 @@
...
+      GenericKubernetesApi<KubernetesObject, KubernetesListObject> api =
+          new GenericKubernetesApi(
+              targetObj.getClass(),
+              KubernetesListObject.class,
+              gvk.getGroup(),
+              gvk.getVersion(),
+              apiResource.get().getResourcePlural(),
+              apiClient);
+      if (apiResource.get().getNamespaced()) {
+        String namespace =
```
brendandburns, line 66: I think we want to use the `namespace` member variable from `ApiClientBuilder` if it's not null. Then fall back to this if it is null.

extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlCreate.java
```diff
@@ -0,0 +1,82 @@
...
+                : targetObj.getMetadata().getNamespace();
+        KubernetesApiResponse<KubernetesObject> response =
+            api.create(namespace, targetObj, new CreateOptions());
+        return response.getObject();
+      } else {
+        KubernetesApiResponse<KubernetesObject> response =
+            api.create(targetObj, new CreateOptions());
+        return response.getObject();
+      }
+    } catch (Throwable t) {
```
brendandburns, line 78: Catch the specific errors, not generic `Throwable`

util/src/main/java/io/kubernetes/client/util/ModelMapper.java
```diff
@@ -242,9 +252,21 @@ public GroupVersionKind(String group, String version, String kind) {
       this.kind = kind;
     }
 
-    private String group;
-    private String version;
-    private String kind;
+    private final String group;
```
brendandburns, line 255: We should really use `GroupVersionKind` objects for storing these values.

brendandburns:
A few comments, but basically LGTM.

extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlCreate.java
yue9944882: >  package private (afaik)

https://stackoverflow.com/a/33633500/9568880 the default visibility of constructor aligns w/ the class, so we need to make it explicitly package-private

extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlCreate.java
yue9944882: updated

extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlCreate.java
yue9944882: yeah we need an in-memory cache for api discovery, i added a TODO comment and will do it in a follow-up. WDYT?

extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlCreate.java
yue9944882: added a `NamespacedApiClientBuilder` so that we can expicitly specify the namespace

extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlCreate.java
yue9944882: fixed

util/src/main/java/io/kubernetes/client/util/ModelMapper.java
yue9944882: rebased and fixed

PR #1193, opened 2020-08-25
Clean up various files

sarveshkaushal:
Following clean up:
1. Removed unnecessary semicolon.
2. Wildcard type argument ? explicitly extends Object class
3. Simplified if-else
4. Removed redundant variables

### DIFF
--- a/util/src/main/java/io/kubernetes/client/util/Yaml.java
+++ b/util/src/main/java/io/kubernetes/client/util/Yaml.java
@@ -12,6 +12,7 @@ limitations under the License.
 */
 package io.kubernetes.client.util;
 
+import io.kubernetes.client.common.KubernetesType;
 import io.kubernetes.client.custom.IntOrString;
 import io.kubernetes.client.custom.Quantity;
 import java.io.File;
@@ -201,7 +202,7 @@ public class Yaml {
    * @param data The list of YAML API objects
    * @return A String representing the list of YAML API objects.
    */
-  public static String dumpAll(Iterator<? extends Object> data) {
+  public static String dumpAll(Iterator<? extends KubernetesType> data) {
     return getSnakeYaml().dumpAll(data);
   }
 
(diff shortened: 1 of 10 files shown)


util/src/main/java/io/kubernetes/client/util/Yaml.java
```diff
@@ -201,7 +201,7 @@ public static void dump(Object object, Writer writer) {
    * @param data The list of YAML API objects
    * @return A String representing the list of YAML API objects.
    */
-  public static String dumpAll(Iterator<? extends Object> data) {
```
brendandburns, line 204: Now that we have `KubernetesType` I think this should be `? extends KubernetesType`

util/src/main/java/io/kubernetes/client/util/Yaml.java
```diff
@@ -211,7 +211,7 @@ public static String dumpAll(Iterator<? extends Object> data) {
    * @param data The list of YAML API objects.
    * @param output The writer to output the YAML String to.
    */
-  public static void dumpAll(Iterator<? extends Object> data, Writer output) {
```
brendandburns, line 214: same comment as above.

brendandburns:
one comment.

sarveshkaushal:
I have added the requested change.

util/src/main/java/io/kubernetes/client/util/Yaml.java
sarveshkaushal: Done

util/src/main/java/io/kubernetes/client/util/Yaml.java
sarveshkaushal: Done

yue9944882 (commented):
merging as @brendandburns 's comments been addressed

/lgtm
/approve

PR #1208, opened 2020-08-31
Code clean up

sarveshkaushal:
Following changes:

1. Removed the exception that is never thrown.
2. 'public' modifier for interface methods is redundant.
3. 'static' modifier for enum is redundant for inner enum.


brendandburns:
/lgtm
/approve

Are there spotless rules for formatting that would catch/apply these style rules?

PR #1209, opened 2020-08-31
generate model classes for Prometheus Operator v0.38.1

simingweng:
This PR includes:
1. expand the `-u` options in the common `generate.sh` script to access multiple URLs, because some upstream project may have multiple CRDs
2. create the `update.sh` script to maintain the generation of model classes from Prometheus Operator CRDs
3. generate model classes for Prometheus Operator v0.38.1, which is the current default version in the Prometheus Operator helm chart


yue9944882 (commented):
/lgtm

thanks! @simingweng

PR #1210, opened 2020-09-01
Publishes and documents crd-model-gen images

yue9944882:
/cc @brendandburns @simingweng


brendandburns:
Looks good to me, but the e2e tests failed with some `AbstractArchiver` stack overflow. Looks like a bug in the plugin?

I will re-run it.

yue9944882:
@brendanburns i opened the fix in #1212 and it looks working

PR #1222, opened 2020-09-03
Remove a sleep from CopyTest to deflake it.

brendandburns:


util/src/test/java/io/kubernetes/client/CopyTest.java
```diff
@@ -79,12 +82,13 @@ public void run() {
                   InputStream is = copy.copyFileFromPod(pod, "container", "/some/path/to/file");
                 } catch (IOException | ApiException e) {
                   e.printStackTrace();
+                } finally {
+                  s.release();
```
yue9944882, line 86: i think we should release the semaphore after complete reading from the stream, isnt it?

yue9944882: ah never mind. the test only verifies if the endpoint was requested.

yue9944882 (commented):
/lgtm

PR #1225, opened 2020-09-03
Refactor model mapper: adding group-version-resource mapping

yue9944882:


e2e/src/test/groovy/io/kubernetes/client/e2e/util/ModelMapperTest.groovy
```diff
@@ -0,0 +1,45 @@
...
+import io.kubernetes.client.openapi.models.V1Deployment
+import io.kubernetes.client.openapi.models.V1Pod
+import io.kubernetes.client.openapi.models.V1beta1CustomResourceDefinition
+import io.kubernetes.client.util.ClientBuilder
+import io.kubernetes.client.util.ModelMapper
+import spock.lang.Specification
+
+class ModelMapperTest extends Specification {
+
+    def "api-disocvery should work"() {
```
brendandburns, line 27: `api-discovery should work` (typo)

util/src/main/java/io/kubernetes/client/util/ModelMapper.java
```diff
@@ -42,16 +44,23 @@
...
+
   // Model's api-version midfix to kubernetes api-version
-  private static List<String> apiVersions = new ArrayList<>();
+  private static List<String> preBuiltApiVersions = new ArrayList<>();
+
+  private static Map<GroupVersionKind, Class<?>> classesByGVK = new ConcurrentHashMap<>();
+
+  private static Map<GroupVersionResource, Class<?>> classesByGVR = new ConcurrentHashMap<>();
+
+  private static Set<Discovery.APIResource> lastAPIDiscovery = new HashSet<>();
```
brendandburns, line 60: Should this be a ConcurrentHashMap too? (you can use it like a `Set`)

util/src/main/java/io/kubernetes/client/util/ModelMapper.java
```diff
@@ -114,33 +147,35 @@ public static void addModelMap(String group, String version, String kind, Class<
...
-   * Returns all the registered model classes.
+   * Gets the GVK by the given model class.
    *
-   * @return the all known classes
+   * @param clazz the clazz
+   * @return the group version kind by class
    */
-  public static Map<GroupVersionKind, Class<?>> getAllKnownClasses() {
-    return ImmutableMap.copyOf(classes);
+  public static GroupVersionKind getGroupVersionKindByClass(Class<?> clazz) {
```
brendandburns, line 163: Do we expect this to be called a bunch? The performance of this is pretty slow, so we may want to make the map bi-directional.

brendandburns:
Small  comments.

e2e/src/test/groovy/io/kubernetes/client/e2e/util/ModelMapperTest.groovy
```diff
@@ -0,0 +1,45 @@
...
+        given:
+        def apiClient = ClientBuilder.defaultClient();
+        def discovery = new Discovery(apiClient)
+        ModelMapper.refresh(discovery);
+
+        expect:
+
```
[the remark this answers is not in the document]
yue9944882: done

e2e/src/test/groovy/io/kubernetes/client/e2e/util/ModelMapperTest.groovy
yue9944882: fixed

util/src/main/java/io/kubernetes/client/util/ModelMapper.java
yue9944882: the map is actually thread-safe b/c it's immutable. we're replace the `lastAPIDiscovery` w/ a new immutable map upon every refreshing

util/src/main/java/io/kubernetes/client/util/ModelMapper.java
yue9944882: yeah ideally we should make this bi-directional. i was making it a GVK->class map b/c class->GVK doesnt look like a frequent case i suppose.. i left a TODO comment here and will do this in a follow-up. wdyt?

PR #1226, opened 2020-09-03
Improve logging in leader election

### LINKED ISSUE #1200: Lower logger levels
Would this project be open to a PR that lowers the log level of several messages? In general, I find that this library logs a lot of messages in the error level for transient issues that are retryable. For example, https://github.com/kubernetes-client/java/blob/31fbb3a52c6ca7a6ad6d71cd154b387191705adc/extended/src/main/java/io/kubernetes/client/extended/leaderelection/resourcelock/ConfigMapLock.java#L117 could be a warn, since the fact that it failed is already encoded in the return type (so the caller can decide whether that is an error or it is going to be retried down the line).

These error logs make my Ops team's life harder, since they end up as issues in [Sentry](https://sentry.io/), so that's my primary motivator to change the log level for some of the messages.

edrevo:
Fixes #1200 

CLA signature still pending. Working on it.

### DIFF
--- a/extended/src/main/java/io/kubernetes/client/extended/leaderelection/resourcelock/ConfigMapLock.java
+++ b/extended/src/main/java/io/kubernetes/client/extended/leaderelection/resourcelock/ConfigMapLock.java
@@ -20,6 +20,7 @@ import io.kubernetes.client.openapi.Configuration;
 import io.kubernetes.client.openapi.apis.CoreV1Api;
 import io.kubernetes.client.openapi.models.V1ConfigMap;
 import io.kubernetes.client.openapi.models.V1ObjectMeta;
+import java.net.HttpURLConnection;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicReference;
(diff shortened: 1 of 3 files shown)


edrevo:
CLA signed

extended/src/main/java/io/kubernetes/client/extended/leaderelection/resourcelock/ConfigMapLock.java
```diff
@@ -93,8 +94,8 @@ public boolean create(LeaderElectionRecord record) {
           coreV1Client.createNamespacedConfigMap(namespace, configMap, null, null, null);
       configMapRefer.set(createdConfigMap);
       return true;
-    } catch (Throwable t) {
-      log.error("failed to create leader election record as {}", t.getMessage());
+    } catch (ApiException e) {
+      log.error("received {} when creating configmap lock", e.getCode(), e);
```
brendandburns, line 98: I think we expect the possibility of `409` here too, so we should do the same `info/error` check here.

multiple different containers all starting at the same time will all initially try to create the lock, see the code here:

https://github.com/kubernetes-client/java/blob/master/extended/src/main/java/io/kubernetes/client/extended/leaderelection/LeaderElector.java#L248

cc @yue9944882 to double-check my understanding.

brendandburns:
Thanks for the PR!

One small comment.

extended/src/main/java/io/kubernetes/client/extended/leaderelection/resourcelock/ConfigMapLock.java
yue9944882: i think so. an `ApiException` will also be thrown here upon `409` but it's actually an expected behavior so it will be better if it's info-level instead of alerting error messages

edrevo: ah, yes, you are right! fixed

brendandburns:
Test failure is a flake, related to #1228 (I think) re-running

brendandburns:
@edrevo can you rebase this, we (I hope) fixed the flakiness.

yue9944882 (commented):
/lgtm

PR #1228, opened 2020-09-04
Change wiremock to use dynamic ports.

brendandburns:
Switching to semaphores made the test more flaky. I think this is because we've got the same port for all tests (I think) switch to dynamic ports is a good idea anyway.


yue9944882 (approved):
/lgtm

PR #1235, opened 2020-09-07
Enhancing model-mapper: adding bi-directional caches

yue9944882:


util/src/main/java/io/kubernetes/client/util/ModelMapper.java
```diff
@@ -114,8 +115,34 @@ public static void addModelMap(String group, String version, String kind, Class<
    */
   public static void addModelMap(
       String group, String version, String kind, String resourceNamePlural, Class<?> clazz) {
+    // TODO(yue9944882): consistency between bi-directional maps
```
brendandburns, line 118: I would just add a wrapper class that contains two `ConcurrentHashMap` and implements `Map` wdyt?

util/src/main/java/io/kubernetes/client/util/ModelMapper.java
```diff
@@ -53,12 +53,13 @@
   // Model's api-version midfix to kubernetes api-version
   private static List<String> preBuiltApiVersions = new ArrayList<>();
 
-  // TODO(yue9944882): make the map bi-directional
   private static Map<GroupVersionKind, Class<?>> classesByGVK = new ConcurrentHashMap<>();
-
-  // TODO(yue9944882): make the map bi-directional
   private static Map<GroupVersionResource, Class<?>> classesByGVR = new ConcurrentHashMap<>();
 
+  private static Map<Class<?>, GroupVersionKind> gvkByClasses = new ConcurrentHashMap<>();
```
brendandburns, line 59: I think we should encapsulate these two maps into a single bi-directional map. Ok for this to be a TODO.

brendandburns:
Small comments, also tests are failing.

util/src/main/java/io/kubernetes/client/util/ModelMapper.java
```diff
@@ -325,4 +362,22 @@ private static void initModelMap() throws IOException {
         .findFirst()
         .orElse(new MutablePair(null, name));
   }
+
+  static class BiDirectionalMap<K, V> {
+    private Map<K, V> kvMap = new HashMap<>();
```
brendandburns, line 367: I think we want this to be `ConcurrentHashMap<>()` ?

yue9944882: the getter/setter of the bi-directional-map are locked by `synchronized` (also to keep the consistency between the two maps) so a plain hash-map is sufficient, isn't it?

PR #1237, opened 2020-09-07
fix JUnit version property usage

sullis:


brendandburns:
This `pom.xml` is regenerated whenever we generate the code. Can you add it to .openapi-generator-ignore, so that it doesn't get overwritten next time?

sullis:
> This `pom.xml` is regenerated whenever we generate the code. Can you add it to .openapi-generator-ignore, so that it doesn't get overwritten next time?

Done.

kubernetes/.openapi-generator-ignore
```diff
@@ -4,6 +4,7 @@
 *sbt*
 git_push.sh
 README.md
+pom.xml
```
yue9944882, line 7: the default behavior of the underlying code-generator is to omit the pom.xml file when it's already present, so adding this the list won't break anything..

yue9944882 (commented):
/lgtm

kubernetes/.openapi-generator-ignore
brendandburns: oops, my bad. I asked him to do this...

PR #1239, opened 2020-09-08
Fixes failing test: Sleep 2s before verfication

yue9944882:
This reverts commit 07376cdd3dad5309c25ff790e880030d76d1974b.


yue9944882:
@brendandburns this temporary fix will be a debt but this is a minimal fix to tests pass for now..

brendandburns:
I think actually switching to the `wireMockRule` instance may fix things w/o the sleep?

brendandburns:
I'm going to merge this so that we can move forward, but I think that we can remove the sleep in a separate PR and test that.

I think the problem was using a static verify instead of the local wireMockRule.

PR #1242, opened 2020-09-09
Move away from static wire mocks.

brendandburns:
More flake reduction (I hope) move away from the default wireMock to the test specific `wireMockRule`


yue9944882 (commented):
/lgtm

PR #1243, opened 2020-09-09
Switch to a striped lock for better concurrency.

brendandburns:
Switch `BiDirectionalMap` to use a striped lock for better concurrency on par with `ConcurrentHashMap` see:

https://guava.dev/releases/20.0/api/docs/com/google/common/util/concurrent/Striped.html


yue9944882 (commented):
/lgtm

PR #1244, opened 2020-09-09
Refactor(kubectl): Automatic GVR discovery

yue9944882:


extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlDrain.java
```diff
@@ -59,7 +59,7 @@ public V1Node execute() throws KubectlException {
 
   private V1Node doDrain() throws KubectlException, ApiException, IOException {
     CoreV1Api api = new CoreV1Api(apiClient);
-    V1Node node = performCordon(api);
+    V1Node node = performCordon();
```
brendandburns, line 62: This means that two different client objects are created. I guess it's not that big a deal, but it seems wasteful in terms of memory.

brendandburns: Ah, nevermind, I see that the api is not actually used above. Please ignore this comment.

PR #1246, opened 2020-09-09
Restructure WebSocket initiation to make it more synchronous and deflake a test in the process.

brendandburns:


brendandburns:
hrmph there's something weird going on with this test. marking this PR WIP.

util/src/test/java/io/kubernetes/client/CopyTest.java
```diff
@@ -67,14 +69,16 @@ public void testUrl() throws IOException, ApiException, InterruptedException {
...
     try {
       copy.copyFileFromPod(pod, "container", "/some/path/to/file");
     } catch (IOException | ApiException e) {
       e.printStackTrace();
+    } finally {
+      s.release();
     }
-
-    Thread.sleep(2000);
+    s.acquire();
```
yue9944882, line 81: in this particular unit test, the client will be starting a websocket handshake against the wiremock server but because wiremock doesn't fully support websocket so the process of hand-shaking will be aborted abnormally. out of some reason, the wiremock will take a second-ish period to count the unfinished requests.. i guess that's why we have to sleep in this test..

brendandburns: I re-did how we initiate web sockets to make it more synchronous and resolve this.

brendandburns:
@yue9944882 I restructure WebSocket initiation a little bit so that the `WebSocket.stream(...)` method waits until the stream has been successfully opened (or failed) before returning.

I think this will generally make everything more sane/better, but please take a careful look since this code is a little twisty.

yue9944882:
```
No requests exactly matched. Most similar request was:  expected:<
GET
/api/v1/namespaces/default/pods/apod/attach?stdin=false&stdout=false&stderr=false&tty=false&container=acontainer

stdin: false
Query: stdout = true
stderr: false
tty: false
container: acontainer
> but was:<
GET
/api/v1/namespaces/default/pods/apod/attach?stdin=false&stdout=false&stderr=false&tty=false&container=acontainer

stdin: false
stdout: false
stderr: false
tty: false
container: acontainer
>
```

oddly the `stdout` query doesn't match any more. is the behavior changed?

brendandburns:
Ok, I updated the `Attach` code a little bit to check for errors and throw, this should now work (I hope!)

yue9944882:
@brendandburns can you rebase onto https://github.com/kubernetes-client/java/pull/1272 so that the failing test can pass?

yue9944882:
@brendandburns is it closed on purpose?

PR #1247, opened 2020-09-09
Update spotless version.

brendandburns:
Replacement for #1240 with formatting fixes added.

### DIFF
--- a/examples/src/main/java/io/kubernetes/client/examples/ProtoExample.java
+++ b/examples/src/main/java/io/kubernetes/client/examples/ProtoExample.java
@@ -53,8 +53,7 @@ public class ProtoExample {
     System.out.println(ns);
     if (ns.object != null) {
       namespace =
-          ns.object
-              .toBuilder()
+          ns.object.toBuilder()
               .setSpec(NamespaceSpec.newBuilder().addFinalizers("test").build())
               .build();
       // This is how you would update an object, but you can't actually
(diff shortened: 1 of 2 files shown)


brendandburns:
I'm really confused. #1240 seems to indicate this code change is needed (test failing) but this has tests failing too...

¯\_(ツ)_/¯

yue9944882:
yeah, actually i attempted to fix it when i saw this.. wonder if if it's a bug in the upstream plugin..

brendandburns:
@yue9944882 I took a deeper look at this, it looks like it is a Java8 vs java11 thing.

Java8/9/10 are failing, Java11/12/13 are passing.

What do you want to do? I suggest that we disable the formatting test for Java 8/9/10.

yue9944882:
> @yue9944882 I took a deeper look at this, it looks like it is a Java8 vs java11 thing.
Java8/9/10 are failing, Java11/12/13 are passing.
What do you want to do? I suggest that we disable the formatting test for Java 8/9/10.

@brendandburns thanks for the clue, i updated https://github.com/kubernetes-client/java/pull/1262 to move the format verification to a separated job and it seems working now. one fun fact i confirmed is that currently the output of formatter plugin is diverging between java8 and java11 (e.g. running formatter under java8 doesnt demand any change in the ProtoExample file, while in java11 it does). i used java8 formatting in the new github action job b/c the current maven compiler plugin uses java8.. wdyt?

brendandburns:
Fine with me, merged the fixed PR and closing this one.

PR #1251, opened 2020-09-10
Feat: Kubectl get implementation

yue9944882:
ideally i think this can be a replacement for GenericKubernetesApi

### DIFF
--- a/extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlGet.java
+++ b/extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlGet.java
@@ -1,0 +1,135 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package io.kubernetes.client.extended.kubectl;
+
+import io.kubernetes.client.common.KubernetesListObject;
+import io.kubernetes.client.common.KubernetesObject;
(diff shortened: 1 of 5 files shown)


extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlGet.java
```diff
@@ -0,0 +1,130 @@
...
+    } catch (ApiException e) {
+      throw new KubectlException(e);
+    }
+  }
+
+  private boolean isNamespaced() {
+    return !StringUtils.isEmpty(namespace);
+  }
+
+  public class KubectlGetSingle implements Kubectl.Executable<ApiType> {
```
brendandburns, line 94: This is going to get messy/confusing to people because:
 
`Kubectl.get().namespace(foo).name(bar).execute()` will work, but `Kubectl.get().name(bar).namespace(foo)` won't.

I think we should either get rid of this class and merge the execute methods, or make this class extend from `KubectlGet`

brendandburns:
Generally looks good, one comment about the interface.

extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlGet.java
yue9944882: > Kubectl.get().namespace(foo).name(bar).execute() will work, but Kubectl.get().name(bar).namespace(foo) won't.

yeah i agree this can be confusing to the users. while the point here is that we're returning single object upon specifying resource name, and return a list of objects if the name absent. java doesn't support multi-return or overriding return type for the execute method  so merging or overriding the execute methods won't work.. 

in a word, if we're to make both `namespace(foo).name(bar)` and `name(bar).namespace(foo)` work, the execute method will have to return a list of object even when getting one object, which can be even more confusing i suppose?

brendandburns: Hrm, why not just have, and then you can do:

`return new KubectlGetSingle().namespace(namespace).name(name)` in the function that adds the name, then it will work either way.

```
public KubectlGetSingle<ApiType> extends ResourceBuilder<...>, implements Kubectl.Executable<ApiType>

yue9944882: done

PR #1259, opened 2020-09-11
fix issue 912: do dedupe with resourceVersion

fishautumn:
reason: on timeout-reconnect it will add new a new copy to all objects in FIFO items map.
fix: check resourceVersion when adding object copy, skip if newly added object has same resourceVersion with previous latest object.

### DIFF
--- a/util/src/main/java/io/kubernetes/client/informer/cache/DeltaFIFO.java
+++ b/util/src/main/java/io/kubernetes/client/informer/cache/DeltaFIFO.java
@@ -447,7 +447,8 @@ public class DeltaFIFO {
     MutablePair<DeltaType, KubernetesObject> deletionDelta = isDeletionDup(d1, d2);
 
     // TODO: remove this after the cause of memory leakage is confirmed
-    // Squashing deltas w/ the same resource version, note that is a temporary fix that eases memory intensity.
+    // Squashing deltas w/ the same resource version, note that is a temporary fix that eases memory
+    // intensity.
     if (deletionDelta != null) {
       return deletionDelta;
     }
--- a/util/src/main/java/io/kubernetes/client/informer/cache/DeltaFIFO.java
+++ b/util/src/main/java/io/kubernetes/client/informer/cache/DeltaFIFO.java
@@ -448,8 +448,11 @@ public class DeltaFIFO {
     if (deletionDelta != null) {
       return deletionDelta;
     }
-    if (d1.getLeft() != DeltaType.Deleted && d2.getLeft() != DeltaType.Deleted &&
-            StringUtils.equals(d1.getRight().getMetadata().getResourceVersion(), d2.getRight().getMetadata().getResourceVersion())) {
+    if (d1.getLeft() != DeltaType.Deleted
+        && d2.getLeft() != DeltaType.Deleted
+        && StringUtils.equals(
+            d1.getRight().getMetadata().getResourceVersion(),
+            d2.getRight().getMetadata().getResourceVersion())) {
       return d1;
     }
     return null;
--- a/util/src/main/java/io/kubernetes/client/informer/cache/DeltaFIFO.java
+++ b/util/src/main/java/io/kubernetes/client/informer/cache/DeltaFIFO.java
@@ -29,6 +29,7 @@ import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import org.apache.commons.collections4.CollectionUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.MutablePair;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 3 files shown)


brendandburns:
I believe that the Go implementation of the informer also returns an update on re-list.  I think we'd prefer to match that implementation for the principle of least suprise.

@yue9944882 wdyt?

fishautumn:
in golang implementation, the readtimeout of watch is 5~10minutes (randomly).
in java implementation, ReflectorRunnable specifies 5 minutes, while BatchV1Api only applys it as a URL query parameter, not changing okhttp client options. default okhttp client's readtimeout is 10 seconds.

if the leak issue exists in both implementation, java implementation will get OOM in 1/60~1/30 time compared with golang implementation.

yue9944882:
> I believe that the Go implementation of the informer also returns an update on re-list

the Go implementation doesn't actually perform any de-dup when dealing w/ two lists w/ the same resource version --- this can happen upon some relatively inactive resources like namespaces/serviceaccounts/... i wrote a test to show the behavior how delta-fifo processes identical lists, and i did the same on golang library and the result shows that they're working exactly the same. so i wonder if the memory leakage is caused by the abortion of the processor thread consuming the DeltaFIFO. @fishautumn can you share w/ us how you reproduce/locate the leakage?

but i think the proposed change in this pull is still valuable at least before we getting know the cause

util/src/main/java/io/kubernetes/client/informer/cache/DeltaFIFO.java
```diff
@@ -443,7 +444,18 @@ private void syncKeyLocked(String key) {
    */
   private MutablePair<DeltaType, KubernetesObject> isDuplicate(
       MutablePair<DeltaType, KubernetesObject> d1, MutablePair<DeltaType, KubernetesObject> d2) {
-    return isDeletionDup(d1, d2);
+    MutablePair<DeltaType, KubernetesObject> deletionDelta = isDeletionDup(d1, d2);
+    if (deletionDelta != null) {
+      return deletionDelta;
+    }
+    if (d1.getLeft() != DeltaType.Deleted
```
yue9944882, line 451: // TODO: remove this after the cause of memory leakage is confirmed
// Squashing deltas w/ the same resource version, note that is a temporary fix that eases memory intensity.

fishautumn:
in ReflectorRunnable.run(), it calls list method without resourceVersion. golang client Reflector.Run() calls list method with resourceVersion.

I suspect this is the reason of java client memory leak.

fishautumn:
I suspect java client has issue with the http readtimeout setting with watching apis. BatchV1Api only applys timeout to URL parameter, not changing okhttp client options, while default okhttp client's readtimeout is 10 seconds. this speeds up the java client OOM.

fishautumn:
@yue9944882 here is my steps to repro it:
1. create a informer for V1Job list
2. create a thread that create a job every 10 seconds
3. after running for 17 hours, the program consumes all memory and spends all CPU for GC activity.

fishautumn:
> in ReflectorRunnable.run(), it calls list method without resourceVersion. golang client Reflector.Run() calls list method with resourceVersion.
> 
> I suspect this is the reason of java client memory leak.

I tried adding resourceVersion to list call, but it doesn't wok. according to api reference, resourceVersion is only working in watch call. [list-job-v1-batch](https://v1-17.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#list-job-v1-batch)

fishautumn:
so the main difference with golang implementation is the timeout setting. java client is 10 seconds, while golang client is 5~10 minutes (randomly).

this makes it requires 45 times of list size to repro the issue in golang client.

yue9944882:
> I tried adding resourceVersion to list call, but it doesn't wok. according to api reference, resourceVersion is only working in watch call. list-job-v1-batch
> so the main difference with golang implementation is the timeout setting. java client is 10 seconds, while golang client is 5~10 minutes (randomly).
> this makes it requires 45 times of list size to repro the issue in golang client.

@fishautumn thanks, the informer explicitly requires the okhttp client's read-timeout to be infinite to align w/ the golang library, see [here](https://github.com/kubernetes-client/java/blob/master/examples/src/main/java/io/kubernetes/client/examples/InformerExample.java#L38-L40). there's no such thing as read-time in the golang client. if that is the case, i'm leaning on adding another defensive check (probably upon constructing the informer instance) that the given ApiClient#OkHttpClient's read-time is set to 0. the current state of the pull is good to me. @brendandburns 

/lgtm

brendandburns:
SGTM, throw `IllegalStateException` or `IllegalArgumentException` probably.

fishautumn:
> @fishautumn thanks, the informer explicitly requires the okhttp client's read-timeout to be infinite to align w/ the golang library, see [here](https://github.com/kubernetes-client/java/blob/master/examples/src/main/java/io/kubernetes/client/examples/InformerExample.java#L38-L40). there's no such thing as read-time in the golang client. if that is the case, i'm leaning on adding another defensive check (probably upon constructing the informer instance) that the given ApiClient#OkHttpClient's read-time is set to 0. the current state of the pull is good to me. @brendandburns
> 
> /lgtm

ok3httpclient and ApiClient are passing to informer through a CallGenerator object. I cannot find a proper way to do the defensive check.

fishautumn:
* pre-check is adding to SharedInformerFactory. it throws IllegalArgumentException if ApiClient's timeout is not zero.
* there is a piece of code that is setting timeout to zero if not, but I suspect it's not working for watch call. I refined it.
* I suspect the timeout is also not working for GenericKubernetesApi, so I added a comment. fix me if I'm wrong.

yue9944882 (commented):
/lgtm

PR #1275, opened 2020-09-17
Release 10.0.0

### LINKED ISSUE #1120: Cut a 10.0.0 release?
We just cut a 9.0.0 release, but it might make sense to also cut a 10.0.0 with 1.18 library support.

On the other hand, 1.18 removes the `v1beta1/Deployment` APIs so it will definitely be a breaking change for some folks.

@yue9944882 wdyt?

brendandburns:
Closes #1120 

This has been pushed out to Sonatype and should be visible in 24hr.

### DIFF
--- a/client-java-contrib/cert-manager/pom.xml
+++ b/client-java-contrib/cert-manager/pom.xml
@@ -6,11 +6,11 @@
     <parent>
         <groupId>io.kubernetes</groupId>
         <artifactId>client-java-parent</artifactId>
-        <version>10.0.0</version>
+        <version>10.0.1-SNAPSHOT</version>
         <relativePath>../../pom.xml</relativePath>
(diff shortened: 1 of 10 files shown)


brendandburns:
@yue9944882 

This actually adjusted the versions for all of the contrib libraries too. wdyt?

yue9944882:
ah dang, we should have override the release plugin for contrib modules so that they can be skipped. but it's okay to published for one time i guess, but we will just not be actively supporting it?

brendandburns:
I can revert the pom.xml changes and then build a `10.0.1` release if we want to.

Let me know what you think.

yue9944882 (commented):
@brendandburns i think it's okay to keep the change for this release branch

/lgtm

PR #1285, opened 2020-09-21
Bugfix(kubectl): Name argument is not required for kubectl replace

yue9944882:
and a couple of refactors in the e2e tests


brendandburns:
Looks good, but why are the e2e tests being deleted?

yue9944882:
> Looks good, but why are the e2e tests being deleted?

@brendandburns they're not deleted, it's squashed into one test class.

brendandburns:
ok, thanks.

/lgtm
/approve

PR #1302, opened 2020-10-01
GenericKubernetesApi httpStatusCode

mvaal:
Currently returns -1 for all successful calls.


mvaal:
Ignore this... basically, seems like you get either a status, or an object back (which i guess can return status). Seems a little confusing from the user side.

yue9944882:
> basically, seems like you get either a status, or an object back (which i guess can return status). 

@mvaal yeah actually this is exactly the kubernetes apiserver's behavior --- returning multiple types is not supported in java --- and that's one of points drives us to provide this generic api. i will improve the document as enhancement

PR #1304, opened 2020-10-04
Improved README.md

codewithkanishq:
I improved a little user experience changes kindly see to it.
Waiting for a positive reply from your side


brendandburns:
This change isn't useful.

We would value contributions to the project, please take a look at the open issues and fix one of those. Thanks.

I think it is Hacktoberfest spam. Closing.

PR #1305, opened 2020-10-04
Update README.md

Amanpatel13:
# tech stack
java 
hope it is usefull


brendandburns:
This change isn't useful.

We would value contributions to the project, please take a look at the open issues and fix one of those. Thanks.

I think it is Hacktoberfest spam. Closing.

PR #1314, opened 2020-10-08
Fix pager iterator and add test.

### LINKED ISSUE #1309: Wrong Iterator interface implementation
https://github.com/kubernetes-client/java/blob/31fbb3a52c6ca7a6ad6d71cd154b387191705adc/extended/src/main/java/io/kubernetes/client/extended/pager/Pager.java#L119

Two subsequent calls of `hasNext()`return two different values when no items were found.

First call always returns `true` so foreach loop fails with:
```
java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
at java.util.ArrayList.rangeCheck(ArrayList.java:657)
at java.util.ArrayList.get(ArrayList.java:433)
at io.kubernetes.client.extended.pager.Pager$PagerIterator.next(Pager.java:144)
```
Which goes against Iterator's contract:
https://docs.oracle.com/javase/7/docs/api/java/util/Iterator.html#hasNext()

brendandburns:
Fixes #1309

### DIFF
--- a/extended/src/main/java/io/kubernetes/client/extended/pager/Pager.java
+++ b/extended/src/main/java/io/kubernetes/client/extended/pager/Pager.java
@@ -59,7 +59,9 @@ public class Pager<ApiType extends KubernetesObject, ApiListType extends Kuberne
    */
   @Override
   public Iterator<ApiType> iterator() {
-    return new PagerIterator();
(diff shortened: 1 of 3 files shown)


yue9944882 (commented):
/lgtm

PR #1315, opened 2020-10-08
Buffer writes > 16 MiB into multiple web socket messages.

### LINKED ISSUE #1311: Websocket writes - 16MiB limit
According to `okhttp3.internal.ws.RealWebSocket.send(bytes: ByteString): boolean` javadocs, if the payload is larger than 16MiB, the write will be terminated as soon as 16MiB has been uploaded and the connection will be closed.

1. I did not find a single word about this limitation in the kubernetes-client/java/README.md documentation. 

2. If the limitation is present, kubernetes-client could buffer the stream to <16MiB chunks and upload them separately. Or decorate the OutputStream and count bytes passed on to the websocket, and issue a new socket.send() every time 16MiB is transferred (to mitigate buffer-induced-delays in my previous proposal). It's the limitation introduced by the libraries chosen by the client's authors, so I don't believe client's users should be concerned about the client's implementation details and their limitations.

3. This affects both: Copy and Exec (which is natural, since Copy is a derivate of Exec)

Here's the javadoc:

```
  /**
   * Attempts to enqueue {@code bytes} to be sent as a the data of a binary (type {@code 0x2})
   * message.
   *
   * <p>This method returns true if the message was enqueued. Messages that would overflow the
   * outgoing message buffer (16 MiB) will be rejected and trigger a
   * {@linkplain #close graceful shutdown} of this web socket. This method returns false in that
   * case, and in any other case where this web socket is closing, closed, or canceled.
   *
   * <p>This method returns immediately.
   */
  boolean send(ByteString bytes);
```

brendandburns:
Fixes #1311

### DIFF
--- a/util/src/main/java/io/kubernetes/client/util/WebSocketStreamHandler.java
+++ b/util/src/main/java/io/kubernetes/client/util/WebSocketStreamHandler.java
@@ -241,10 +241,20 @@ public class WebSocketStreamHandler implements WebSockets.SocketListener, Closea
           }
         }
       }
-      byte[] buffer = new byte[length + 1];
-      buffer[0] = stream;
-      System.arraycopy(b, offset, buffer, 1, length);
-      WebSocketStreamHandler.this.socket.send(ByteString.of(buffer));
+      int bytesWritten = 0;
+      int remaining = length;
+      while (bytesWritten < length) {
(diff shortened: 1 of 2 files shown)


yue9944882 (commented):
/lgtm

PR #1316, opened 2020-10-08
Add a helper method to obtain the namespace of a pod.

### LINKED ISSUE #1313: Get Namespace When Running In Pod
By default there is a file called `namespace` located at `/var/run/secrets/kubernetes.io/serviceaccount` when running in a pod.  Is there a way of reading that files value in the client library?  I poked around but couldnt find anything.

brendandburns:
Fixes #1313


util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
```diff
@@ -48,6 +48,7 @@
   private byte[] caCertBytes = null;
   private boolean verifyingSsl = true;
   private Authentication authentication;
+  private String namespace;
```
yue9944882, line 51: is the new field in use? @brendandburns

brendandburns: oops, removed.

yue9944882 (commented):
/lgtm

ryanjbaxter:
Thanks!

PR #1320, opened 2020-10-10
Add async methods for Copy that return Futures.

brendandburns:


yue9944882 (commented):
/lgtm

ah dang i didn't even notice it!

Issue #1310, opened 2020-10-07
websocket writes -- no feedback when the write is complete


netikras:
Whenever a WRITE is issued to the attached container (e.g. file upload, long command, etc.), the data-to-be-written is added to the write queue, which is drained asynchronously by OKHttp3. And there is no way to get feedback when the write is complete. If I'm writing an integration with k8s, I'd like to know when my file has been uploaded, so that I could go to the next step and use it in the container. OFC I could do Thread.sleep(120_000L), but it is not exactly the right way to go IMO...

Please either make the writes synchronous, so the user can handle the concurrency flows himself, or add an ability to get feedback when the write is complete (a Future<?> would be nice). Without this feedback, any commands sent to the pod and any file uploads are subject for race condition without any way to apply synchronization techniques.

netikras:
Related: https://github.com/kubernetes-client/java/issues/1127

brendandburns:
Yep, this is fair. I assume that you are using the `Copy` helper class?

netikras:
@brendanburns I was using a custom Copy implementation, based on Exec and `dd if=/dev/stdin` (just like `Copy` does), because 
- I cannot guarantee all the pods will have base64 installed
- I need to upload files from the working memory - NOT from the filesystem, and `Copy` only accepts a `Path` as a source (Why not InputStream though?)
- If the pod is unable to accept data with `dd` - I can swap the acceptance mechanism with any other command, reading from stdin (e.g. tee)
- lack of _append_ functionality in `Copy`


And just now I have tried the `Copy` utility and I can confirm that it is also uploading data completely asynchronously.

P.S. I have found kind of a workaround for that - caching the command's onClose event and completing the future then. However, this approach refrains me from reusing the connection/stream multiple times.
Since I have a sort-of working workaround now [the `Copy` is still not a subject for], the 16MiB problem [in another ticket I've raised today] is a lot more severe than this one.

netikras:
BTW, what's with the base64 in the `Copy` util? Why compress the data and then explode it with b64? WebSockets' should be perfectly capable to deal with binary streams AFAIK?

brendandburns:
In general, I agree with your feedback that we need to have a `Future` to represent the completion of an Exec. You should be able to wait on the `Process` to complete though which should indicate that it's done.

Regarding base64, I agree it is probably unnecessary, it was used for single file cases since it's easier than tarballing a single file, but for multiple files it's probably not needed.

netikras:
[...]

Issue #295, opened 2018-06-20
Support for Kubernetes 1.10?


ggprod:
Hello,

does this library support Kubernetes 1.10 yet?

Thanks

kondapally1989:
yes it does. it seems README is not updated.

ggprod:
Thanks!  Yes, I suspected that

PR #1322, opened 2020-10-10
Exec.exec returning Future and hiding Process as an implementation detail

netikras:
as per https://github.com/kubernetes-client/java/issues/1310


netikras:
I signed it

netikras:
image
If only I knew what's wrong... :)

brendandburns:
You need to run `mvn spotless:apply`

kubernetes/src/main/java/io/kubernetes/client/custom/AsyncPump.java
```diff
@@ -0,0 +1,123 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package io.kubernetes.client.custom;
+
+import java.io.InputStream;
+import java.util.concurrent.*;
```
brendandburns, line 16: No wildcard imports please.

kubernetes/src/main/java/io/kubernetes/client/custom/AsyncPump.java
```diff
@@ -0,0 +1,123 @@
...
+
+public class AsyncPump implements Runnable {
```
brendandburns, line 18: Quick JavaDoc for the use of this class.

kubernetes/src/main/java/io/kubernetes/client/custom/AsyncPump.java
```diff
@@ -0,0 +1,123 @@
...
+    this(consumer, NoisyRunnable.NOP);
+  }
+
+  public AsyncPump setExecutorService(ExecutorService executorService) {
+    this.executorService = executorService;
+    return this;
+  }
+
+  public AsyncPump stop() {
+    if (thread != null)
```
brendandburns, line 51: Curly braces around all conditions.

kubernetes/src/main/java/io/kubernetes/client/custom/AsyncPump.java
```diff
@@ -0,0 +1,123 @@
...
+  public void run() {
+    int nRead;
+    try {
+      while ((nRead = is.read(buff)) > -1) {
+        byte[] giveAway = new byte[nRead];
+        System.arraycopy(buff, 0, giveAway, 0, nRead);
+        consumer.accept(giveAway);
+      }
+      onFinish.run();
+    } catch (Exception e) {
```
brendandburns, line 102: Catch specific exceptions.

kubernetes/src/main/java/io/kubernetes/client/custom/AsyncPump.java
```diff
@@ -0,0 +1,123 @@
...
+      try {
+        errorHandler.accept(e);
+      } catch (Exception exception) {
```
brendandburns, line 105: here too

kubernetes/src/main/java/io/kubernetes/client/custom/AsyncPump.java
```diff
@@ -0,0 +1,123 @@
...
+        promise.completeExceptionally(exception);
+      }
+    }
+    promise.complete(null);
+  }
+
+  public interface NoisyConsumer<T> {
```
brendandburns, line 112: What is the purpose of these interfaces?

util/src/main/java/io/kubernetes/client/Exec.java
```diff
@@ -26,19 +28,19 @@
...
 import io.kubernetes.client.util.WebSocketStreamHandler;
 import io.kubernetes.client.util.WebSockets;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.InputStreamReader;
-import java.io.OutputStream;
-import java.io.Reader;
-import java.io.UnsupportedEncodingException;
+
+import java.io.*;
```
brendandburns, line 32: As above no wildcard imports please.

kubernetes/src/main/java/io/kubernetes/client/custom/AsyncPump.java
netikras: These two do what normal Runnable and Consumer do, except they allow to throw an exception. Allows writing cleaner lambdas w/o try-catch blocks inside (ofc all the failures are to be handled by the caller). However, I only use them in the AsyncPump, so I could move them to there as public inner interfaces if you'd like me to.

kubernetes/src/main/java/io/kubernetes/client/custom/AsyncPump.java
netikras: Can't do that here. onFinish is a callback interface - the caller has no clue what exceptions can occur, but whatever exceptions happen, they are delegated to the errorHandler. That's what the NoisyConsumer is for. The consumer can catch exceptions by itself if it sees the need to, and the ones it misses (e.g. unchecked ones, or missed deliberately) will be caught by the AsyncPump.

kubernetes/src/main/java/io/kubernetes/client/custom/AsyncPump.java
netikras: Oh, and that's probably my mistake - it's a copy-paste of the utility from my project (as I have copy-pasted Exec.exec() and Exec.execSimple() from there too). If you see the AsyncPump is not likely to be reusable in this project, I can easily replace it with raw `new Thread(() -> {/**/}).start()`

yue9944882 (commented):
LGTM

can you format the code by running `mvn spotless:apply`?

netikras:
Absolutely. Done

util/src/main/java/io/kubernetes/client/Exec.java
```diff
@@ -191,6 +199,135 @@ public Process exec(
...
+          (code, timeout) -> {
+            process.destroy();
+            waitForProcessToExit(process, timeout, cmdStr, err -> onError.accept(err, io));
+            // processWaitingThread will handle the rest
+          });
+      io.setStdin(process.getOutputStream());
+      io.setStdout(process.getInputStream());
+      io.setStderr(process.getErrorStream());
+      Thread processWaitingThread =
+          new Thread(
```
brendandburns, line 261: This method is going to be hard to test because of this. Can you use an `ExecutorService` instead? That can be injected in the test.

util/src/main/java/io/kubernetes/client/Exec.java
```diff
@@ -47,6 +54,7 @@
   private static final Logger log = LoggerFactory.getLogger(Exec.class);
 
   private ApiClient apiClient;
+  private long graceTimeoutMs = 5_000L;
```
brendandburns, line 57: Given that this is only used in one method, make this a parameter instead? (or delete)

util/src/main/java/io/kubernetes/client/Exec.java
```diff
@@ -191,6 +199,135 @@ public Process exec(
...
+      } catch (InterruptedException e) {
+        onError.accept(e);
+      }
+      log.debug("Process closed={}: {}", exited, cmdStr);
+      if (!exited && process.isAlive()) {
+        beforeTimeout = false;
+        log.warn("Process timed out after {} ms. Shutting down: {}", timeoutMs, cmdStr);
+        process.destroy();
+        try {
+          exited = process.waitFor(graceTimeoutMs, TimeUnit.MILLISECONDS);
```
brendandburns, line 305: It's a little weird to wait after the user specified timeout. I think we should delete all of the grace period logic. The user should just pass in whatever timeout they want.

netikras: Have you ever closely monitored a Tomcat process shutting down? :) It takes a few seconds to clear all the pools, close all the connections, commit all the remaining transactions after the JVM receives a signal #15. Similar delays are observed in multi-classloader applications (e.g. osgi-based). Shutting down the application gracefully might take a tad more than just a heartbeat.

In line 303 I'm (hopefully) sending a sigterm, so the application starts its shutdown sequence. Some applications might take considerable amounts of time to shutdown (e.g. a JVM with a debugger attached), thus I cannot guarantee the process will be stopped immediately after the timeout in line #295. Telling the caller the process has stopped could lead to resources' exhaustion in some scenarios (i.e. BUGS). So I'm giving the process a chance to finish its business itself in 5 seconds (by default). If it fails to do so (could be many reasons), then I'm using my last card - sending (hopefully) a sigkill.

Sending a sigkill must always be the last resort.

I'd really like to keep the grace period in the code. I can hardcode the grace period if you'd like me to (could be extracted later on if required), but I'd sleep better knowing I've pushed the correct process termination procedure to a public repo :)

util/src/main/java/io/kubernetes/client/Exec.java
netikras: Sure thing!

```
  private void runAsync(String taskName, Runnable task) {
    if (this.executorService == null) {
      Thread thread = new Thread(task);
      thread.setName(taskName);
      thread.start();
    } else {
      executorService.submit(task);
    }
  }
```

Will this cut it? Using an `ExecutorService` if there is supplied one, and falling back to standalone Thread if there isn't one. And a simple Exec.setExecutorService() to supply the executor. Although the executor would also be used in this sole method (unless reused somewhere else later on).

I really don't feel like adding yet another parameter to the method, since there already are 8+ and the ES would be mostly optional.

Another option - use a protected field. or even a `protected runAsync(String taskName, Runnable task){...}`, which could be overridden in tests. Which approach would you prefer?

util/src/main/java/io/kubernetes/client/Exec.java
netikras: I'll delete it then.

util/src/main/java/io/kubernetes/client/Exec.java
brendandburns: This approach is fine. Thanks!

util/src/main/java/io/kubernetes/client/Exec.java
brendandburns: But this isn't a real process, it's a remote process running on a different service in a container on the cluster, so I don't think that stuff applies here.

netikras: That's the obvious consequence of Liskov's principle violation :) Frankly, I don't know what does this client does when I ask it to terminate. So which one does reliably destroy the connection and remote process? destroy() or destroyForcibly()? I'll use the one you name and remove the grace period then.

util/src/main/java/io/kubernetes/client/Exec.java
netikras: which is "this"? :) 
I have gone with the `protected runAsync(String taskName, Runnable task){...}` (seemed more appropriate). Shall I leave it or replace it with the code block in my comment above?

util/src/main/java/io/kubernetes/client/Exec.java
brendandburns: `destroy()` does the right thing here (it's not 100% 'reliable' in the sense that nothing in distributed systems is 100% "reliable"), but `destroy` closes the web socket connection which should have the net effect of stopping the process on the remote server (eventually), but from the client's perspective, it's terminated once the socket is closed.

Code is here:
https://github.com/kubernetes-client/java/blob/master/util/src/main/java/io/kubernetes/client/Exec.java#L476

brendandburns:
This looks good to me, can you add a unit test and squash the commits? Then it will be ready to merge.

Thanks!

yue9944882 (commented):
/LGTM

yue9944882 (commented):
/approve

PR #1330, opened 2020-10-15
Provide base images for saving installation time

yue9944882:
making `docker.pkg.github.com/..` images public requires `kubernetes-client` admin access so i copied the images under my repository so that users can pull anonymously..

@brendandburns @kumars17


brendandburns:
/lgtm
/approve

@yue9944882 do you need me to do anything to make the images public?

Issue #1329, opened 2020-10-14
Reflector doesn't shutdown gracefully on thread interruption


iyashu:
I'm trying to build an informer using `DefaultSharedIndexInformer` watching for ConfigMap resources and noticed that on stopping the informer (using `DefaultSharedIndexInformer::stop`) `main` function doesn't exits. After taking thread dump using jstack, I've found that reflector thread is not terminated. Below is the complete example to reproduce with java version 1.8.0_191 and client version 9.0.2.
<details>
<summary> Example Code </summary>

```java
package io.kubernetes.client.util.example;

import com.google.gson.reflect.TypeToken;
import io.kubernetes.client.informer.ListerWatcher;
import io.kubernetes.client.informer.ResourceEventHandler;
import io.kubernetes.client.informer.SharedInformer;
import io.kubernetes.client.informer.impl.DefaultSharedIndexInformer;
import io.kubernetes.client.openapi.ApiClient;
import io.kubernetes.client.openapi.ApiException;
import io.kubernetes.client.openapi.apis.CoreV1Api;
import io.kubernetes.client.openapi.models.V1ConfigMap;
import io.kubernetes.client.openapi.models.V1ConfigMapList;
import io.kubernetes.client.util.CallGeneratorParams;
import io.kubernetes.client.util.Config;
import io.kubernetes.client.util.Watch;
import io.kubernetes.client.util.Watchable;
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;

import java.util.concurrent.ExecutorService;
import java.util.concurrent.Executors;
import java.util.concurrent.TimeUnit;
import java.util.concurrent.TimeoutException;

public class SimpleUsageExample {
    private static final Logger logger = LoggerFactory.getLogger(SimpleUsageExample.class);

    private static final ExecutorService DEFAULT_EXECUTOR = Executors.newFixedThreadPool(20, r -> {
        Thread thread = new Thread(r, "ConfigMapThread");
        thread.setDaemon(true);
        return thread;
    });


    public static void main(String[] args) throws Exception {
        logger.info("starting up the main");
        ExecutorService executorService = DEFAULT_EXECUTOR;

        V1ConfigMap configMap;

        ApiClient apiClient = Config.fromConfig("emma.kubeconfig");
        apiClient.setReadTimeout(0);
        String fieldSelector = "metadata.name=test";
        String namespace = "configmap-test";

        CoreV1Api coreV1Api = new CoreV1Api(apiClient);

[...]

PR #1332, opened 2020-10-16
fix reflector graceful shutdown by closing underlying watch

iyashu:
Fixes: #1329 

@brendandburns @yue9944882

### DIFF
--- a/util/src/main/java/io/kubernetes/client/informer/cache/ReflectorRunnable.java
+++ b/util/src/main/java/io/kubernetes/client/informer/cache/ReflectorRunnable.java
@@ -20,6 +20,7 @@ import io.kubernetes.client.openapi.models.V1ListMeta;
 import io.kubernetes.client.openapi.models.V1ObjectMeta;
 import io.kubernetes.client.util.CallGeneratorParams;
 import io.kubernetes.client.util.Watchable;
+import java.io.IOException;
 import java.net.ConnectException;
 import java.time.Duration;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


util/src/main/java/io/kubernetes/client/informer/cache/ReflectorRunnable.java
```diff
@@ -88,24 +89,30 @@ public void run() {
...
-          watch =
+          Watchable<ApiType> newWatch =
               listerWatcher.watch(
                   new CallGeneratorParams(
                       Boolean.TRUE,
                       lastSyncResourceVersion,
                       Long.valueOf(Duration.ofMinutes(5).toMillis()).intValue()));
-          watchHandler(watch);
+
+          synchronized (this) {
```
brendandburns, line 108: What is this synchronization protecting? I see other access to `isActive` that isn't synchronized (for example)

iyashu: @brendandburns as of now synchronized is used to protect the [watch variable](https://github.com/iyashu/java/blob/5f377e6a009682e98f57e21cafd1490c998c3905/util/src/main/java/io/kubernetes/client/informer/cache/ReflectorRunnable.java#L39) holding the current watchable instance. Essentially we want to close the watch from caller thread and re-assign it(as we re-create watches) in runnable thread thus requiring the access synchronization for it. `isActive` other hand, is atomicBoolean tracking if watch is required to be active or not (essentially from the caller thread), so that the runnable loop can return if not required.

brendandburns:
Thanks for the explanation.

/lgtm
/approve

PR #1345, opened 2020-10-27
Feat: Adding lease lock implementation

yue9944882:


brendandburns:
/lgtm
/approve

One question: Should the lock implementation be thread-safe? Because right now I don't think it is. Not sure what the programmer expectations are wrt this.

PR #1352, opened 2020-10-29
unable to use autowired in the reconciler

### LINKED ISSUE #1351: unable to use autowired in the reconciler
as the @org.springframework.beans.factory.annotation.Autowired is not supported in BeanPostProcessor or BeanFactoryPostProcessor ([link](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/beans/factory/annotation/Autowired.html)), 

The reconcilers created for KubernetesReconcilerProcessor (base on BeanFactoryPostProcessor) will be autowired either.
@Value won't work for the same reason

So create a spring FactoryBean for Contoller can fix this problem and can also make Reconciler class as a @Component

mingjie-li:
fix #1351


examples/src/main/java/io/kubernetes/client/examples/spring/SpringControllerExample.java
```diff
@@ -145,8 +145,9 @@ public NodePrintingReconciler(
 
     // *OPTIONAL*
     // If you feed like hold the controller from running util some condition..
+    // need to be a public function
     @KubernetesReconcilerReadyFunc
-    boolean informerReady() {
+    public boolean informerReady() {
```
yue9944882, line 150: thanks!

examples/src/main/java/io/kubernetes/client/examples/springfactorybean/SpringControllerFactoryBeanExample.java
```diff
@@ -0,0 +1,159 @@
...
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.CommandLineRunner;
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.stereotype.Component;
+
+@SpringBootApplication
+public class SpringControllerFactoryBeanExample {
```
yue9944882, line 44: can you replace the old example with this new one? the two examples are pretty identical so i think we can just overwrite the whole old spring example w/o starting a new package.

yue9944882 (commented):
one nit otherwise LGTM

examples/src/main/java/io/kubernetes/client/examples/springfactorybean/SpringControllerFactoryBeanExample.java
mingjie-li: done,
could you tell me when will be the next release?

yue9944882: mm i will defer the answer to @brendandburns, but i suppose we're getting close b/c another new release of kubernetes upstream is comming very soon

yue9944882 (commented):
/lgtm

PR #1360, opened 2020-11-03
Feat: Controller metrics

yue9944882:


brendandburns:
Looks good to me, but looks like the pom file is broken?

yue9944882:
@brendandburns fixed pom, plz take another look

PR #1373, opened 2020-11-11
Feat: Utilities for doing client-side load-balancing based-on endpoints

yue9944882:


extended/src/main/java/io/kubernetes/client/extended/network/RoundRobinLoadBalanceStrategy.java
```diff
@@ -0,0 +1,38 @@
...
+  private final ReentrantLock reentrantLock = new ReentrantLock();
+
+  private int robinIndex = 0;
+
+  @Override
+  public String chooseIP(List<String> availableIPs) {
+    if (availableIPs == null || availableIPs.size() == 0) {
+      throw new IllegalArgumentException("failed choosing IP target: empty candidates");
+    }
+    reentrantLock.lock();
```
brendandburns, line 29: Would it be better to use `AtomicInteger` this lock seems heavy-weight.

brendandburns: Also, though this is "round robin" I think you could actually just choose randomly and you'd get the same behavior, I don't think that round robin requires a deterministic order.

brendandburns:
A few comments on the round robin implementation. Also, probably worth adding a 'todo' to add a supplier that uses an Informer instead of polling.

extended/src/main/java/io/kubernetes/client/extended/network/RoundRobinLoadBalanceStrategy.java
yue9944882: > Would it be better to use AtomicInteger this lock seems heavy-weight.

updated

> Also, though this is "round robin" I think you could actually just choose randomly and you'd get the same behavior, I don't think that round robin requires a deterministic order.

i presume the randomized strategy you're saying is a bit different from the proposing one here. for the cases where the client is raising requests at a lower rate e.g. "watching" requests in one loop. the proposed "round robin" in this pull will make sure the even distribution in order, while if we're doing a randomized "round robin", there's a chance of uneven distribution for lower-rate client, is it?

yue9944882:
> A few comments on the round robin implementation. Also, probably worth adding a 'todo' to add a supplier that uses an Informer instead of polling.

i added a new static helper to adapt the informer into the loadbalancer, PTAL

PR #1382, opened 2020-11-14
fix a simple para error in an example

CaiyuhangC:
listPodForAllNamespaces in the base example should have 9 paras but has 10 paras.


brendandburns:
This PR doesn't compile correctly at HEAD.

The parameter was added in the recent regeneration of the API. The basic problem is that HEAD represents the next version, so our examples can be out of date relative to the current version of the API.

We should probably version our examples so that people know how to use a particular version of the library.

I filed #1383 to track this.

But this PR can't merge, so I'm closing.

PR #1407, opened 2020-12-01
Add source jars to deployment

### LINKED ISSUE #1406: Source jars not generated for snapshots locally
I can see there is some confusion about source jars (#598 #597), but there are no source jars available locally when I build this project from source. I am confused because I see source jars in Maven Central (for releases), but no sign of how they got there. I also don't know if there is a snapshot repository (I see no sign of it in the poms) where I can get them if I don't want to build locally. Can we add the source plugin to the build, maybe?

dsyer:
Fixes #1406

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -391,7 +391,21 @@
       </plugins>
     </pluginManagement>
     <plugins>
-      <plugin>
+    <!-- Sources jar -->
+    <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-source-plugin</artifactId>
+        <version>3.2.1</version>
+        <executions>
+            <execution>
+                <id>attach-sources</id>
+                <goals>
+                    <goal>jar-no-fork</goal>
+                </goals>
+            </execution>
+        </executions>
+    </plugin>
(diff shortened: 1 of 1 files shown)


dsyer:
/assign @brendandburns

I'm having issues getting the bot to recognise the CLA. I'll see if there's anything I can do this end.

dsyer:
> This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository.

I don't event know what that means. Maybe I'll just push the code again? Help?

brendandburns:
The commit stuff seems to have sorted itself out, but how is this different than the previous mention of the same plugin:
https://github.com/kubernetes-client/java/blob/master/pom.xml#L345

I'll freely admit to not being an expert in pom files.

yue9944882 (commented):
/lgtm

dsyer:
>  how is this different than the previous mention of the same plugin

In 2 ways: 1. the previous mention is in the `<pluginManagement/>` section so it only governs the default settings when a plugin is actually invoked, 2. it doesn't declare any goals, so it wouldn't do anything (probably).

PR #1410, opened 2020-12-03
Feat: From yaml injection

yue9944882:
simplies resource loading in spring java.

```java
public class Application {
    // #1 injecting fields
    @FromYaml("my-service.yaml") // will look for the yaml file 1. on the machine 2. in the classpath
    private V1Service service;

    @Configuration
    static class MyConfiguration {
       // #2 injecting beans
       @FromYaml("my-service.yaml")
       public V1Service service() {
           return new V1Service();
       }
    }
}
```


yue9944882:
@ryanjbaxter i suppose this will be also helpful for simplifying the integration tests on the other project.

ryanjbaxter:
@yue9944882 for sure

spring/src/main/java/io/kubernetes/client/spring/extended/manifests/KubernetesFromYamlProcessor.java
```diff
@@ -0,0 +1,111 @@
...
+        throw new BeanCreationException(
+            "No such file " + targetFilePath + " either on the machine or classpaths");
+      }
+    }
+
+    try {
+      String yamlContent = new String(Files.readAllBytes(targetPath));
+      Object loadedObj = Yaml.load(yamlContent);
+      return loadedObj;
+    } catch (IOException e) {
```
brendandburns, line 106: Should we catch a parse error here too?

brendandburns:
Cool stuff, one question about error handling.

spring/src/main/java/io/kubernetes/client/spring/extended/manifests/KubernetesFromYamlProcessor.java
yue9944882: added catching `YAMLException`, is that right?

PR #1417, opened 2020-12-04
Factor out usage of Guava in shared libraries

### LINKED ISSUE #1412: Use JDK instead of Guava
Guava is a poor choice of a dependency for a library jar because it changes a lot and binary incompatibilities arise frequently downstream. It isn't used all that much in this project so it should be easy to get rid of. There was a comment in #1261 to this effect, but perhaps not enough focus for anyone to take any action.

Example: `io.kubernetes.client.util.ModelMapper` uses a mixture of utilities from Guava and Apache Commons Lang3. Probably it could at least consolidate to only use the Apache library. Ideally it should be possible to settle on just using the JDK.

dsyer:
A few utilities have their own shared implementations
in util package (e.g. Strings, Streams).

ThreadFactory implemented as private method where needed.

Use ConcurrentHashMap instead of Stripe.

The ClassPath scan in ModelMapper is replaced with private
method.

Caffeine is used for caches.

Fixes #1412

### DIFF
--- a/examples/pom.xml
+++ b/examples/pom.xml
@@ -10,14 +10,14 @@
 		<relativePath>../pom.xml</relativePath>
 	</parent>
 
-    <version>1.0.0-SNAPSHOT</version>
+	<version>1.0.0-SNAPSHOT</version>
 
 	<artifactId>client-java-examples-parent</artifactId>
 	<packaging>pom</packaging>
 	<name>client-java-examples-parent</name>
 
 	<properties>
-        <guava.version>30.0-jre</guava.version>
+		<guava.version>30.0-jre</guava.version>
 		<kubernetes.client.version>10.0.1-SNAPSHOT</kubernetes.client.version>
 	</properties>
 
@@ -26,15 +26,15 @@
 		<module>examples-release-10</module>
 	</modules>
 
-    <dependencyManagement>
-        <dependencies>
-            <dependency>
-                <groupId>com.google.guava</groupId>
-                <artifactId>guava</artifactId>
-                <version>${guava.version}</version>
-            </dependency>
-        </dependencies>
-    </dependencyManagement>
+	<dependencyManagement>
+		<dependencies>
+			<dependency>
+				<groupId>com.google.guava</groupId>
+				<artifactId>guava</artifactId>
+				<version>${guava.version}</version>
+			</dependency>
+		</dependencies>
+	</dependencyManagement>
 
 	<build>
 		<plugins>
--- a/examples/pom.xml
+++ b/examples/pom.xml
@@ -17,6 +17,7 @@
 	<name>client-java-examples-parent</name>
 
 	<properties>
+        <guava.version>30.0-jre</guava.version>
 		<kubernetes.client.version>10.0.1-SNAPSHOT</kubernetes.client.version>
 	</properties>
 
@@ -24,6 +25,16 @@
 		<module>examples-release-11</module>
 		<module>examples-release-10</module>
 	</modules>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>com.google.guava</groupId>
+                <artifactId>guava</artifactId>
+                <version>${guava.version}</version>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
 
 	<build>
 		<plugins>
--- a/extended/src/main/java/io/kubernetes/client/extended/controller/Controllers.java
+++ b/extended/src/main/java/io/kubernetes/client/extended/controller/Controllers.java
@@ -12,14 +12,17 @@ limitations under the License.
 */
 package io.kubernetes.client.extended.controller;
 
-import com.google.common.util.concurrent.ThreadFactoryBuilder;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Function;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import io.kubernetes.client.common.KubernetesObject;
 import io.kubernetes.client.extended.controller.reconciler.Request;
 import io.kubernetes.client.openapi.models.V1ObjectMeta;
-import java.util.concurrent.ThreadFactory;
-import java.util.function.Function;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /** The Controllers is a set of commonly used utility functions for constructing controller. */
 public class Controllers {
@@ -50,6 +53,16 @@ public class Controllers {
    * @return the thread factory
    */
   public static ThreadFactory namedControllerThreadFactory(String controllerName) {
-    return new ThreadFactoryBuilder().setNameFormat(controllerName + "-%d").build();
+    final ThreadFactory defaultFactory = Executors.defaultThreadFactory();
+    final AtomicInteger threadNumber = new AtomicInteger(1);
+    String format = controllerName + "-%d";
+    return r -> {
+      Thread thread = defaultFactory.newThread(r);
+      if (!thread.isDaemon()) {
+        thread.setDaemon(true);
+      }
+      thread.setName(String.format(format, threadNumber.getAndIncrement()));
+      return thread;
+    };
   }
 }
--- a/extended/src/main/java/io/kubernetes/client/extended/controller/Controllers.java
+++ b/extended/src/main/java/io/kubernetes/client/extended/controller/Controllers.java
@@ -16,10 +16,7 @@ import io.kubernetes.client.common.KubernetesObject;
 import io.kubernetes.client.extended.controller.reconciler.Request;
 import io.kubernetes.client.openapi.models.V1ObjectMeta;
 import io.kubernetes.client.util.Threads;
-
-import java.util.concurrent.Executors;
 import java.util.concurrent.ThreadFactory;
-import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Function;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
--- a/extended/src/main/java/io/kubernetes/client/extended/controller/Controllers.java
+++ b/extended/src/main/java/io/kubernetes/client/extended/controller/Controllers.java
@@ -12,17 +12,15 @@ limitations under the License.
 */
 package io.kubernetes.client.extended.controller;
 
+import io.kubernetes.client.common.KubernetesObject;
+import io.kubernetes.client.extended.controller.reconciler.Request;
+import io.kubernetes.client.openapi.models.V1ObjectMeta;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Function;
-
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import io.kubernetes.client.common.KubernetesObject;
-import io.kubernetes.client.extended.controller.reconciler.Request;
-import io.kubernetes.client.openapi.models.V1ObjectMeta;
 
 /** The Controllers is a set of commonly used utility functions for constructing controller. */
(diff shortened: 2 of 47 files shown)


examples/pom.xml
```diff
@@ -17,6 +17,7 @@
 	<name>client-java-examples-parent</name>
 
 	<properties>
+        <guava.version>30.0-jre</guava.version>
```
yue9944882, line 20: nit: fix the indent to align with the line below

pom.xml
```diff
@@ -503,5 +504,51 @@ limitations under the License.
...
       </build>
     </profile>
+    <profile>
+      <id>imports</id>
+      <activation>
+        <os>
+          <family>!windows</family>
+        </os>
+        <file>
+			<missing>.google</missing>
```
yue9944882, line 514: nit: something is wrong with the indents here, is it?

yue9944882 (commented):
@dsyer thanks, most changes are trivial. 

a few nits otherwise LGTM. btw why are the following empty files added?

> examples/examples-release-11/.google 
> examples/examples-release-10/.google

examples/pom.xml
dsyer: The samples use Guava (could also be removed later). These files mark them as "safe" so that the build doesn't fail when `com.google.common` is detected.

yue9944882 (commented):
/lgtm

the changes are backward-compatible overall

yue9944882:
@brendandburns for review and approval

extended/src/main/java/io/kubernetes/client/extended/controller/Controllers.java
```diff
@@ -50,6 +51,16 @@
    * @return the thread factory
    */
   public static ThreadFactory namedControllerThreadFactory(String controllerName) {
-    return new ThreadFactoryBuilder().setNameFormat(controllerName + "-%d").build();
+    final ThreadFactory defaultFactory = Executors.defaultThreadFactory();
```
brendandburns, line 54: Please separate out this change as it is unrelated to removing Guava.

extended/src/main/java/io/kubernetes/client/extended/leaderelection/LeaderElector.java
```diff
@@ -334,8 +343,17 @@ private void maybeReportTransition() {
     }
   }
 
-  private ThreadFactory makeThreadFactory(String nameFormat) {
-    return new ThreadFactoryBuilder().setDaemon(true).setNameFormat(nameFormat).build();
+  private ThreadFactory makeThreadFactory(String format) {
```
brendandburns, line 346: Please put this change in a different PR (probably the same PR as the above)

pom.xml
```diff
@@ -503,5 +504,51 @@ limitations under the License.
         </plugins>
       </build>
     </profile>
+    <profile>
```
brendandburns, line 507: Please separate this out into a separate PR

In general we're using Spotless for formatting, and I'm not sure we want to introduce another formatting step.

util/src/main/java/io/kubernetes/client/apimachinery/KubernetesRequestDigest.java
```diff
@@ -102,7 +104,8 @@ public String toString() {
     }
 
     String groupVersion;
-    if (Strings.isNullOrEmpty(resourceMeta.getGroupVersionResource().getGroup())) { // core resource
+    if (Strings.isNullOrEmpty(resourceMeta.getGroupVersionResource().getGroup())) { // core
+      // resource
```
brendandburns, line 108: this break is awkward and weird, leave it on the previous line?

util/src/main/java/io/kubernetes/client/informer/cache/Controller.java
```diff
@@ -78,16 +83,25 @@ public Controller(
...
     this.resyncExecutor =
         Executors.newSingleThreadScheduledExecutor(
-            new ThreadFactoryBuilder()
-                .setNameFormat("controller-resync-" + apiTypeClass.getName() + "-%d")
-                .build());
+            threadFactory("controller-reflector-" + apiTypeClass.getName() + "-%d"));
+  }
+
+  private ThreadFactory threadFactory(String format) {
+    final ThreadFactory defaultFactory = Executors.defaultThreadFactory();
```
brendandburns, line 95: same comment as above.

util/src/main/java/io/kubernetes/client/util/ModelMapper.java
```diff
@@ -65,8 +68,9 @@
 
   private static Set<Discovery.APIResource> lastAPIDiscovery = new HashSet<>();
 
-  private static volatile long nextDiscoveryRefreshTimeMillis =
-      0; // zero means api-discovery never done
+  private static volatile long nextDiscoveryRefreshTimeMillis = 0; // zero means
+  // api-discovery
```
brendandburns, line 72: this comment breaking is awkward.

util/src/main/java/io/kubernetes/client/util/ModelMapper.java
```diff
@@ -368,43 +378,63 @@ private static void initModelMap() throws IOException {
...
 
   static class BiDirectionalMap<K, V> {
-    private Map<K, V> kvMap = new HashMap<>();
-    private Map<V, K> vkMap = new HashMap<>();
-    private Striped<ReadWriteLock> lock =
-        Striped.readWriteLock(Runtime.getRuntime().availableProcessors() * 4);
+
+    private Map<K, V> kvMap = new ConcurrentHashMap<>();
+
+    private Map<V, K> vkMap = new ConcurrentHashMap<>();
```
brendandburns, line 384: What is the locking in here? We used Striped so that there would be multiple locks to reduce contention, I'd rather not go back to a single lock.

brendandburns:
A few comments. I realized after the fact that it's possible that the thread factory changes are because you're removing the Guava thread factory builder and re-implementing what it does.

I still think it's worth separating out those changes into a different PR (esp. because the same code is repeated several times) also, it would be nice to have a pointer to the Guava implementation so that we can validate that the behavior is the same.

Thanks!

dsyer:
The thread factory changes are precisely because we are removing Guava. The original implementation is here: https://github.com/google/guava/blob/master/guava/src/com/google/common/util/concurrent/ThreadFactoryBuilder.java

It's quite hard to separate those changes out, so I'd rather not, unless you have some specific reason to do it. I don't have any problem with tidying up or centralizing the thread factory patterns into a utility if you prefer. That could also happen as a separate PR afterwards.

I'll tidy up the line breaks - it's because my IDE is set to wrap at 80 characters.

util/src/main/java/io/kubernetes/client/util/ModelMapper.java
dsyer: `ConcurrentHashMap` has striped locks internally. I didn't see any tests (I know they are hard to write) or comments about the implementation, so I'm not sure what the race condition is that we are trying to avoid. Can you elaborate, and maybe we can improve it?

extended/src/main/java/io/kubernetes/client/extended/controller/Controllers.java
```diff
@@ -50,6 +51,16 @@
...
+    final AtomicInteger threadNumber = new AtomicInteger(1);
+    String format = controllerName + "-%d";
+    return r -> {
+      Thread thread = defaultFactory.newThread(r);
+      if (!thread.isDaemon()) {
```
brendandburns, line 59: Looking at the thread builder code, I don't think that we were previously setting threads to daemon.

extended/src/main/java/io/kubernetes/client/extended/leaderelection/LeaderElector.java
```diff
@@ -334,8 +343,17 @@ private void maybeReportTransition() {
...
+    final ThreadFactory defaultFactory = Executors.defaultThreadFactory();
+    final AtomicInteger threadNumber = new AtomicInteger(1);
```
brendandburns, line 348: This code is repeated, please factor out into a single method that is re-used in various places.

util/src/main/java/io/kubernetes/client/informer/cache/Controller.java
```diff
@@ -78,16 +83,25 @@ public Controller(
...
+    final AtomicInteger threadNumber = new AtomicInteger(1);
```
brendandburns, line 96: here too wrt refactoring to a shared method

util/src/main/java/io/kubernetes/client/util/ModelMapper.java
brendandburns: ok, I'm happy if it has striped locks

There's no race condition to avoid, simply that the bi-directional map is used to store API -> type mappings that are used in a number of different places, and we want to reduce contention on that lookup if there are multiple threads all performing lookups at the same time.

brendandburns:
Thanks for the updates!

One comment about the thread behavior changing to always make them daemon threads (I think). We can discuss if we want that change, but it should be a different PR.

One request for a refactor to a shared implementation of thread creation.

Then I think this is read to merge.

extended/src/main/java/io/kubernetes/client/extended/controller/Controllers.java
dsyer: Guava was (it's the default). I can't think of any reason you'd want these to be non-daemon.

extended/src/main/java/io/kubernetes/client/extended/leaderelection/LeaderElector.java
dsyer: It's not repeated in this class. Do you mean you want to have a utility class that is shared across the Java SDK?

dsyer: Did that.

yue9944882 (commented):
/lgtm

extended/src/main/java/io/kubernetes/client/extended/controller/Controllers.java
dsyer: I switched it back to non-daemon anyway. I don't know why they weren't daemon before but it does seem to work, so it's safer to leave it as is.

pom.xml
brendandburns: Can you comment on how this is resolved? I don't think these changes are related to Guava.

brendandburns:
Basically LGTM, but I noticed there is one `pom.xml` (formatter?) change that I wanted clarification on.

Thanks!

yue9944882:
> Basically LGTM, but I noticed there is one pom.xml (formatter?) change that I wanted clarification on.

the new plugin forbids guava imports which makes sure we dont step backwards @brendandburns

brendandburns:
Ok, I would rather we did that using a spotless regex (as we did with wildcards) rather than a whole new exec/egrep call.

dsyer:
Do you have an example of how to do that with spotless? I find the user docs a bit thin there.

yue9944882:
@brendandburns i think we can resolve that as a part of https://github.com/kubernetes-client/java/pull/1425, so that we can get this in first. wdyt?

dsyer:
I looked at #1425 for inspiration and I can see what you are doing, but it makes things complicated for now because Guava is still allowed in tests (it ships with Wiremock) and in examples. I'd be happy to add another PR after this one replacing the exec plugin call with another section of spotless config. It won't result in a smaller pom.xml (I predict) but we can see.

brendandburns:
I'd rather just rip out the exec/formatting changes from this PR, and then merge the spotless changes in a follow on PR. I think we can be careful for a few days about re-introducing dependencies.

I don't think the wiremock dependency matters for formatting, since it won't result in any import statements.

And I'm comfortable removing Guava from the examples (I don't think it's super necessary in any of them)

dsyer:
> I'd rather just rip out the exec/formatting changes from this PR

OK. Done. We can deal with the examples and tests when we do the spotless configuration I guess (i.e. another PR)?

brendandburns:
/lgtm
/approve

Thanks for the patience!

PR #1418, opened 2020-12-04
Switch to JDK date time library

### LINKED ISSUE #1411: Use JDK for date time not Joda
Joda is not a good choice for a modern library. I doubt if anyone who uses the Kubernetes SDK needs it. Unfortunately it's not easy to get rid of because some of the generated code uses it. You can change the options in the code generator (https://openapi-generator.tech/docs/generators/java/) so that's what we should do I believe, but that involves changes to the "gen" project in kubernetes-client to be co-ordinated with this project. I don't even know how to start that process, so someone will have to make a suggestion.

dsyer:
See https://github.com/kubernetes-client/gen/pull/177 for
the corresponding change in the "gen" project.

Fixes #1411 (but N.B. this is a binary incompatible change).


yue9944882 (commented):
mmmm, all the tests pass so there's no problem talking to an e2e kubernetes cluster in terms of [de]serialization. 

if we're merging this, i think we should provide utilities for jodatime <-> offsetdatetime conversion so that users can bandage the incompatible changes. @brendandburns @dsyer wdyt?

dsyer:
What kind of utilities were you thinking of? It seems like we encourage users to generate their API objects from YAML? If so then the conversion should be automatic (just upgrade the libraries and re-generate your model objects). I suppose there might be some model objects that need to be manipulated in user code, and they might contain date time fields. The conversion of the code in this project was largely trivial, by the way (Java 8 date time library evolved from joda-time so that's not surprising), and I expect users would experience the same level of toil. It is possible to automate such tasks. Not sure if it's worth it.

yue9944882:
for clarification, i was thinking about something like the code snippet below. and the point is to help users do less work in the migration. e.g. users can be already using joda-time in the other non-kubernetes models in their legacy system and they don't have to switch the presence of joda-time everywhere. just think about the kids :)

```java
public class DateTimeUtilities {
   public static toOffsetDateTime(JodaTime time) {...}
   public static toJodaTime(OffsetDateTime time) {...}
}
```

brendandburns:
Agree, this is a breaking change for users of this library and so I'd like to make the transition easier via some utilities.

yue9944882:
@dsyer how do you think about reviving the thread?

dsyer:
If you think it would work as a new major version I can rebase onto master I guess. I’m on PTO till Monday.

yue9944882:
>If you think it would work as a new major version I can rebase onto master I guess. I’m on PTO till Monday.

thanks, we're hitting an issue that joda-time doesn't support micro/nano sec timestamp, replacing it in the next release sounds good to me. cc @brendandburns

yue9944882:
@dsyer btw can you re-generate together w/ 1.20 kubernetes? (by bumping source branch name to `release-1.20`)

https://github.com/kubernetes-client/java/blob/a8e52f7db545eea51b024e812024613b6afe8556/settings#L18

dsyer:
I rebased. Also added a further commit with the k8s version change (so you can leave that off if you want).

dsyer:
FWIW the `LeaderElectorTest` works for me locally. The `CSRTest` does not but I don't really know why and that one seems to work in CI anyway. Seems like the tests might be flickering?

brendandburns:
I bounced the tests, we'll see if there was a flake.

dsyer:
That didn't help then. The CI failures are different to my local failures still. :woman_shrugging: 

Not sure how to debug that.

brendandburns:
Chances are the test is racy. Usually the best way to repro that I have found is to run the test on a really, really low powered cloud VM (e.g. Azure B-series or AWS *.micro series)

brendandburns:
I actually see LeaderElectorTest failing for me locally (timeout) at HEAD, I will attempt to debug on my local machine.

brendandburns:
@dsyer have you rebased this PR to HEAD?

dsyer:
I rebased yesterday. HEAD has moved of course, but that shouldn't affect the tests right?

kubernetes/src/main/java/io/kubernetes/client/openapi/JSON.java
```diff
@@ -17,41 +17,36 @@
...
-  private DateTimeTypeAdapter dateTimeTypeAdapter = new DateTimeTypeAdapter();
+  private OffsetDateTimeTypeAdapter offsetDateTimeTypeAdapter = new OffsetDateTimeTypeAdapter();
   private LocalDateTypeAdapter localDateTypeAdapter = new LocalDateTypeAdapter();
   private ByteArrayAdapter byteArrayAdapter = new ByteArrayAdapter();
 
   public static GsonBuilder createGson() {
     GsonFireBuilder fireBuilder = new GsonFireBuilder();
-    GsonBuilder builder =
-        fireBuilder
-            .registerPreProcessor(V1Status.class, new V1StatusPreProcessor())
```
yue9944882, line 53: @dsyer removal of this line leads to the e2e failure. let's get it back.

yue9944882: did you update the openapi-ignore file before re-generating the project?

https://github.com/kubernetes-client/java/blob/53114b0c517c3b47f865e06d9f18ff6b55f30cc4/kubernetes/.openapi-generator-ignore#L10

dsyer: Ah, well spotted. I didn't see that one. It's a manual change to a generated source. Would be nice to automate it at some point.

dsyer: It looks like it didn't fix the e2e tests though. So more :shrug: .

yue9944882 (commented):
@dsyer thanks! the pull looks good now, the remaining e2e failure is caused by unwounded hacks for joda-time formatter which i suppose is irrelavant to this pull. i will deal w/ the e2e failure in a follow up.

https://github.com/kubernetes-client/java/blob/master/e2e/src/test/java/io/kubernetes/client/e2e/extended/leaderelection/LeaderElectorTest.java#L88-L101

dsyer:
Those changes in `LeaderElectorTest` shouldn't be necessary with JDK time. `OffsetDateTime.parse()` works with "2020-01-04T04:34:56Z" or "2020-01-04T04:34:56.001Z" or "2020-01-04T04:34:56.001002Z".

PR #1419, opened 2020-12-06
Add a KubectlApply processor for Spring.

brendandburns:
KubectlCreate is good if things are constant, but if resources could have changed, it's useful to perform an apply instead.

Note that this is largely, copy/paste from KubectlCreate, we could consider extracting a super-class for shared methods if we wanted to.

### DIFF
--- a/spring/src/main/java/io/kubernetes/client/spring/extended/manifests/KubernetesKubectlApplyProcessor.java
+++ b/spring/src/main/java/io/kubernetes/client/spring/extended/manifests/KubernetesKubectlApplyProcessor.java
@@ -1,0 +1,72 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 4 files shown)


spring/src/main/java/io/kubernetes/client/spring/extended/manifests/KubernetesKubectlApplyProcessor.java
```diff
@@ -0,0 +1,72 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.BeansException;
+import org.springframework.beans.factory.BeanCreationException;
+import org.springframework.beans.factory.BeanFactory;
+import org.springframework.beans.factory.BeanFactoryAware;
+import org.springframework.beans.factory.ListableBeanFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.config.BeanPostProcessor;
+
+public class KubernetesKubectlApplyProcessor implements BeanPostProcessor, BeanFactoryAware {
```
yue9944882, line 30: i suppose we can share some code between the apply and the create processor 🤔 we can do that in a follow-up refactoring pull

yue9944882 (commented):
/lgtm

PR #1423, opened 2020-12-08
Make sundr codegen optional so it only applies to the k8s API code

### LINKED ISSUE #1422: Dependecy hygiene: sundr-codegen should be optional
As was noted here, the sundr code generation processor pollutes the classpath of downstream projects (including, ironically, the "extended" jar in this project, which is using commons-lang classes from the sundr jar): https://github.com/sundrio/sundrio/issues/153. Sundr really shouldn't be shading 3rd party libraries without repackaging, but it's kind of a nightmare for users of the k8s client jars, and the dependency almost certainly ought to be optional here anyway.

dsyer:
Fixes #1422

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 .settings/
 .project
 .vscode/
+.attach_pid*
 
 # Maven builds
 */target/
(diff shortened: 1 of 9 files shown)


yue9944882 (commented):
/lgtm

PR #1425, opened 2020-12-08
Breaking sources when wildcard or guava imports present

yue9944882:


pom.xml
```diff
@@ -453,11 +453,16 @@
               </includes>
               <trimTrailingWhitespace />
               <endWithNewline />
+              <replaceRegex>
+                <name>Remove wildcard imports</name>
+                <searchRegex>import (static )?\s+\S+\.\*;</searchRegex>
```
brendandburns, line 458: change this to `import (static)?\s+\S+\.\*;(\n);` to pick up the newline

pom.xml
```diff
@@ -453,11 +453,16 @@
...
+                <replacement>INVALID IMPORTS (WILDCARD)</replacement>
```
brendandburns, line 459: Change this to `\n`

brendandburns:
Let's change this so that it just replaces wildcards with a newline. Otherwise LGTM.

pom.xml
brendandburns: I think `$` represents end of expression rather than end of line (`\n` or '\r\n`)

yue9944882: @brendandburns i tested locally and it seems that the plugin matches the source files by lines so `$` is matching EOL. e.g. adding an `import com.google.guava.*;` line in the middle of the file will be captured.

PR #1430, opened 2020-12-10
Prevents event recorder worker thread leakage

yue9944882:
we found worker leakage in some environments, this pull prevents potential leakage


brendandburns:
/lgtm
/approve

Should we look for this in other threads we create?

PR #1438, opened 2020-12-14
Fix Guava regexp in spotless config

dsyer:
There was a regex in the spotless config in top level pom.xml designed to assert that Guava was not being used, but the pattern was wrong. Once you fix that, you either need to rewrite some of the tests that were still using it, or don't apply that rule to tests.


yue9944882 (commented):
@dsyer lgtm, can you squash into one commit?

dsyer:
> can you squash into one commit?

Done.

PR #1445, opened 2020-12-16
Workaround for gson-syntax-exception upon deletion even if the deletion succeeds

yue9944882:
Fixes: https://github.com/kubernetes-client/java/issues/86

this pull will be suppressing the GsonSyntaxException on deletion by adapting non-status resources into status during deserialization.

@brendandburns i think we should get this into 11.0.x releases.

### DIFF
--- a/e2e/src/test/groovy/io/kubernetes/client/e2e/basic/CoreV1ApiTest.groovy
+++ b/e2e/src/test/groovy/io/kubernetes/client/e2e/basic/CoreV1ApiTest.groovy
@@ -1,0 +1,41 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


brendandburns:
Do we really want to do this? Doesn't this make every delete look like it's not done yet?

yue9944882:
> Doesn't this make every delete look like it's not done yet?

@brendandburns upon deletion calls, the kube-apiserver can return either a status object or the deleted state of the target object. however the method generated from openapi is only returning the status object.

the only change this pull will bring to the project is making gson suppressing the runtime exception when deserializing a non-status object (e.g. deployment, pod) into a `V1Status` which is happening when we successfully did a background deletion. this is actually done by casting/transforming the deleted object into a status object by removing "status" fields in order to solve the gson exception --- there's still information loss but it's way better than throwing a gson runtime exception surprisingly even if the deletion succeeded..

brendandburns:
Ok, I went and dug up the actual code in the apiserver:

https://github.com/kubernetes/apiserver/blob/93ffe41e75d9d1c6cfe6d9224f8f5f474606b5b2/pkg/endpoints/handlers/delete.go#L144

I had thought that it returned V1Status when the delete wasn't completed, and the object when the delete was completed, but it looks like it's actually dependent on the deleter.

Given that, I think that this change is ok, even though it is a little lossy.

Let's file an issue to fix this better via oneOf instead of this work-around, so that we don't forget about the problem.

/lgtm
/approve

PR #1453, opened 2020-12-21
Release 11.0.0

brendandburns:
Ok, so this is the real `11.0.0` release.

To get it to work, I had to remove all of the examples modules from the examples `pom.xml`

I will come along and do a follow up PR to re-add the latest ones.

We also need to figure out a better way to make this automatic. The release plugin is overwriting versions and messing things up. We need to block it from writing versions in some of the poms


brendandburns:
@yue9944882 friendly ping on this one.

yue9944882 (commented):
/lgtm

PR #1460, opened 2020-12-26
Leader to give up leadership lock on graceful shutdown

himanshug:
So that other candidate can become the next leader quickly or else other candidate might need to wait for full `leaseDuration` before getting the lock.

### DIFF
--- a/extended/src/test/java/io/kubernetes/client/extended/leaderelection/LeaderElectorTest.java
+++ b/extended/src/test/java/io/kubernetes/client/extended/leaderelection/LeaderElectorTest.java
@@ -1,0 +1,82 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package io.kubernetes.client.extended.leaderelection;
+
+import org.junit.Test;
+
+import java.time.Duration;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Leader Election tests using "simulated" locks created by {@link LockSmith}
+ */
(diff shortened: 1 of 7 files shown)


extended/src/test/java/io/kubernetes/client/extended/leaderelection/LeaderElectorTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+/** Leader Election tests using "simulated" locks created by {@link LockSmith} */
+public class LeaderElectorTest {
+  /**
+   * Tests that when a leader candidate is stopped gracefully, second candidate immediately becomes
+   * leader.
+   */
+  @Test(timeout = 20000L)
+  public void testLeaderGracefulShutdown() throws Exception {
+    LockSmith lockSmith = new LockSmith();
```
brendandburns, line 28: I feel like there is value in using WireMock and actually talking HTTP for the locks. How hard is it to switch to testing using WireMock like we do elsewhere?

brendandburns:
Thanks, this looks good to me, but I'd rather see the tests actually use HTTP rather than an in-memory mock.

Also, I'd like @yue9944882 to take a look at this before we merge it, since he wrote most of this code.

extended/src/test/java/io/kubernetes/client/extended/leaderelection/LeaderElectorTest.java
himanshug: Thanks for the review. 

Here we are more interested in testing the code of LeaderElector which can be tested as long as a well-behaving `Lock` implementation is provided. Also, truth is that it was just super simple/quick to do and does the job :) ... and followed in spirit the existing unit tests for leader election related code.

With WireMock, we would actually also test the implementation of Lock itself i.e. ConfigMapLock class etc. I haven't used WireMock in the past, so not sure about the work required, but will try to look into it.

that said, not sure but, I see there is an e2e build that appears to setup a real k8s cluster, how would you feel about this test running as an e2e integration test against real k8s cluster instead of the WireMock ?

yue9944882 (commented):
thanks! it's generally a good idea to the perform graceful fore-ground shutdown.

/lgtm

extended/src/test/java/io/kubernetes/client/extended/leaderelection/LeaderElectorTest.java
brendandburns: e2e integration is even better :)

brendandburns:
/approve

Would love to have the integration test as a follow on PR. Thanks!

PR #1466, opened 2021-01-02
minor improvement

wind57:


wind57:
I signed it

util/src/main/java/io/kubernetes/client/informer/EventType.java
```diff
@@ -23,11 +29,15 @@
 
   ERROR;
```
wind57, line 31: Every time you call `values()` on an enum, there is an inner copy created. We can pre-compute the needed `Map` before doing the search as a `static final instance` and simply do a hash lookup. `String::valueOf` is needed only to take care of `null`

wind57:
As a general note, I do not know if these kind of PRs are encouraged, I do understand that every project has its own rules, internally. If they are not, I will close this one and not open new ones. Just let me know. thank you.

yue9944882 (commented):
@wind57 am positive w/ these minor PRs if it has its own technical merit. additionally we always need to take care of compatibility.  and plz squash the commits.

util/src/main/java/io/kubernetes/client/informer/EventType.java
```diff
@@ -23,11 +29,15 @@
...
+
   /**
-   * getByType returns the corresponding EventType by type.
+   * returns the corresponding EventType by type.
    *
    * @param type specific code
    * @return corresponding EventType
+   * @deprecated will be removed in a future release. use : findByType
    */
   public static EventType getByType(String type) {
```
yue9944882, line 42: annotate w/ `@Deprecated`?

wind57: done + squashed. thank you.

brendandburns:
/lgtm
/approve

I'm fine with these small incremental improvements. I don't think we want refactors just for the sake of refactoring, but even minor performance improvements are worthwhile.

Thanks!

wind57:
@brendanburns exactly. this is not about refactoring for the sake of simply changing code - that's not why I made these. thank you for approving. there will be coming.

PR #1467, opened 2021-01-02
clean-up, fix a bug, tests

wind57:


util/src/main/java/io/kubernetes/client/informer/ResyncRunnable.java
```diff
@@ -18,30 +18,26 @@
 import org.slf4j.LoggerFactory;
 
 /**
- * ResyncRunnable class implements Runnable interface. It calls the resync function of Store
```
wind57, line 21: there is no sense in saying that `Store` is getting called, when in reality `Store` is not even passed in the constructor. I guess this is an old left over, when the code moved, but the documentation was left intact

util/src/main/java/io/kubernetes/client/informer/ResyncRunnable.java
```diff
@@ -18,30 +18,26 @@
...
+ * ResyncRunnable class implements Runnable interface. It calls the resync function of {@link
+ * DeltaFIFO}
  */
 public class ResyncRunnable implements Runnable {
 
   private static final Logger log = LoggerFactory.getLogger(ResyncRunnable.class);
 
-  private DeltaFIFO store;
-  private Supplier<Boolean> shouldResyncFunc;
+  private final DeltaFIFO store;
```
wind57, line 28: could be made `final`, of course.

util/src/main/java/io/kubernetes/client/informer/ResyncRunnable.java
```diff
@@ -18,30 +18,26 @@
...
     this.shouldResyncFunc = shouldResyncFunc;
   }
 
   public void run() {
-    if (log.isDebugEnabled()) {
-      log.debug("ResyncRunnable#resync ticker tick");
-    }
+    log.debug("ResyncRunnable#resync ticker tick");
 
-    if (shouldResyncFunc == null || shouldResyncFunc.get()) {
```
wind57, line 41: no need for the `log.isDebugEnabled()` check. this is done internally anyway

util/src/main/java/io/kubernetes/client/informer/ResyncRunnable.java
```diff
@@ -18,30 +18,26 @@
...
-      if (log.isDebugEnabled()) {
-        log.debug("ResyncRunnable#force resync");
-      }
+    if (shouldResyncFunc == null || (shouldResyncFunc.get() != null && shouldResyncFunc.get())) {
```
wind57, line 39: a bug fix in case `() -> null` (in which case this will be treated as `false`)

util/src/test/java/io/kubernetes/client/informer/ResyncRunnableTest.java
```diff
@@ -0,0 +1,58 @@
...
+import static org.mockito.Mockito.verify;
+
+import io.kubernetes.client.informer.cache.DeltaFIFO;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.junit.MockitoJUnitRunner;
+
+/** @author wind57 */
```
wind57, line 24: just some tests that take care of proper flow in logic

brendandburns:
Looks fine to me. @yue9944882 wdyt?

yue9944882 (commented):
/lgtm

PR #1468, opened 2021-01-02
first commit

wind57:


util/src/main/java/io/kubernetes/client/util/Strings.java
```diff
@@ -14,10 +14,14 @@
 
 import javax.annotation.Nullable;
 
-public class Strings {
+public final class Strings {
```
wind57, line 17: 1. utility classes are usually `final` and prohibit instantiation.

util/src/main/java/io/kubernetes/client/util/Strings.java
```diff
@@ -14,10 +14,14 @@
...
+
+  private Strings() {
+    throw new AssertionError("not supported");
+  }
 
   public static boolean isNullOrEmpty(String value) {
-    return value == null || value.length() == 0;
+    return value == null || value.isEmpty();
```
wind57, line 24: 2. same thing, but using a dedicated method. why, though? because `java.lang.String` is highly optimized, you never know when `isEmpty` is going to become an intrinsic. delegating to such methods, is always a good idea.

util/src/main/java/io/kubernetes/client/util/Preconditions.java
```diff
@@ -0,0 +1,45 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package io.kubernetes.client.util;
+
+import java.util.function.Predicate;
+import java.util.function.Supplier;
+
+/**
```
wind57, line 18: 3. creating a dedicated utility for preconditions, to be re-used everywhere where `if(...) {throw new IllegalArgumentException...}`

util/src/test/java/io/kubernetes/client/util/PreconditionsTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+import static io.kubernetes.client.util.Preconditions.precondition;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import org.assertj.core.api.Assertions;
+import org.junit.Test;
+
+public class PreconditionsTest {
+
+  @Test
```
wind57, line 23: 4. test the `Preconditions` class

util/src/main/java/io/kubernetes/client/apimachinery/GroupVersion.java
```diff
@@ -12,24 +12,24 @@
...
 
+import static io.kubernetes.client.util.Preconditions.precondition;
+
 import io.kubernetes.client.common.KubernetesObject;
 import io.kubernetes.client.util.Strings;
 import java.util.Objects;
 
 public class GroupVersion {
 
+  private static final GroupVersion GROUP_VERSION_V1 = new GroupVersion("", "v1");
```
wind57, line 23: 5. it is an immutable class, for `v1` let's return the same instance, instead of creating a new one.

util/src/main/java/io/kubernetes/client/apimachinery/GroupVersion.java
```diff
@@ -12,24 +12,24 @@
...
+
   public static GroupVersion parse(String apiVersion) {
-    if (Strings.isNullOrEmpty(apiVersion)) {
-      throw new IllegalArgumentException("No apiVersion found on object");
-    }
     if ("v1".equals(apiVersion)) {
```
wind57, line 26: 6. invert the checks: first do the `v1.equals(apiVersion)`, _then_ the `precondition` check. `v1.equals...` check is cheaper than the `precondition` one and I also guess far more often used. As such, it makes a lot more sense to invert these.

wind57: I have also slightly edited the error messages, so that they make slightly more sense.

util/src/test/java/io/kubernetes/client/apimachinery/GroupVersionTest.java
```diff
@@ -21,17 +23,37 @@
...
-          GroupVersion.parse(new V1Pod().apiVersion("foo/bar/f"));
-        });
+  public void parseV1() {
+    GroupVersion left = new GroupVersion("", "v1");
+    GroupVersion right = GroupVersion.parse(new V1Pod().apiVersion("v1"));
+    assertThat(left).isEqualTo(right);
+  }
+
+  @Test
+  public void parseAppsV1() {
```
wind57, line 33: 7. use `assertj` in tests and proper separate tests in separate methods.

brendandburns: Is there value in using assertj? I'd prefer not to add anoother dependency. Can we just use JUnit asserts?

util/src/test/java/io/kubernetes/client/util/PreconditionsTest.java
```diff
@@ -0,0 +1,36 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package io.kubernetes.client.util;
+
+import static io.kubernetes.client.util.Preconditions.precondition;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
```
brendandburns, line 16: As above, why not just use junit?

brendandburns: Nevermind, you didn't modify pom.xml and the code compiles, so I guess we must already depend on it.

brendandburns:
Optimizations/changes make sense to me. Thanks for starting to contribute to the project!

/lgtm
/approve

PR #1470, opened 2021-01-04
Add updateStatus() convenience methods to GenericKubernetesApi

### LINKED ISSUE #1442: Implement StatusWriter so you can patch just the status field in an API object
In the Java client is there an equivalent of StatusWriter in Go (where you can update just the status of an API object)? I don't see it anywhere so I guess this is a feature request.

dsyer:
Fixes #1442, as a draft at least. We could refactor it to be more like the Golang API, e.g. 

```java
GenericKubernetesApi<T,L> api = ...;
api.status().update(...);
```

instead of

```java
GenericKubernetesApi<T,L> api = ...;
api.updateStatus(...);
```

WDYT?

### DIFF
--- a/util/src/main/java/io/kubernetes/client/util/generic/GenericKubernetesApi.java
+++ b/util/src/main/java/io/kubernetes/client/util/generic/GenericKubernetesApi.java
@@ -37,6 +37,8 @@ import io.kubernetes.client.util.generic.options.ListOptions;
 import io.kubernetes.client.util.generic.options.PatchOptions;
 import io.kubernetes.client.util.generic.options.UpdateOptions;
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


brendandburns:
/lgtm
/approve

I think that `updateStatus` is more congruent with the other Java APIs. I don't think that direct emulation of all of the client-go calls is necessarily the goal, I'd prefer to make this library as consistent with itself as possible.

karunasagark:
@brendandburns @yue9944882, can we plan to include this in next release? maybe 11.1.0?

yue9944882:
@karunasagark oops, we just had a new 11.0.x release a few hours ago, i suppose we will be getting a next release after getting a few more cherry-picks.

PR #1475, opened 2021-01-05
WIP: Revert changes to the versioning from release plugin.

brendandburns:
The `mvn release` plugin overwrote these versions incorrectly. Reverting that change.

### DIFF
--- a/examples/examples-release-10/pom.xml
+++ b/examples/examples-release-10/pom.xml
@@ -7,7 +7,7 @@
     <name>client-java-examples-release-10</name>
     <version>1.0.0</version>
     <properties>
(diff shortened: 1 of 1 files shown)


dsyer:
I think I can see what you are doing here, but I'm still confused. The versions of these contrib jars in Maven Central are 10.0.0 and 10.0.1. Maybe that was a screw up with the release plugin? But where are the 0.x releases? Even the snapshots as listed in this PR are not installed locally when I run `mvn install` from the top of this project (they are not part of the reactor build). Maybe they need to have their own examples in the contrib dir? Would that be clearer?

PR #1477, opened 2021-01-06
Support rfc3339 micro-sec (6 digit) timestamp

yue9944882:
@brendandburns @dsyer 

it's a follow-up of #1418.

this pull will force the timestamp types to be serialized including 6 digit of fraction in order to support the precision of micro-second.

the original kubernetes go types basically accepts two kinds of timestamp in RFC3339:

1. `RFC3339     = "2006-01-02T15:04:05Z07:00"` for the go type [metav1.Time](https://github.com/kubernetes/kubernetes/blob/39483aa0957be65b04686173d2970b508b3974ca/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/time.go#L31-L33)
2. `RFC3339Micro = "2006-01-02T15:04:05.000000Z07:00"` for the go type [metav1.MicroTime](https://github.com/kubernetes/kubernetes/blob/39483aa0957be65b04686173d2970b508b3974ca/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/micro_time.go#L31-L33)

at the kubernetes server side, (2) can be implicitly casted to (1) if extra precision is provided. ideally java model should discriminate (1) and (2) w/ separate java types so that we can switch the precision dynamically upon serialization. however, that's infeasible for now b/c the published openapi spec from kubernetes server is marking (1) and (2) as the same openapi schema types. 

the bright side is that discriminating time types in different precision w/ different java types can be odd aesthetically..

### DIFF
--- a/extended/src/main/java/io/kubernetes/client/extended/leaderelection/resourcelock/LeaseLock.java
+++ b/extended/src/main/java/io/kubernetes/client/extended/leaderelection/resourcelock/LeaseLock.java
@@ -135,15 +135,23 @@ public class LeaseLock implements Lock {
   }
 
   private V1LeaseSpec getLeaseFromRecord(LeaderElectionRecord record) {
-    return new V1LeaseSpec()
-        .acquireTime(
-            OffsetDateTime.ofInstant(
-                Instant.ofEpochMilli(record.getAcquireTime().getTime()), ZoneOffset.UTC))
-        .renewTime(
-            OffsetDateTime.ofInstant(
-                Instant.ofEpochMilli(record.getRenewTime().getTime()), ZoneOffset.UTC))
-        .holderIdentity(record.getHolderIdentity())
-        .leaseDurationSeconds(record.getLeaseDurationSeconds())
-        .leaseTransitions(record.getLeaderTransitions());
+    V1LeaseSpec spec =
+        new V1LeaseSpec()
+            .holderIdentity(record.getHolderIdentity())
+            .leaseDurationSeconds(record.getLeaseDurationSeconds())
+            .leaseTransitions(record.getLeaderTransitions());
+    if (record.getAcquireTime() != null) {
+      spec =
+          spec.acquireTime(
+              OffsetDateTime.ofInstant(
+                  Instant.ofEpochMilli(record.getAcquireTime().getTime()), ZoneOffset.UTC));
+    }
+    if (record.getRenewTime() != null) {
+      spec =
+          spec.renewTime(
+              OffsetDateTime.ofInstant(
+                  Instant.ofEpochMilli(record.getRenewTime().getTime()), ZoneOffset.UTC));
+    }
+    return spec;
   }
 }
--- a/kubernetes/src/main/java/io/kubernetes/client/openapi/JSON.java
+++ b/kubernetes/src/main/java/io/kubernetes/client/openapi/JSON.java
@@ -45,7 +45,7 @@ public class JSON {
 
   private boolean isLenientOnJson = false;
 
-  public static final DateTimeFormatter RFC3339MICRO_FORMATTER =
+  private static final DateTimeFormatter RFC3339MICRO_FORMATTER =
       new DateTimeFormatterBuilder()
           .parseDefaulting(ChronoField.OFFSET_SECONDS, 0)
           .append(DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss"))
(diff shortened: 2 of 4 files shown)


kubernetes/src/main/java/io/kubernetes/client/openapi/JSON.java
```diff
@@ -42,11 +44,22 @@
...
+  public static final DateTimeFormatter RFC3339MICRO_FORMATTER =
+      new DateTimeFormatterBuilder()
+          .parseDefaulting(ChronoField.OFFSET_SECONDS, 0)
+          .append(DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss"))
+          .optionalStart()
+          .appendFraction(ChronoField.NANO_OF_SECOND, 6, 6, true)
+          .optionalEnd()
+          .appendLiteral("Z")
+          .toFormatter();
+
```
dsyer, line 56: I'm not sure this is necessary. Why do we need an explicit formatter for a format that already works by default with `OffsetDateTime` (I believe)?

extended/src/main/java/io/kubernetes/client/extended/leaderelection/resourcelock/LeaseLock.java
```diff
@@ -135,15 +135,23 @@ private LeaderElectionRecord getRecordFromLease(V1LeaseSpec lease) {
...
+    V1LeaseSpec spec =
+        new V1LeaseSpec()
+            .holderIdentity(record.getHolderIdentity())
+            .leaseDurationSeconds(record.getLeaseDurationSeconds())
+            .leaseTransitions(record.getLeaderTransitions());
+    if (record.getAcquireTime() != null) {
+      spec =
+          spec.acquireTime(
+              OffsetDateTime.ofInstant(
+                  Instant.ofEpochMilli(record.getAcquireTime().getTime()), ZoneOffset.UTC));
```
dsyer, line 147: This drops the milliseconds, right? Is that the intention? It also assumes that the timestamps are all UTC (but I think that is guaranteed by the k8s spec if they come from the server).

yue9944882: yes, the leader-elector manages resource-locks w/ millis-sec timestamp, so it doesn't make a difference whether `LeaseLock` provides extra precision.

kubernetes/src/main/java/io/kubernetes/client/openapi/JSON.java
yue9944882: kubernetes server-side requires 6-digit fraction for micro timestamp types (i.e. LeaseSpec#accquireTime) where "0" should be padded in the end if necessary. however, the default `DateTimeFormatter.ISO_OFFSET_DATE_TIME` doesn't support right-padding for micro-sec timestamps. e.g. by the default formatter`2018-04-03T11:32:26.1234Z` is a valid timestamp, but kubernetes server will be rejecting b/c lacking right-padded "0"s, it's expected to be `2018-04-03T11:32:26.123400Z`

brendandburns:
This LGTM but I will wait for an ack from @dsyer

dsyer:
I see what this does now, and while I hate to add even more manual overrides to the `JSON` utility, I guess this is the right thing to do. I think I would prefer `RFC3339MICRO_FORMATTER` to be private though (it's an implementation detail, used once in a test, but you could just as easily call `JSON.serialize()` there instead of `formatter.format()`).

dsyer:
I also think there might be a problem with using `java.util.Date` in the `LeaderElectionRecord` because it might use the system default timezone whereas the incoming timestamps from the API server are all in UTC. Seems like it would be simpler just to pass the `OffsetDateTime` through from the API to the leader elector.

yue9944882:
> I see what this does now, and while I hate to add even more manual overrides to the JSON utility, I guess this is the right thing to do. I think I would prefer RFC3339MICRO_FORMATTER to be private though (it's an implementation detail, used once in a test, but you could just as easily call JSON.serialize() there instead of formatter.format()).

@dsyer i made the formatter private in the latest commit

> I also think there might be a problem with using java.util.Date in the LeaderElectionRecord because it might use the system default timezone whereas the incoming timestamps from the API server are all in UTC. Seems like it would be simpler just to pass the OffsetDateTime through from the API to the leader elector.

refactoring the leader-elector to use OffsetDateTime to track the lease record sounds good to me. but i think we should defer that in a separate follow-up to unwound the e2e failure ASAP.

dsyer:
LGTM then.

Issue #1480, opened 2021-01-06
LeaderElectorTest is flaky in GitHub Actions


brendandburns:
It looks like requests to the Kubernetes API server are timing out, here is a detailed log:

```
[INFO] Running io.kubernetes.client.e2e.extended.leaderelection.LeaderElectorTest
8601 [controller-reflector-io.kubernetes.client.openapi.models.V1Namespace-1] DEBUG io.kubernetes.client.informer.cache.ReflectorRunnable - class io.kubernetes.client.openapi.models.V1Namespace#Extract resourceVersion 1006 list meta
8605 [controller-reflector-io.kubernetes.client.openapi.models.V1Namespace-1] DEBUG io.kubernetes.client.informer.cache.ReflectorRunnable - class io.kubernetes.client.openapi.models.V1Namespace#Start watching with 1006...
8606 [controller-reflector-io.kubernetes.client.openapi.models.V1Namespace-1] DEBUG io.kubernetes.client.informer.cache.ReflectorRunnable - class io.kubernetes.client.openapi.models.V1Namespace#Start watch with resource version 1006
8623 [controller-reflector-io.kubernetes.client.openapi.models.V1Namespace-1] DEBUG io.kubernetes.client.informer.cache.ReflectorRunnable - class io.kubernetes.client.openapi.models.V1Namespace#Receiving resourceVersion 42
8624 [controller-reflector-io.kubernetes.client.openapi.models.V1Namespace-1] DEBUG io.kubernetes.client.informer.cache.ReflectorRunnable - class io.kubernetes.client.openapi.models.V1Namespace#Receiving resourceVersion 43
8624 [controller-reflector-io.kubernetes.client.openapi.models.V1Namespace-1] DEBUG io.kubernetes.client.informer.cache.ReflectorRunnable - class io.kubernetes.client.openapi.models.V1Namespace#Receiving resourceVersion 155
8625 [controller-reflector-io.kubernetes.client.openapi.models.V1Namespace-1] DEBUG io.kubernetes.client.informer.cache.ReflectorRunnable - class io.kubernetes.client.openapi.models.V1Namespace#Receiving resourceVersion 244
8626 [controller-reflector-io.kubernetes.client.openapi.models.V1Namespace-1] DEBUG io.kubernetes.client.informer.cache.ReflectorRunnable - class io.kubernetes.client.openapi.models.V1Namespace#Receiving resourceVersion 982
[...]

PR #1481, opened 2021-01-06
Add logging dependency to e2e tests.

brendandburns:
Improves logging for e2e tests.

Ref #1480

### DIFF
--- a/e2e/pom.xml
+++ b/e2e/pom.xml
@@ -41,6 +41,19 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>${slf4j.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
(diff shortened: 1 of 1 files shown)


yue9944882:
@brendandburns do we still want this? it's intended for easing flakiness which is already solved i suppose?

brendandburns:
@yue9944882 I think we do still want this. Without it the real cluster integration tests don't have any logs and so failures are really hard to debug. So I think it is worth leaving this in for future debugging of flakes or other failures.

While with unit tests it's fairly easy to debug w/o logs, the real cluster integration tests are much harder to debug that way without all of the logging that the various components do in a real system.

brendandburns:
@yue9944882 friendly ping on this PR.

Thanks!

brendandburns:
e2e flaked again:
https://github.com/kubernetes-client/java/pull/1574

Having these logs would make it feasible to debug such failures.

brendandburns:
@yue9944882 friendly ping on this one.

brendandburns:
@yue9944882 can I get an LGTM on this one?

yue9944882 (commented):
/lgtm

PR #1490, opened 2021-01-11
use Preconditions::precondition + add tests

wind57:


util/src/main/java/io/kubernetes/client/apimachinery/GroupVersionKind.java
```diff
@@ -20,10 +22,7 @@
 
   public GroupVersionKind(String group, String version, String kind) {
     super(group, version);
-    if (kind == null) {
-      throw new IllegalArgumentException("kind must not be null");
-    }
-    this.kind = kind;
+    this.kind = precondition(kind, Objects::isNull, () -> "kind must not be null");
```
brendandburns, line 25: I don't know how I feel about this from a readability perspective.

@yue9944882 wdyt?

wind57: @brendanburns interesting. you had no problem with [this, though](https://github.com/kubernetes-client/java/pull/1468/files#diff-f6f5d3fa8282576aad84c7431df956e40e20b563ff1a0f544dc1e9257105e21aR30)

util/src/test/java/io/kubernetes/client/apimachinery/NamespaceNameTest.java
```diff
@@ -0,0 +1,41 @@
...
+limitations under the License.
+*/
+package io.kubernetes.client.apimachinery;
+
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import org.assertj.core.api.Assertions;
+import org.junit.Test;
+
+/** @author wind57 */
```
yue9944882, line 20: i suggest we drop the authoring in the tests

util/src/main/java/io/kubernetes/client/apimachinery/GroupVersionKind.java
yue9944882: i think i improves readability in a way, the refactoring is trivial so it looks good

brendandburns: ok. I'm fine with it. The other one feels cleaner somehow, I think it was because it removed two mostly identical assignments, but clearly it's a little inconsistent on my part.

util/src/test/java/io/kubernetes/client/apimachinery/NamespaceNameTest.java
brendandburns: agree with that.

brendandburns:
Please address the @author comment and then we can merge this.

Thanks for the patience!

wind57:
@brendanburns thank you for giving me the opportunity to contribute. These PRs are mainly for me to familiarize with the code, that is just my way of doing it. I will make more changes with far more substance in time.

PR #1516, opened 2021-01-30
Register ControllerFactory bean definition as opposite to bean instance

asavov:
With this change `KubernetesReconcilerProcessor` conforms to `BeanFactoryPostProcessor`
contract to register `BeanDefinitions` and not to manipulate bean instances.
As a result Reconciler beans are:
1) **not** created at all by `postProcessBeanFactory` (which was the core issue)
2) participate in Spring container management hooks, and hence `@Autowired`

### DIFF
--- a/spring/src/main/java/io/kubernetes/client/spring/extended/controller/KubernetesReconcilerProcessor.java
+++ b/spring/src/main/java/io/kubernetes/client/spring/extended/controller/KubernetesReconcilerProcessor.java
@@ -12,19 +12,17 @@ limitations under the License.
 */
 package io.kubernetes.client.spring.extended.controller;
 
-import io.kubernetes.client.extended.controller.Controller;
-import io.kubernetes.client.extended.controller.ControllerManager;
+import static org.springframework.util.Assert.notNull;
+
 import io.kubernetes.client.extended.controller.reconciler.Reconciler;
 import io.kubernetes.client.informer.SharedInformerFactory;
-import io.kubernetes.client.spring.extended.controller.annotation.KubernetesReconciler;
 import io.kubernetes.client.spring.extended.controller.factory.KubernetesControllerFactory;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-import org.springframework.beans.BeansException;
-import org.springframework.beans.factory.config.BeanFactoryPostProcessor;
+import java.util.function.Supplier;
+import org.springframework.beans.factory.config.BeanDefinition;
 import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
+import org.springframework.beans.factory.support.BeanDefinitionBuilder;
+import org.springframework.beans.factory.support.BeanDefinitionRegistry;
+import org.springframework.beans.factory.support.BeanDefinitionRegistryPostProcessor;
 import org.springframework.core.Ordered;
 
 /**
@@ -34,18 +32,21 @@ import org.springframework.core.Ordered;
  * <p>It will create a {@link io.kubernetes.client.extended.controller.Controller} for every
  * reconciler instances registered in the spring bean-factory.
  */
-public class KubernetesReconcilerProcessor implements BeanFactoryPostProcessor, Ordered {
+public class KubernetesReconcilerProcessor implements BeanDefinitionRegistryPostProcessor, Ordered {
 
-  private static final Logger log = LoggerFactory.getLogger(KubernetesReconcilerProcessor.class);
+  public static final String DEFAULT_SHARED_INFORMER_FACTORY_BEAN_NAME = "sharedInformerFactory";
 
-  private ControllerManager controllerManager;
+  private final String sharedInformerFactoryBeanName;
 
-  private ExecutorService controllerManagerDaemon = Executors.newSingleThreadExecutor();
+  private BeanDefinitionRegistry beanDefinitionRegistry;
 
-  private SharedInformerFactory sharedInformerFactory;
+  public KubernetesReconcilerProcessor() {
+    this(DEFAULT_SHARED_INFORMER_FACTORY_BEAN_NAME);
+  }
 
-  public KubernetesReconcilerProcessor(SharedInformerFactory sharedInformerFactory) {
-    this.sharedInformerFactory = sharedInformerFactory;
+  public KubernetesReconcilerProcessor(String sharedInformerFactoryBeanName) {
+    notNull(sharedInformerFactoryBeanName, "SharedInformerFactory bean name is required");
+    this.sharedInformerFactoryBeanName = sharedInformerFactoryBeanName;
   }
 
   @Override
@@ -54,20 +55,27 @@ public class KubernetesReconcilerProcessor implements BeanFactoryPostProcessor, 
   }
 
   @Override
-  public void postProcessBeanFactory(ConfigurableListableBeanFactory beanFactory)
-      throws BeansException {
-    String[] names = beanFactory.getBeanNamesForType(Reconciler.class);
-    for (String name : names) {
-      Reconciler reconciler = (Reconciler) beanFactory.getBean(name);
-      KubernetesReconciler kubernetesReconciler =
-          reconciler.getClass().getAnnotation(KubernetesReconciler.class);
-      String reconcilerName = kubernetesReconciler.value();
+  public void postProcessBeanDefinitionRegistry(BeanDefinitionRegistry registry) {
+    this.beanDefinitionRegistry = registry;
+  }
 
-      KubernetesControllerFactory controllerFactory =
-          new KubernetesControllerFactory(sharedInformerFactory, reconciler);
+  @Override
+  public void postProcessBeanFactory(ConfigurableListableBeanFactory beanFactory) {
+    for (String reconcilerName : beanFactory.getBeanNamesForType(Reconciler.class)) {
 
-      Controller controller = controllerFactory.getObject();
-      beanFactory.registerSingleton(reconcilerName, controller);
+      Supplier<KubernetesControllerFactory> kubernetesControllerFactorySupplier =
+          () ->
+              new KubernetesControllerFactory(
+                  beanFactory.getBean(sharedInformerFactoryBeanName, SharedInformerFactory.class),
+                  beanFactory.getBean(reconcilerName, Reconciler.class));
+
+      BeanDefinition controllerFactoryBeanDefinition =
+          BeanDefinitionBuilder.genericBeanDefinition(
+                  KubernetesControllerFactory.class, kubernetesControllerFactorySupplier)
+              .getBeanDefinition();
+
+      beanDefinitionRegistry.registerBeanDefinition(
+          reconcilerName + "Controller", controllerFactoryBeanDefinition);
     }
   }
 }
--- a/spring/src/test/java/io/kubernetes/client/spring/extended/controller/KubernetesReconcilerProcessorTest.java
+++ b/spring/src/test/java/io/kubernetes/client/spring/extended/controller/KubernetesReconcilerProcessorTest.java
@@ -1,0 +1,107 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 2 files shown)


yue9944882 (commented):
/lgtm

@asavov can you remove `@Autowired` from the commit message, otherwise it will lead to spammy github references on the page

brendandburns:
Looks like you need to run `mvn spotless:apply` to fix formatting.

asavov:
> /lgtm
> 
> @asavov can you remove `@Autowired` from the commit message, otherwise it will lead to spammy github references on the page

Done with updated commit.

asavov:
> Looks like you need to run `mvn spotless:apply` to fix formatting.

Done. And now seems ok.

asavov:
@dsyer @brendanburns https://github.com/kubernetes-client/java/pull/1516/checks has failed. Not sure it's related to this PR. How to Re-run it through UI? Or should I upload new "dummy" commit?

asavov:
**You just signed CNCF Individual Contributor License Agreement.** Seems this one is also checked.

dsyer:
Looking good. Sorry for the CLA hassle (at least you only have to do it once).

I would make a couple of small (but could be important) changes. I don’t see a need to inject the `SharedInformerFactory` and it will cause the same kind of instantiation cascade as the `getBean()` calls we were trying to avoid. Also we might need to be more defensive about the downcast of the `BeanFactory`.

asavov:
> Looking good. Sorry for the CLA hassle (at least you only have to do it once).

No worries at all. It's done.

> I don’t see a need to inject the `SharedInformerFactory` and it will cause the same kind of instantiation cascade as the `getBean()` calls we were trying to avoid. 

```
  static class TestReconciler implements Reconciler {

    @Autowired private SharedInformer<V1Pod> informerToBeInjected;
```
Do you refer this snippet? The whole idea is to be able to do exactly this logic: autowire other beans, benefiting from Spring container. I just used SharedInformer, cause it's natural fit, still it might be Env, or AppCtx or @Value(${some.prop}). Would provided more details, cause maybe I miss something.

> Also we might need to be more defensive about the downcast of the `BeanFactory`.

The way to go is to extend from `BeanDefinitionRegistryPostProcessor` and implement `postProcessBeanDefinitionRegistry`. Then the cast is not needed any more. I'm ok to do it, if you think that's the right approach.

dsyer:
`BeanDefinitionRegistryPostProcessor` would probably be better. The informer factory I mentioned wasn’t in a test though, I thought I saw it being injected directly into the post processor.

asavov:
> `BeanDefinitionRegistryPostProcessor` would probably be better. The informer factory I mentioned wasn’t in a test though, I thought I saw it being injected directly into the post processor.

```
public class KubernetesReconcilerProcessor implements BeanFactoryPostProcessor, Ordered {

  private final SharedInformerFactory sharedInformerFactory;

  public KubernetesReconcilerProcessor(SharedInformerFactory sharedInformerFactory) {
    this.sharedInformerFactory = sharedInformerFactory;
  }
```
You are absolutely right. Hmm. What's the Spring/right way for a `BeanFactoryPostProcessor` to get this information? A pointer to valid example within a Spring component would definitely help.

dsyer:
One way would be to set a constructor arg ref (not value) if you can get the bean name by type for the `SharedInformerFactory` from the `ListableBeanFactory`. If you look at callers of `BeanDefinitionBuilder.addConstructorArgReference()` you will find existing examples. Some of them work by using a naming convention - that's the simplest - and some have more elaborate ways of looking up the bean definition name for the reference. Or you can use a `Supplier` to defer the computation. I think this is safe (to create a `Bar` from an existing `Foo`):

```java
	@Bean
	public static BeanDefinitionRegistryPostProcessor processor() {
		return new BeanDefinitionRegistryPostProcessor() {

			private ConfigurableListableBeanFactory beanFactory;

			@Override
			public void postProcessBeanFactory(
					ConfigurableListableBeanFactory beanFactory) throws BeansException {
				this.beanFactory = beanFactory;
			}

			@Override
			public void postProcessBeanDefinitionRegistry(BeanDefinitionRegistry registry)
					throws BeansException {
				BeanDefinitionBuilder definition = BeanDefinitionBuilder
						.genericBeanDefinition(Bar.class,
								() -> new Bar(beanFactory.getBean(Foo.class)));
				registry.registerBeanDefinition("bar", definition.getBeanDefinition());
			}
		};
	}
```

asavov:
Looking at the code of `KubernetesInformerFactoryProcessor` and `KubernetesReconcilerProcessor` I came to a conclusion (maybe I'm wrong) that something very generic is mis-implemented. Those classes are both `BeanDefinitionRegistryPostProcessor`, which implies they work/manipulate on `BeanDefinitions` or refs or lazy suppliers, still they both accept ready Bean Instances as constr params (`ApiClient` and `SharedInformerFactory`).
How about having some live chat to further discuss this or ... cause if we want to impl it right it seems like a breaking change. Or maybe I lack enough knowledge on the subject.

dsyer:
Those two post processors are unrelated to this change though, right? They need to be fixed in probably almost identical ways, if we decide we want to keep them, but I don't think we have to interrupt this PR process to deal with that.

asavov:
> Those two post processors are unrelated to this change though, right? They need to be fixed in probably almost identical ways, if we decide we want to keep them, but I don't think we have to interrupt this PR process to deal with that.

I miss that, cause the focus of the change is `KubernetesReconcilerProcessor`. Or?

dsyer:
I'm sorry, you are right. You fixed one of them in this PR, though? So we can leave the other one for another effort?

asavov:
> I'm sorry, you are right. You fixed one of them in this PR, though? So we can leave the other one for another effort?

Should I go with current version (if approved) OR remove
```
  public KubernetesReconcilerProcessor(SharedInformerFactory sharedInformerFactory) {
    this.sharedInformerFactory = sharedInformerFactory;
  }
```
causing SharedInformerFactory to be pre-instantiated and default to the SINGLETON `SharedInformerFactory` instance to be used/selected by KubernetesReconcilerProcessor#postProcessBeanFactory. Thus no bean instantiation at least by `KubernetesReconcilerProcessor`, based on some defaults, let's say one-and-only-one instance of SharedInformerFactory.

```
  @Override
  public void postProcessBeanFactory(ConfigurableListableBeanFactory beanFactory) {
    for (String reconcilerName : beanFactory.getBeanNamesForType(Reconciler.class)) {
      BeanDefinition controllerFactoryBeanDefinition =
          BeanDefinitionBuilder.genericBeanDefinition(KubernetesControllerFactory.class)
              .addConstructorArgReference(<the name of singleton shared informer factory>)
              .addConstructorArgReference(reconcilerName)
              .getBeanDefinition();

      // Replace this with call to BDRegistry passed by BeanDefinitionRegistryPostProcessor
      ((DefaultListableBeanFactory) beanFactory)
          .registerBeanDefinition(reconcilerName + "Controller", controllerFactoryBeanDefinition);
    }
  }

```

dsyer:
That would work. We'd have to be sure about the bean name. The lazy instantiation alternative with a `Supplier` is better though IMO (no reflection).

asavov:
@dsyer Does this version satisfy-them-all ? :)
```
public class KubernetesReconcilerProcessor implements BeanDefinitionRegistryPostProcessor, Ordered {

  public static final String DEFAULT_SHARED_INFORMER_FACTORY_BEAN_NAME = "sharedInformerFactory";

  private final SharedInformerFactory sharedInformerFactory;

  private final String sharedInformerFactoryBeanName;

  private ConfigurableListableBeanFactory beanFactory;

  @Deprecated
  // TODO: Should we keep this for backward compatibility?
  public KubernetesReconcilerProcessor(SharedInformerFactory sharedInformerFactory) {
    this.sharedInformerFactory = sharedInformerFactory;
    this.sharedInformerFactoryBeanName = null;
  }

  public KubernetesReconcilerProcessor() {
    this(DEFAULT_SHARED_INFORMER_FACTORY_BEAN_NAME);
  }

  public KubernetesReconcilerProcessor(String sharedInformerFactoryBeanName) {
    notNull(sharedInformerFactoryBeanName, "SharedInformerFactory bean name is required");
    this.sharedInformerFactory = null;
    this.sharedInformerFactoryBeanName = DEFAULT_SHARED_INFORMER_FACTORY_BEAN_NAME;
  }

  @Override
  public int getOrder() {
    return KubernetesInformerFactoryProcessor.ORDER + 1;
  }

  @Override
  public void postProcessBeanFactory(ConfigurableListableBeanFactory beanFactory) {
    this.beanFactory = beanFactory;
  }

  @Override
  public void postProcessBeanDefinitionRegistry(BeanDefinitionRegistry registry) {
    for (String reconcilerName : beanFactory.getBeanNamesForType(Reconciler.class)) {

      Supplier<KubernetesControllerFactory> kubernetesControllerFactorySupplier;

      if (sharedInformerFactory != null) {
        kubernetesControllerFactorySupplier =
            () ->
                new KubernetesControllerFactory(
                    sharedInformerFactory, beanFactory.getBean(reconcilerName, Reconciler.class));
      } else {
        kubernetesControllerFactorySupplier =
            () ->
                new KubernetesControllerFactory(
                    beanFactory.getBean(sharedInformerFactoryBeanName, SharedInformerFactory.class),
                    beanFactory.getBean(reconcilerName, Reconciler.class));
      }

      BeanDefinition controllerFactoryBeanDefinition =
          BeanDefinitionBuilder.genericBeanDefinition(
                  KubernetesControllerFactory.class, kubernetesControllerFactorySupplier)
              .getBeanDefinition();

      registry.registerBeanDefinition(
          reconcilerName + "Controller", controllerFactoryBeanDefinition);
    }
  }
}
```

dsyer:
That would work, but I don't see why we need to keep the old constructor, or the code path with `if (sharedInformerFactory != null)`.

asavov:
> That would work, but I don't see why we need to keep the old constructor, or the code path with `if (sharedInformerFactory != null)`.

If backward compatibility is not an issue - my pleasure to simplify it :)

dsyer:
IMO no-one is using this class directly - it's all via autoconfiguration. Plus although it still says 11.0.1-SNAPSHOT in the top level POM, I know the next release will be 12.0.0 because of the Joda time changes. So I don't think we should care about backwards compatibility here. @yue9944882 would have to say for sure, but she also said that other day that this whole module is considered a prototype still, so breaking changes are expected.

asavov:
> IMO no-one is using this class directly - it's all via autoconfiguration. Plus although it still says 11.0.1-SNAPSHOT in the top level POM, I know the next release will be 12.0.0 because of the Joda time changes. So I don't think we should care about backwards compatibility here. @yue9944882 would have to say for sure, but she also said that other day that this whole module is considered a prototype still, so breaking changes are expected.

PR update with new code as discussed.

spring/src/main/java/io/kubernetes/client/spring/extended/controller/KubernetesReconcilerProcessor.java
```diff
@@ -54,20 +55,27 @@ public int getOrder() {
...
-      String reconcilerName = kubernetesReconciler.value();
+  public void postProcessBeanDefinitionRegistry(BeanDefinitionRegistry registry) {
+    this.beanDefinitionRegistry = registry;
+  }
+
+  @Override
+  public void postProcessBeanFactory(ConfigurableListableBeanFactory beanFactory) {
+    for (String reconcilerName : beanFactory.getBeanNamesForType(Reconciler.class)) {
+
+      Supplier<KubernetesControllerFactory> kubernetesControllerFactorySupplier =
```
dsyer, line 66: I guess this can be inlined. It's a matter of taste though.

asavov: Agree it's opinionated. I prefer readability vs. compact code.
By the way the PR is finally green... :)

dsyer:
LGTM

asavov:
> LGTM

Nice! and thank you.

This is my first contribution, so the PR is merge by someone on kubernetes-client team with write access. No more actions on my site?

dsyer:
Yes, we have to wait for @brendanburns and @yue9944882 . If they have no more changes to request they can merge it without you doing anything else.

yue9944882 (commented):
/lgtm
/approve

thanks @dsyer @asavov!

dsyer:
@asavov are you going to do the same thing for `KubernetesInformerFactoryProcessor`?

asavov:
> KubernetesInformerFactoryProcessor

@dsyer I'd like to. Still at the moment not enough bandwidth.

dsyer:
See #1532 for my solution

Issue #1505, opened 2021-01-21
Rewrite support for k8s loadbalancer in spring module


dsyer:
I don't like any of the annotations in the spring module much (can we review them all at some point?), but `@KubernetesEndpointsLoadBalanced` is by far the most egregious. The fact that it only works with field injection (a notorious antipattern) is the worst part, but I have other questions. Could we maybe re-implement it as a `@Qualifier` or just a plain factory?

brendandburns:
cc @yue9944882 I'm definitely not enough of a Spring expert to comment too deeply. Can you provide more details of what you would change?

yue9944882:
@dsyer the load-balancer support is yet experimental. am always open to the changes, plz feel free to submit pull requests to the codebase

PR #1518, opened 2021-02-01
Create LoadBalancer by hand instead of via annotation

dsyer:
See #1505


yue9944882 (commented):
/lgtm
/approve

thanks @dsyer!

PR #1522, opened 2021-02-03
Make Maven reactor build work

dsyer:
This will make IDEs work better (especially VSCode) because
otherwise they cannot compute the classpath for the examples.

### DIFF
--- a/examples/examples-release-10/pom.xml
+++ b/examples/examples-release-10/pom.xml
@@ -1,13 +1,18 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
-    <groupId>io.kubernetes</groupId>
+    <parent>
+        <groupId>io.kubernetes</groupId>
+        <artifactId>client-java-examples-parent</artifactId>
+        <version>12.0.0-SNAPSHOT</version>
+        <relativePath>..</relativePath>
+    </parent>
+
     <artifactId>client-java-examples-release-10</artifactId>
     <packaging>bundle</packaging>
     <name>client-java-examples-release-10</name>
-    <version>1.0.0</version>
+    <version>12.0.0-SNAPSHOT</version>
     <properties>
-        <kubernetes.client.version>10.0.1</kubernetes.client.version>
         <java.version>1.8</java.version>
         <maven.compiler.source>${java.version}</maven.compiler.source>
         <maven.compiler.target>${java.version}</maven.compiler.target>
@@ -27,27 +32,27 @@
         <dependency>
             <groupId>io.kubernetes</groupId>
             <artifactId>client-java-api</artifactId>
-            <version>${kubernetes.client.version}</version>
+            <version>10.0.1</version>
         </dependency>
         <dependency>
             <groupId>io.kubernetes</groupId>
             <artifactId>client-java</artifactId>
-            <version>${kubernetes.client.version}</version>
+            <version>10.0.1</version>
         </dependency>
         <dependency>
             <groupId>io.kubernetes</groupId>
             <artifactId>client-java-extended</artifactId>
-            <version>${kubernetes.client.version}</version>
+            <version>10.0.1</version>
         </dependency>
         <dependency>
             <groupId>io.kubernetes</groupId>
             <artifactId>client-java-spring-integration</artifactId>
-            <version>${kubernetes.client.version}</version>
+            <version>10.0.1</version>
         </dependency>
         <dependency>
             <groupId>io.kubernetes</groupId>
             <artifactId>client-java-proto</artifactId>
-            <version>${kubernetes.client.version}</version>
+            <version>10.0.1</version>
         </dependency>
         <dependency>
             <groupId>commons-cli</groupId>
--- a/examples/pom.xml
+++ b/examples/pom.xml
@@ -4,21 +4,20 @@
 	<parent>
 		<groupId>io.kubernetes</groupId>
 		<artifactId>client-java-parent</artifactId>
-		<version>11.0.1-SNAPSHOT</version>
+		<version>12.0.0-SNAPSHOT</version>
 		<relativePath>../pom.xml</relativePath>
 	</parent>
 
-	<version>11.0.1-SNAPSHOT</version>
+	<version>12.0.0-SNAPSHOT</version>
 
 	<artifactId>client-java-examples-parent</artifactId>
 	<packaging>pom</packaging>
 	<name>client-java-examples-parent</name>
 
-	<properties>
-		<kubernetes.client.version>11.0.1-SNAPSHOT</kubernetes.client.version>
-	</properties>
-
 	<modules>
+		<module>examples-release-10</module>
+		<module>examples-release-11</module>
+		<module>examples-release-12</module>
 	</modules>
 
 	<build>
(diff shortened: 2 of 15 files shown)


examples/examples-release-10/pom.xml
```diff
@@ -1,7 +1,13 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
-    <groupId>io.kubernetes</groupId>
+    <parent>
+        <groupId>io.kubernetes</groupId>
+        <artifactId>client-java-examples-parent</artifactId>
+        <version>11.0.1-SNAPSHOT</version>
```
brendandburns, line 7: I don't think we want this, do we? This makes these examples part of the overall release build (which definitely messed things up before.

examples/examples-release-11/src/main/java/io/kubernetes/client/examples/ProtoExample.java
```diff
@@ -53,8 +53,7 @@ public static void main(String[] args) throws IOException, ApiException, Interru
     System.out.println(ns);
     if (ns.object != null) {
       namespace =
-          ns.object
```
brendandburns, line 56: There is a delta between formatting w/ Java8 and formatting w/ Java11+ *sigh* 

You'll need to revert this.

examples/examples-release-12/src/main/java/io/kubernetes/client/examples/ProtoExample.java
```diff
@@ -53,8 +53,7 @@ public static void main(String[] args) throws IOException, ApiException, Interru
     System.out.println(ns);
     if (ns.object != null) {
       namespace =
-          ns.object
```
brendandburns, line 56: Here too.

examples/pom.xml
```diff
@@ -15,10 +15,13 @@
 	<name>client-java-examples-parent</name>
 
 	<properties>
-		<kubernetes.client.version>11.0.1-SNAPSHOT</kubernetes.client.version>
+		<kubernetes.client.version>${project.version}</kubernetes.client.version>
 	</properties>
 
 	<modules>
```
brendandburns, line 21: I explicitly pulled this out last time I did a release because I didn't want these included in the release.

examples/examples-release-10/pom.xml
dsyer: It doesn't get deployed because the parent has `<skip>true</true>` in the the deploy plugin. What else could go wrong?

examples/examples-release-11/src/main/java/io/kubernetes/client/examples/ProtoExample.java
dsyer: Wait, so if I run with Java8 it will work? This format change was created by the spotless plugin AFAIK.

examples/pom.xml
dsyer: This isn't anything to do with a release. It's just a property that is used to specify a dependency.

examples/examples-release-11/src/main/java/io/kubernetes/client/examples/ProtoExample.java
dsyer: Looks like it :-)

examples/examples-release-10/pom.xml
brendandburns: The maven release plugin looks for these strings and replaces them every time you run a release.

examples/pom.xml
brendandburns: The maven release plugin pulls in everything that is referenced by the top level POM (at least it did the last time I ran a release) so including these as sub-modules for this pom causes them to get pulled into a release.

examples/examples-release-10/pom.xml
dsyer: The trick is to not make your modules break if the project version changes. I don't think this one will. It looks safe to me.

examples/pom.xml
dsyer: You can switch off plugins with `skip=true` (as is done for the deploy plugin in the examples here). But that's probably not what you want - you want to just make sure that the projects don't break if their artifact version changes. I'm pretty sure these ones don't now. The problem is that without this change *nothing* in the examples is built automatically from the top level, so it's easy to miss something important when refactoring.

yue9944882: > The problem is that without this change nothing in the examples is built automatically from the top level, so it's easy to miss something important when refactoring.

my IDE (intellj-idea) won't build the examples unless applying some tricky manual configuration. i suppose we can explicitly skip the deploying steps for the release plugin [here's an example configuration](https://stackoverflow.com/a/31495604)

dsyer: We already have that (https://github.com/kubernetes-client/java/blob/master/examples/pom.xml#L35).

yue9944882: interesting.. i see the module `client-java-examples-parent` published in the [last release](https://search.maven.org/artifact/io.kubernetes/client-java-examples-parent/11.0.0/pom). this should have been skipped, is it?

dsyer: It would have been skipped if this change had been applied. I don’t know how things looked when the last release was done.

dsyer: Actually I don't know because I don't know how the release was performed. Is there a script or something? I don't see it in the Github actions.

dsyer: This is strange, for instance, given that I believe the release plugin has been used: https://github.com/kubernetes-client/java/blob/master/pom.xml#L421 (`<skip>true</skip>` in the parent). I'm not sure that's a good idea. But `mvn deploy` would work, once the versions were set up correctly. Personally I don't find the release plugin very useful, but if you do want to sue it, probably it's best not to disable it at the top level?

brendandburns: @dsyer 

The current release process is documented here:
https://github.com/kubernetes-client/java/blob/master/RELEASES.md

I'm definitely open to suggestions about changing/improving the release process, especially if it included a GitHub action for continuous snapshot releases. I currently find the release process quite cumbersome.

Happy for someone who is more knowledgeable than me about releasing things via Maven take a look. What we have was cobbled together by amateurs :)

dsyer: A GitHub Action for snapshots would be excellent, but I don't think it's within the scope of this PR. You can easily set one up if you have permission to deploy snapshots somewhere. You could probably ask Sonatype for permission based on whatever credentials you use for the releases.

It looks like there is a bug in the release plugin (https://issues.apache.org/jira/browse/MRELEASE-799) which makes the examples a bit awkward to configure - they have to use a hard-coded version for the kubernetes-api dependencies. That seems like a small price to pay to get this whole thing working properly. (IMO the examples for older versions should just be in a branch in git, but we discussed that before and you didn't agree, so we have to work with this.)

My latest change includes a workaround for that bug, and I tested it by running `mvn release:prepare` (which failed when it tried to `git push` the change, but that's expected since I don't have permission).

yue9944882: +1 it will be great if we can publish snapshots on a daily basis via GH actions automatically

brendandburns:
One last question and then I think this is good to go:

Why use `12.0.0` instead of `12.0.0-SNAPSHOT` don't we want to use a snapshot version for now so that when we do an actual `12.0.0` release people with local builds of 12.0.0 don't use an old build?

Also, needs a rebase.

dsyer:
> Why use 12.0.0 instead of 12.0.0-SNAPSHOT

Duh. That was a mistake (caused by the release plugin, but lets continue to work with that while we have no other easy options). Fixed it when I rebased.

brendandburns:
/lgtm
/approve

Thanks for the patience on this one.

PR #1527, opened 2021-02-06
addEventHandlerWithResyncPeriod fix

georgiano:
resyncPeriodMillis is analyzed but never used inside addEventHandlerWithResyncPeriod() method.
probably there is a typo on line 158 and first parameter should be resyncPeriodMillis


yue9944882:
@georgiano can you sign the CLA so that we can proceed this?

PR #1528, opened 2021-02-07
Feat: Dynamic kubernetes api for classless api requesting

yue9944882:


util/src/main/java/io/kubernetes/client/util/generic/dynamic/DynamicKubernetesApi.java
```diff
@@ -0,0 +1,78 @@
...
+        DynamicKubernetesListObject.class,
+        apiGroup,
+        apiVersion,
+        resourcePlural,
+        apiClient);
+    TypeAdapter<KubernetesObject> objAdapter =
+        apiClient.getJSON().getGson().getAdapter(KubernetesObject.class);
+    TypeAdapter<KubernetesListObject> objListAdapter =
+        apiClient.getJSON().getGson().getAdapter(KubernetesListObject.class);
+    if (false
```
brendandburns, line 52: delete `false`? It's not doing anything here.

brendandburns:
One comment on some stray code. I think this would also benefit from an example to show how to use it.

util/src/main/java/io/kubernetes/client/util/generic/dynamic/DynamicKubernetesApi.java
yue9944882: deleted

brendandburns:
@yue9944882 Is it possible to add an example? If not, I can lgtm and we can add it later.

yue9944882:
@brendandburns just added an example for the dynamic client as well as the wiki page

davidxia:
Thanks so much for this feature! I love it, and my team is using it now.

PR #1532, opened 2021-02-08
Ensure KubernetesInformerFactoryProcessor does not access any beans

dsyer:
A `BeanDefinitionRegistryPostProcessor` should not ideally refer to any
bean instances, except lazily. This change makes the bean definitions
that are installed by a `@KubernetesInformer` lazier by putting them
behind a `Supplier`

See also #1516


spring/src/main/java/io/kubernetes/client/spring/extended/controller/KubernetesInformerFactoryProcessor.java
```diff
@@ -121,6 +61,75 @@ public int getOrder() {
...
+        ResolvableType.forClassWithGenerics(
+            SharedInformer.class, kubernetesInformer.apiTypeClass());
+    informerBean.setTargetType(informerType);
+    registry.registerBeanDefinition(kubernetesInformer.apiTypeClass().getName(), informerBean);
+  }
+
+  private void registerLister(
+      BeanDefinitionRegistry registry, KubernetesInformer kubernetesInformer) {
+    RootBeanDefinition listerBean =
+        (RootBeanDefinition)
```
asavov, line 98: how about removing the cast ? or RootBeanDefinition  is the only way to call setTargetType?
how about using directly BeanDefinitionBuilder#genericBeanDefinition(java.lang.Class<T>, Supplier<T>) ? or same as above?

spring/src/main/java/io/kubernetes/client/spring/extended/controller/KubernetesInformerFactoryProcessor.java
```diff
@@ -121,6 +61,75 @@ public int getOrder() {
...
+        () -> lister(kubernetesInformer.apiTypeClass(), kubernetesInformer));
+    ResolvableType listerType =
+        ResolvableType.forClassWithGenerics(Lister.class, kubernetesInformer.apiTypeClass());
+    listerBean.setTargetType(listerType);
+    registry.registerBeanDefinition(listerType.toString(), listerBean);
+  }
+
+  @SuppressWarnings({"unchecked", "rawtypes"})
+  private <T extends KubernetesObject> Lister<T> lister(
+      Class<T> type, KubernetesInformer kubernetesInformer) {
```
asavov, line 110: neither args are used. what's the intention?

spring/src/main/java/io/kubernetes/client/spring/extended/controller/KubernetesInformerFactoryProcessor.java
```diff
@@ -121,6 +61,75 @@ public int getOrder() {
...
+    SharedIndexInformer sharedInformer = this.beanFactory.getBean(SharedIndexInformer.class);
```
asavov, line 111: what if more SharedIndexInformer are registered?
this impl does not "couple" with corresponding informer, as previously existed.

spring/src/main/java/io/kubernetes/client/spring/extended/controller/KubernetesInformerFactoryProcessor.java
```diff
@@ -121,6 +61,75 @@ public int getOrder() {
   @Override
   public void postProcessBeanDefinitionRegistry(BeanDefinitionRegistry registry)
       throws BeansException {
-    this.beanDefinitionRegistry = registry;
+    if (registry instanceof BeanFactory) {
+      for (String name : registry.getBeanDefinitionNames()) {
+        Class<?> cls = ((BeanFactory) registry).getType(name);
```
asavov, line 66: seems all beans are inspected for @ KubernetesInformers , which IMO is suboptimal. maybe some restriction (as prev existed) is not a bad idea. How about KubernetesInformers at Reconcilers only?

Isn't it problematic that there's no check for duplication and two Components might register same PodInformer twice. Cause now you have different sources of KubernetesInformers, while prev it was only through singleton SharedInformerFactory.

spring/src/test/java/io/kubernetes/client/spring/extended/controller/KubernetesReconcilerCreatorTest.java
```diff
@@ -94,7 +94,7 @@ public SharedInformerFactory sharedInformerFactory() {
                   apiVersion = "v1",
                   resourcePlural = "configmaps")),
     })
-    static class TestSharedInformerFactory extends SharedInformerFactory {}
+    static class TestSharedInformerFactory {}
```
asavov, line 97: IMO the naming might be more close to the semantic. It's just a "holder" for @KubernetesInformers, kind of configuration.

spring/src/main/java/io/kubernetes/client/spring/extended/controller/KubernetesInformerFactoryProcessor.java
```diff
@@ -121,6 +61,75 @@ public int getOrder() {
   @Override
   public void postProcessBeanDefinitionRegistry(BeanDefinitionRegistry registry)
       throws BeansException {
-    this.beanDefinitionRegistry = registry;
+    if (registry instanceof BeanFactory) {
```
asavov, line 64: Is there a way to prevent from casting?

asavov (commented):
Great starter. Still IMO needs some polishing.

spring/src/main/java/io/kubernetes/client/spring/extended/controller/KubernetesInformerFactoryProcessor.java
dsyer: It’s the only way. It’s safe by construction.

spring/src/main/java/io/kubernetes/client/spring/extended/controller/KubernetesInformerFactoryProcessor.java
dsyer: Good catch. I meant to fix that. Need a new test.

spring/src/main/java/io/kubernetes/client/spring/extended/controller/KubernetesInformerFactoryProcessor.java
dsyer: I think that’s beyond the scope of this PR. The inefficiency is also insignificant - either we have to check all beans or Spring does. The work gets done one way or the other. (Unless we really want to restrict the type of the annotated beans, which was the original design, but doesn’t make any sense to me.)

spring/src/test/java/io/kubernetes/client/spring/extended/controller/KubernetesReconcilerCreatorTest.java
dsyer: Just lazy here. The name was from the old test.

spring/src/main/java/io/kubernetes/client/spring/extended/controller/KubernetesInformerFactoryProcessor.java
dsyer: No

spring/src/main/java/io/kubernetes/client/spring/extended/controller/KubernetesInformerFactoryProcessor.java
dsyer: See below.

spring/src/main/java/io/kubernetes/client/spring/extended/controller/KubernetesInformerFactoryProcessor.java
dsyer: I suppose we could make this whole thing an `ApplicationContextInitializer` instead of a `BeanDefinitionRegistryPostProcessor`. But they are equivalent and both the downcast and the initializer only but always work with Spring Boot.

spring/src/main/java/io/kubernetes/client/spring/extended/controller/KubernetesInformerFactoryProcessor.java
asavov: > Isn't it problematic that there's no check for duplication and two Components might register same PodInformer twice. Cause now you have different sources of KubernetesInformers, while prev it was only through singleton SharedInformerFactory.

How about at least to create a test with two sources of KubernetesInformers to see (and be prepared) what will happen?

spring/src/main/java/io/kubernetes/client/spring/extended/controller/KubernetesInformerFactoryProcessor.java
asavov: How about to revert the if condition then, so we have flat (and less indented) code?
Also do the casting once and re-use it from within the loop.

spring/src/main/java/io/kubernetes/client/spring/extended/controller/KubernetesInformerFactoryProcessor.java
dsyer: Not sure what you mean by "2 sources". There is only one `SharedIndexInformer` per `ApiType` and it's the same one as the code in master, so where would the second one come from? Why would anyone need 2?

spring/src/main/java/io/kubernetes/client/spring/extended/controller/KubernetesInformerFactoryProcessor.java
dsyer: I wouldn't want to do that, and risk a class cast exception, even theoretically. If indented code hurts your eyes you can propose a patch later, I guess.

Not sure what you mean about the re-use. It is only used once, so it's neater to do it in line IMO.

spring/src/main/java/io/kubernetes/client/spring/extended/controller/KubernetesInformerFactoryProcessor.java
asavov: Till now one was supposed to create **one** `ShatedInformerFactory` annotated with `KubernetesInformers`. This is what I call one-single-source of `KubernetesInformers`.
With this review one might provide more components annotated with KubernetesInformers. (this is what I call multiple sources of ...) such as:

```
    @KubernetesInformers({
      @KubernetesInformer(
          apiTypeClass = V1Pod.class,
          apiListTypeClass = V1PodList.class,
          groupVersionResource = @GroupVersionResource(resourcePlural = "pods"))
    })
    @Component
    static class SourceOneOfKubernetesInformers {}

    @KubernetesInformers({
      @KubernetesInformer(
          apiTypeClass = V1Pod.class,
          apiListTypeClass = V1PodList.class,
          groupVersionResource = @GroupVersionResource(resourcePlural = "pods"))
    })
    @Component
    static class SourceTwoOfKubernetesInformers {}
```
So I'm concerned about such a scenario, and how it's handled by the code.

spring/src/main/java/io/kubernetes/client/spring/extended/controller/KubernetesInformerFactoryProcessor.java
asavov: > How about to revert the if condition then, so we have flat (and less indented) code?

Excuse my language! `Revert if condition` means, not remove it but:

if (!(registry instanceof BeanFactory)) {
  return;
}
// Current code goes here with no indentation

asavov: > Not sure what you mean about the re-use. It is only used once, so it's neater to do it in line IMO.

```
BeanFactory bf = (BeanFactory) registry;
for (String name : registry.getBeanDefinitionNames()) {
        Class<?> cls = bf.getType(name);
```
Do not do the cast on every bean name from within the loop, but do it once prior the loop.

dsyer: That's an extra line of code. I don't see the value.

asavov: Your take is final!

spring/src/main/java/io/kubernetes/client/spring/extended/controller/KubernetesInformerFactoryProcessor.java
dsyer: There was nothing stopping you from doing that before (but on the same bean). It would fail at runtime, so you would know it was wrong, even if the exception might be confusing without some polish. I don't see any reason to fix it as part of this PR because it's not a new problem and it's going to fail at runtime anyway.

asavov: Agree with you. 
Still spotting the duplication on a single ShatedInformerFactory seems much easier that spotting it on two different components (in two diff packages) if the message is not hinting you. Seems as good-to-have, and not must-to-have for this PR.

yue9944882 (commented):
@dsyer @asavov thanks! (sry for delaying, just got back from lunar new year PTO)

Issue #1533, opened 2021-02-09
getRelistResourceVersion() is too old and causes infinite errors


tony-clarke-amdocs:
I tried updating to client 11.0.0. However, the logs get filled with the following errors:

```
2021-02-09 08:37:29.076 ERROR 78258 --- [s.V1ConfigMap-1] i.k.c.informer.cache.ReflectorRunnable   : class io.kubernetes.client.openapi.models.V1ConfigMap#Reflector loop failed unexpectedly

io.kubernetes.client.openapi.ApiException: Gone
	at io.kubernetes.client.openapi.ApiClient.handleResponse(ApiClient.java:993) ~[client-java-api-11.0.0.jar:na]
	at io.kubernetes.client.openapi.ApiClient.execute(ApiClient.java:905) ~[client-java-api-11.0.0.jar:na]
	at io.kubernetes.client.informer.SharedInformerFactory$1.list(SharedInformerFactory.java:207) ~[client-java-11.0.0.jar:na]
	at io.kubernetes.client.informer.cache.ReflectorRunnable.run(ReflectorRunnable.java:77) ~[client-java-11.0.0.jar:na]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) ~[na:1.8.0_211]
	at java.util.concurrent.FutureTask.runAndReset$$$capture(FutureTask.java:308) ~[na:1.8.0_211]
	at java.util.concurrent.FutureTask.runAndReset(FutureTask.java) ~[na:1.8.0_211]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180) ~[na:1.8.0_211]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294) ~[na:1.8.0_211]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149) ~[na:1.8.0_211]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624) ~[na:1.8.0_211]
	at java.lang.Thread.run(Thread.java:748) ~[na:1.8.0_211]

[...]

PR #1543, opened 2021-02-18
Fixes reflector list expiration

yue9944882:
Fixes: https://github.com/kubernetes-client/java/issues/1533

### DIFF
--- a/util/src/test/java/io/kubernetes/client/informer/cache/ReflectorRunnableTest.java
+++ b/util/src/test/java/io/kubernetes/client/informer/cache/ReflectorRunnableTest.java
@@ -12,6 +12,7 @@ limitations under the License.
 */
 package io.kubernetes.client.informer.cache;
 
+import static org.junit.Assert.assertFalse;
 import static org.mockito.Mockito.any;
 import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.times;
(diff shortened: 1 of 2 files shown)


yue9944882:
/cc @tony-clarke-amdocs 

@brendandburns i suppose we should pick this back to 10 and 11 release, on which branch should open the cherry-picking pull?

dsyer:
Master is beyond 11.0 now right (it has the breaking changes in the time API)? So we need to bump master to 12.0.0-SNAPSHOT and backport this change to 11.0 (at least).

brendandburns:
@yue9944882 

The right branches are `release-10` and `release-11` 

Thanks!

util/src/test/java/io/kubernetes/client/informer/cache/ReflectorRunnableTest.java
```diff
@@ -215,4 +217,42 @@ public V1PodList list(CallGeneratorParams params) throws ApiException {
...
   }
+
+  @Test
+  public void testReflectorListShouldHandleExpiredResourceVersion() throws ApiException {
+    String expectedResourceVersion = "100";
+    when(listerWatcher.list(any()))
+        .thenReturn(
+            new V1PodList().metadata(new V1ListMeta().resourceVersion(expectedResourceVersion)));
+    // constantly failing watches will make the reflector run only one time
+    when(listerWatcher.watch(any())).thenThrow(new ApiException(HttpURLConnection.HTTP_GONE, ""));
```
aaruna, line 228: What about the scenario wherein this API returns responseCode 200 but the body has an event that has status.code=410 ?
https://github.com/kubernetes-client/java/pull/1611/files#diff-31beacc81d066c2ab68de67a3587fee7bd7ef10d9ba42dc0a180b621751c6bffR226

PR #1550, opened 2021-02-22
Supporting password protected key-store

yue9944882:
Fixes: https://github.com/kubernetes-client/java/issues

now we can override the password for key-store by invoking `ClientBuilder#setKeyStorePassphrase` in the builder flow


yue9944882:
@agr-priyanshu c @brendandburns

agr-priyanshu:
Hi, thanks for the patch. I see this commit is providing capability to pass KeyStorePassword while creating the SSL context which will solve the current issue but still there is a potential dependency on user-defined KeyStore which might be a bit problematic. From Java; it is not enforced that KeyStore and Key must have the same password and there are setups where KeyStore have PW1 and Key1 have PW2 and Key2 have PW3.

I cam across this project [sslcontext-kickstart](https://github.com/Hakky54/sslcontext-kickstart) which have very nice handling of all combinations; something may be we can look for in future adoption.

For the current issue I also tried submitting a PR 1551 (mainly to keep the dedicated SSL context) but since its my first here; it is stuck in CLA approval.

Issue #1536, opened 2021-02-10
EventLogger produces unacceptable JSON Patches when updating events


chris-chambers:
The event patch creation code (for updating events whose count will be > 1) produces a JSON patch that the API server doesn't accept (returns HTTP 422).

https://github.com/kubernetes-client/java/blob/46607021cf4a0d5dfc6ef3414eb0ce7b3d1c1210/extended/src/main/java/io/kubernetes/client/extended/event/legacy/EventLogger.java#L44-L49

It looks like this was originally ported from Go client, which has `strategicpatch.CreateTwoWayMergePatch` to handle patch creation.  This version is creating the patch content manually, but has a couple of errors:

- The `\"count\":\"%d\"` section should have its value unquoted: `\"count\":%d`
- The date should, I think, be formatted as an ISO-8601 string value, rather than the behavior of `Configuration.getDefaultApiClient().getJSON().serialize(now)`, which produces a structured JSON representation of the datetime

The result of this error is that logging any event with the same content (so that an existing event should have its count updated, rather than creating a new event) fails silently.

I monkey-patched like this and got it working:

```java
 new V1Patch( 
     String.format( 
         "{\"message\":\"%s\",\"count\":%d,\"lastTimestamp\":\"%s\"}",
         event.getMessage(), 
         event.getCount(), 
         DateTimeFormatter.ISO_OFFSET_DATE_TIME.format(now))); 
```

But, I'm not familiar with this code, so maybe there is a better solution.

chris-chambers:
As an aside, I see that these classes are marked `legacy`.  Are there other classes I should prefer to use for emitting events?

brendandburns:
cc @yue9944882

yue9944882:
@chris-chambers thank you for reporting this, i just opened a fix for this issue.

> As an aside, I see that these classes are marked legacy. Are there other classes I should prefer to use for emitting events?

a newer event API was recently introduced which basically uses `events.k8s.io/events` api instead of `v1/events` (i.e. legacy events api). as far as i can see, even if the older api now goes "legacy" but it will keep working for a long period of time. in a word, we dont have to land on the new api and the legacy should be working fine.

PR #1559, opened 2021-02-24
Fixes event patching - unquote event count

yue9944882:
Fixes: https://github.com/kubernetes-client/java/issues/1536

cc @chris-chambers @brendandburns

### DIFF
--- a/extended/src/main/java/io/kubernetes/client/extended/event/legacy/EventLogger.java
+++ b/extended/src/main/java/io/kubernetes/client/extended/event/legacy/EventLogger.java
@@ -43,7 +43,7 @@ public class EventLogger {
       patch =
           new V1Patch(
               String.format(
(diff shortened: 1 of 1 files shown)


brendandburns:
/lgtm
/approve

@yue9944882 is this worth cherry-picking?

chris-chambers:
Does this really completely fix it?  In my tests, `Configuration.getDefaultApiClient().getJSON().serialize(now)` did not produce a valid date string.  It produced a JSON object representation that Kubernetes rejected.  I had to switch to `DateTimeFormatter.ISO_OFFSET_DATE_TIME.format(now)` to get a working event update.  (There may be a more appropriate date formatting call, but that one worked for me.)

yue9944882:
@chris-chambers ah true, sry for being dump 🙁 really shouldn't have updated the timestamp serialization. i opened https://github.com/kubernetes-client/java/pull/1569 to fix it.

PR #1560, opened 2021-02-24
Spotless: groovy formatting

yue9944882:
spotless maven plugin began to support groovy formatting since Sep 2020, this pull adds groovy configuration


brendandburns:
/lgtm
/approve

groovy!

PR #1567, opened 2021-02-27
Add option to prevent LeaderElector from getting elected

PSanetra:
This PR adds a configuration option to prevent the LeaderElector instance to get elected as leader. This is useful if it is desired to have certain instances, which should just get informed about the current leader, but should never get elected.

### DIFF
--- a/extended/src/main/java/io/kubernetes/client/extended/leaderelection/LeaderElector.java
+++ b/extended/src/main/java/io/kubernetes/client/extended/leaderelection/LeaderElector.java
@@ -119,6 +119,16 @@ public class LeaderElector implements AutoCloseable {
   /**
    * Runs the leader election in foreground.
    *
+   * @param onNewLeaderHook called when the client observes a leader that is not the previously
+   *     observed leader. This includes the first observed leader when the client starts.
+   */
+  public void run(Consumer<String> onNewLeaderHook) {
+    run(() -> {}, () -> {}, onNewLeaderHook);
+  }
+
+  /**
+   * Runs the leader election in foreground.
+   *
    * @param startLeadingHook called when a LeaderElector client starts leading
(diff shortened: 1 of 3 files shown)


extended/src/main/java/io/kubernetes/client/extended/leaderelection/LeaderElector.java
```diff
@@ -116,6 +116,16 @@ public void run(Runnable startLeadingHook, Runnable stopLeadingHook) {
     run(startLeadingHook, stopLeadingHook, null);
   }
 
+  /**
+   * Runs the leader election in foreground.
+   *
+   * @param onNewLeaderHook called when the client observes a leader that is not the previously
+   *     observed leader. This includes the first observed leader when the client starts.
+   */
+  public void run(Consumer<String> onNewLeaderHook) {
```
yue9944882, line 125: do we need this new `run` method? (looks that it's only used in unit-testing)

yue9944882 (commented):
one nit, otherwise LGTM

brendandburns:
Hrm, I'm not sure if I like this design.

Instead of having this be a part of the standard leader election class where you can't get elected, maybe a better thing would be to subclass a `LeaderObserver` to make it explicit that you are watching, not participating?

I worry about people skimming code and being confused why they're not getting elected.  I feel like explicitly splitting out the `LeaderObserver` will make this more apparent to anyone reading the code.

wdyt?

PSanetra:
@brendanburns I think an advantage of the current design may be that you can change the behavior at runtime by changing the configuration.

brendandburns:
@PSanetra I don't see how having a `LeaderObserver` class changes that. You can always change what you instantiate.

Additionally, I'm not sure I understand a situation when you would want to dynamically decide whether you were an observer or a participant in the election. Can you explain your use case further?

PR #1572, opened 2021-03-02
Fix the build for JODA time

brendandburns:
There was a cherry-pick of an event logger fix, but it was after the switch to OffsetDateTime which isn't present in the 11.0.x branch.


yue9944882 (commented):
/lgtm
/approve

Issue #1578, opened 2021-03-06
ReflectorRunnable.watchHandler seems to be stuck after about 5min of inactivity


karunasagark:
I'm working on custom controller where I noticed this issue. After about 5min of inactivity i.e. no updates in the api server, the controller stops receiving watch events from the api server. So any add/delete on the custom resource after the 5min of inactivity on api server, is not reconciled by the controller .

On further debugging, what I've seen is that ReflectorRunnable.watchHandler which should be constantly reading from the response stream seems to be stuck. The `Receiving resourceVersion` log from ReflectorRunnable doesn't show up. Also, `kubectl get <customresource> -w` sees the updates as expected.

The minimal repo is here [kubetest.zip](https://github.com/kubernetes-client/java/files/6094485/kubetest.zip) and I'm able to reproduce this consistently. I tried running this using both java 8 and java 11, with the same result as above. I got a jstack dump when the issue occurred, but couldn't notice any issue, the reflector thread was in runnable state.

```
"controller-reflector-com.XXXXX-0" #23 prio=5 os_prio=0 cpu=39.77ms elapsed=986.61s allocated=1080K defined_classes=39 tid=0x00007f19b856c800 nid=0x1bbe9 runnable  [0x00007f19b5ff2000]
   java.lang.Thread.State: RUNNABLE
        at java.net.SocketInputStream.socketRead0(java.base@11.0.10/Native Method)
        at java.net.SocketInputStream.socketRead(java.base@11.0.10/SocketInputStream.java:115)
        at java.net.SocketInputStream.read(java.base@11.0.10/SocketInputStream.java:168)
        at java.net.SocketInputStream.read(java.base@11.0.10/SocketInputStream.java:140)
        at sun.security.ssl.SSLSocketInputRecord.read(java.base@11.0.10/SSLSocketInputRecord.java:478)
        at sun.security.ssl.SSLSocketInputRecord.readHeader(java.base@11.0.10/SSLSocketInputRecord.java:472)
        at sun.security.ssl.SSLSocketInputRecord.bytesInCompletePacket(java.base@11.0.10/SSLSocketInputRecord.java:70)
        at sun.security.ssl.SSLSocketImpl.readApplicationRecord(java.base@11.0.10/SSLSocketImpl.java:1354)
        at sun.security.ssl.SSLSocketImpl$AppInputStream.read(java.base@11.0.10/SSLSocketImpl.java:963)
        at okio.Okio$2.read(Okio.java:140)
[...]

PR #1588, opened 2021-03-12
Enforcing read-timeout for informer-factory processor via property "kubernetes.informer.clientReadTimeout" (default 5min)

yue9944882:
https://github.com/kubernetes-client/java/blob/d9271d47aa6f00e7e28b3df3281ef4f282442acb/spring/src/main/java/io/kubernetes/client/spring/extended/controller/KubernetesInformerFactoryProcessor.java#L95-L96

these lines were dropped during refactoring informer-processor and this pull adds it back. tho this can workaround by explicitly setting the read-timeout to zero upon instanciating the ApiClient bean.

/cc @brendandburns @dsyer


brendandburns:
I'm still not sure we want to force the timeout to infinite. There are lots of examples where the network is a little flaky and timing out a some interval (eg. 3 minutes) makes sense.

brendandburns:
e.g. https://github.com/kubernetes-client/java/issues/1578

yue9944882:
@brendandburns i updated readtimeout to 5min to align w/ the reflector

PR #1589, opened 2021-03-12
Flake: Fixes leader-election e2e flakiness by relaxing timeout to 45s

yue9944882:
in the flaking e2e test, the original timeout is 30s and meanwhile the `leaseDuration=30s` and `retryInterval=5s`. so the lock competitors are jittered by `(jitter_factor (0~1.2) + 1) * 5s <= 11s` hence extending the timeout margin w/ 15s will easing the flakiness


yue9944882:
see the failures at https://github.com/kubernetes-client/java/pull/1588

PR #1603, opened 2021-03-17
Chore: Support KubernetesInformers annotation on bean constructor

yue9944882:


brendandburns:
I tried to merge the conflicts, but I think I messed something up.

Issue #1606, opened 2021-03-19
Error Logs of ReflectorRunnable caused by too 'old resource version'


FlorianBuchegger:
**Describe the bug**
We are using the SharedIndexInformer and notice error logs appearing from time to time, f.e.:
```
2021-03-11 03:35:41.406 UTC ERROR 76235 --- [.V1Deployment-1] i.k.c.informer.cache.ReflectorRunnable   : class io.kubernetes.client.openapi.models.V1Deployment#Reflector loop failed unexpectedly
java.lang.RuntimeException: got ERROR event and its status: class V1Status {
    apiVersion: v1
    code: 410
    details: null
    kind: Status
    message: too old resource version: 8886375 (8888886)
    metadata: class V1ListMeta {
        _continue: null
        remainingItemCount: null
        resourceVersion: null
        selfLink: null
    }
    reason: Expired
    status: Failure
}
	at io.kubernetes.client.informer.cache.ReflectorRunnable.watchHandler(ReflectorRunnable.java:226) ~[client-java-11.0.1.jar:na]
	at io.kubernetes.client.informer.cache.ReflectorRunnable.run(ReflectorRunnable.java:124) ~[client-java-11.0.1.jar:na]
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515) ~[na:na]
	at java.base/java.util.concurrent.FutureTask.runAndReset(FutureTask.java:305) ~[na:na]
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:305) ~[na:na]
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128) ~[na:na]
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628) ~[na:na]
	at java.base/java.lang.Thread.run(Thread.java:834) ~[na:na]
```
As we do not want to dissable log output for the whole SharedIndexInformer class, we wonder if the log level for this particular problem could be reduced to info?

** Client Version **
`11.0.1`

** Kubernetes Version **
`1.19.6`

** Java Version **
Java 11

**To Reproduce**
Have the SharedIndexInformer run a long time without changes in kubernetes.

**Expected behavior**
The log level should be INFO in this case as the watch gets fixed automatically.

**Additional context**
[...]

PR #1611, opened 2021-03-24
Chore: Reducing rewatch logging verbosity upon normal re-connection

yue9944882:
Fixes: https://github.com/kubernetes-client/java/issues/1606

### DIFF
--- a/util/src/main/java/io/kubernetes/client/informer/cache/ReflectorRunnable.java
+++ b/util/src/main/java/io/kubernetes/client/informer/cache/ReflectorRunnable.java
@@ -223,10 +223,15 @@ public class ReflectorRunnable<
         continue;
       }
       if (eventType.get() == EventType.ERROR) {
-        String errorMessage =
-            String.format("got ERROR event and its status: %s", item.status.toString());
-        log.error(errorMessage);
-        throw new RuntimeException(errorMessage);
+        if (item.status != null && item.status.getCode() == HttpURLConnection.HTTP_GONE) {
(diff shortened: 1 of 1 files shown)


util/src/main/java/io/kubernetes/client/informer/cache/ReflectorRunnable.java
```diff
@@ -223,10 +223,15 @@ private void watchHandler(Watchable<ApiType> watch) {
         continue;
       }
       if (eventType.get() == EventType.ERROR) {
-        String errorMessage =
-            String.format("got ERROR event and its status: %s", item.status.toString());
-        log.error(errorMessage);
-        throw new RuntimeException(errorMessage);
+        if (item.status != null && item.status.getCode() == HttpURLConnection.HTTP_GONE) {
+          log.info("Watch connection expired: {}", item.status.getMessage());
```
aaruna, line 227: Should `isLastSyncResourceVersionUnavailable` be set to `true` here? @yue9944882

Related PR: https://github.com/kubernetes-client/java/pull/1543/files#diff-31beacc81d066c2ab68de67a3587fee7bd7ef10d9ba42dc0a180b621751c6bffR160

haoming-db: This is a bug which has broken us.
Setting `isLastSyncResourceVersionUnavailable` to `true` will not fix the problem, because `isLastSyncResourceVersionUnavailable` is not referenced in the `while(true)` loop anyway.
We should keep the `throw new RuntimeException`.

PR #1617, opened 2021-03-30
Bugfix: Properly handling non-410 ApiExcetion in the reflector

yue9944882:
In the pull https://github.com/kubernetes-client/java/pull/1543, we added catcher for intercepting 410 status code but the non-410 errors will be missing. this pull makes the reflector properly handles the non-410 codes.


brendandburns:
Is there a test that is needed to make sure we don't have regressions on this in the future?

yue9944882:
@brendandburns added tests for validating the behavior

PR #1623, opened 2021-03-31
Improve error condition handling somewhat.

brendandburns:


util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
```diff
@@ -254,8 +260,11 @@ public static ClientBuilder cluster() throws IOException {
   }
 
   protected ClientBuilder setBasePath(String host, String port) {
+    if (host == null) {
```
yue9944882, line 263:
```suggestion
    if (StringUtils.isEmpty(host))
```
to handle the empty string, wdyt?

brendandburns:
Suggestion integrated, please re-check.

Thanks!

PR #1626, opened 2021-04-02
Add more debug/error logs and ensure controllerManager shutdown in case of exception

aaruna:


extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
```diff
@@ -154,9 +154,10 @@ public void run() {
               this.worker();
             } catch (Throwable t) {
               log.error("Unexpected controller loop abortion", t);
+            } finally {
```
brendandburns, line 157: I don't think that this finally clause adds anything here, please remove.

brendandburns:
Thanks for the PR, one small comment.

extended/src/main/java/io/kubernetes/client/extended/controller/DefaultController.java
aaruna: imho still having this in finally explains the intent (and safeguard against any changes in the above `catch`) that we want the `latch.countDown()` to always be executed when exiting this block. WDYT?

brendandburns: I think it's misleading. I guess it is a style thing. In my mind it's equivalent to:

```java
if (foo) {
  return bar;
}
return baz
```

vs
```java
if (foo) {
  return bar;
} else {
  return baz;
}
```

I don't think the extra verbosity of having `finally` in there serves any purpose, it's pretty clear that the code will run no matter what.

aaruna: Unlike the above `if` example, there is a difference here (comments inline)

```
try {
 ......
} catch (Throwable t) {
 ......
}


// This may not get executed if the code inside catch block throws an exception
// do something here
........ 
```

vs

```
try {
 ......
} catch (Throwable t) {
 ......
} finally {
  // This will always get executed even if the code inside catch block throws an exception
  // do something here
  ........ 
}
```

yue9944882: @aaruna is right, unlike `if` clause the `finally` clause will be executed with or without any exception happended https://docs.oracle.com/javase/tutorial/essential/exceptions/finally.html

yue9944882 (commented):
/lgtm
/approve

@brendandburns am getting this in, plz let me know if there's anything i missed or will track in a follow-up

PR #1627, opened 2021-04-02
Fix name and shutdown resync executor in Controller.shutdown

aaruna:


aaruna:
Thanks for your review and merging this PR

PR #1628, opened 2021-04-02
Reflector list should handle expired resource version from watch handler as well

aaruna:


brendandburns:
This looks good to me, but I want @yue9944882 to take a look at this also, since he was the primary author of this code.

aaruna:
@yue9944882 Any thoughts on this?

util/src/main/java/io/kubernetes/client/informer/cache/ReflectorRunnable.java
```diff
@@ -224,7 +224,11 @@ private void watchHandler(Watchable<ApiType> watch) {
       }
       if (eventType.get() == EventType.ERROR) {
         if (item.status != null && item.status.getCode() == HttpURLConnection.HTTP_GONE) {
-          log.info("Watch connection expired: {}", item.status.getMessage());
+          log.info(
+              "ResourceVersion {} and Watch connection expired: {} , will retry w/o resourceVersion next time",
+              getRelistResourceVersion(),
+              item.status.getMessage());
+          isLastSyncResourceVersionUnavailable = true;
```
yue9944882, line 231: @aaruna upon 410 status code, the watch handler is supposed to return and we deal w/ the RV expiration at the list call. this is working as designed. https://github.com/kubernetes/kubernetes/blob/a96000311f5beca111debc8727018e42cbb5dc79/staging/src/k8s.io/client-go/tools/cache/reflector.go#L129-L131

yue9944882:
/hold

for discussion

util/src/main/java/io/kubernetes/client/informer/cache/ReflectorRunnable.java
aaruna: Are you referring to the next list call? Because the current watch handler won't throw an ApiException and only that is handled in the catch block above. (The unit test added in this PR shows an example of this scenario)

CC @jiangzhou in case I have missed anything in the scenario above.

yue9944882: > Are you referring to the next list call?

yeah. IIRC watch handler doesnt directly throw an ApiException but it will deal w/ error status nested in the decoded watch event. 
https://github.com/kubernetes-client/java/blob/c580990fa9de11a549cc1ae4219cb4fae994b9b5/util/src/main/java/io/kubernetes/client/informer/cache/ReflectorRunnable.java#L227-L237

brendandburns:
@yue9944882 what is the status of this PR/Discussion?

util/src/main/java/io/kubernetes/client/informer/cache/ReflectorRunnable.java
aaruna: But that's not resetting `isLastSyncResourceVersionUnavailable=true`. The unit test below has an example

tony-clarke-amdocs:
I have upgrade to client 13.0.0 and now we are floored with these messages. It looks like a very clear regression. I am getting [deja vu](https://github.com/kubernetes-client/java/issues/1533). It looks like the same pattern, we are not resetting isLastSyncResourceVersionUnavailable back to true when HTTP_GONE occurs. @aaruna change looks good to me. 13.0.0 is unusable as it is. Waiting for update on this. @aaruna can you rebase?

aaruna:
@tony-clarke-amdocs I have rebased the PR.

tony-clarke-amdocs:
Thanks @aaruna. @brendandburns @yue9944882 Can you guys review when you get a chance. It's a blocker for us.

tony-clarke-amdocs:
@brendandburns @yue9944882 Any ETA for this PR? If there is anything we can do to move it forward please let us know.

yue9944882:
/close 

picked and merged  #1927

PR #1655, opened 2021-04-16
Chore: Removing version from default user-agent header

yue9944882:
unless we figure out how to efficiently update the version string (`11.0.1-SNAPSHOT`) somehow..

### DIFF
--- a/kubernetes/src/main/java/io/kubernetes/client/openapi/ApiClient.java
+++ b/kubernetes/src/main/java/io/kubernetes/client/openapi/ApiClient.java
@@ -103,7 +103,8 @@ public class ApiClient {
     json = new JSON();
 
     // Set default User-Agent.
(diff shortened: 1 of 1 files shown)


brendandburns:
Can we add a GitHub issue and a TODO so that we remember to fix this?

kubernetes/src/main/java/io/kubernetes/client/openapi/ApiClient.java
```diff
@@ -103,7 +103,7 @@ private void init() {
     json = new JSON();
 
     // Set default User-Agent.
-    setUserAgent("Kubernetes Java Client/11.0.1-SNAPSHOT");
+    setUserAgent("Kubernetes Java Client");
```
brendandburns, line 106: Perhaps we should also add the Java runtime version?

yue9944882: added todo line

Issue #1665, opened 2021-04-21
io.kubernetes.client.spring.extended.manifests.KubernetesFromConfigMapTest is flaky.


brendandburns:
```
Error:  Tests run: 3, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 4.795 s <<< FAILURE! - in io.kubernetes.client.spring.extended.manifests.KubernetesFromConfigMapTest
Error:  testValueUpdate(io.kubernetes.client.spring.extended.manifests.KubernetesFromConfigMapTest)  Time elapsed: 1.77 s  <<< FAILURE!
junit.framework.ComparisonFailure: expected:<bar2> but was:<null>
	at junit.framework.Assert.assertEquals(Assert.java:100)
	at junit.framework.Assert.assertEquals(Assert.java:107)
	at junit.framework.TestCase.assertEquals(TestCase.java:260)
	at io.kubernetes.client.spring.extended.manifests.KubernetesFromConfigMapTest.testValueUpdate(KubernetesFromConfigMapTest.java:102)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.springframework.test.context.junit4.statements.RunBeforeTestExecutionCallbacks.evaluate(RunBeforeTestExecutionCallbacks.java:74)
	at org.springframework.test.context.junit4.statements.RunAfterTestExecutionCallbacks.evaluate(RunAfterTestExecutionCallbacks.java:84)
	at org.springframework.test.context.junit4.statements.RunBeforeTestMethodCallbacks.evaluate(RunBeforeTestMethodCallbacks.java:75)
[...]

PR #1667, opened 2021-04-23
Flake: Relaxing FromConfigMap tests timeout to 2s

yue9944882:


yue9944882:
ref: https://github.com/kubernetes-client/java/issues/1665

brendandburns:
I'd rather see something deterministic rather than just a longer sleep, but this will fix things somewhat so lgtm

/lgtm
/approve

PR #1669, opened 2021-04-24
Revise the Java versions we test to LTS + latest non-LTS version.

brendandburns:
@yue9944882 I think we shouldn't test/support Java versions that are no longer supported by the Java team.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -28,7 +28,7 @@ jobs:
     strategy:
       matrix:
         # test against latest update of each major Java version, as well as specific updates of LTS versions:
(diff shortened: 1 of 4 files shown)


yue9944882 (commented):
/lgtm
/approve

yue9944882:
@brendandburns ah it seems that gson is not compatible w/ java 16, so the test is failing: https://github.com/google/gson/issues/1875

brendandburns:
Yep, I'm going to see about fixing this.

brendandburns:
@yue9944882 I figured out how to make this work by adding the command line flag: `--illegal-access=warn` to the test JVM.

This probably won't work forever, but it will at least get us working for now.

yue9944882 (commented):
/lgtm
/approve

PR #1676, opened 2021-04-27
Update the CustomConstructor class for SnakeYAML.

brendandburns:
cc @yue9944882


util/src/main/java/io/kubernetes/client/util/Yaml.java
```diff
@@ -357,8 +360,11 @@ protected NodeTuple representJavaBeanProperty(
   }
 
   /** @return An instantiated SnakeYaml Object. */
-  public static org.yaml.snakeyaml.Yaml getSnakeYaml() {
```
yue9944882, line 360: this is a public method, should we mark it as deprecated and keep this for 1~2 releases for backward-compatibility?

brendandburns: Done. (though this should really be private)

yue9944882 (commented):
/lgtm

PR #1685, opened 2021-05-06
Added explicit runtime dependency for logback

dsyer:
The samples will not log anything when they run without a logging
library. Logback is only in test scope because of the parent pom
(I think). So we need to add it back explicitly.


yue9944882 (commented):
/lgtm
/approve

thanks @dsyer

PR #1686, opened 2021-05-06
Fix SpringExampleController so it doesn't crash on startup

### LINKED ISSUE #1684: SpringControllerExample fails to start
Both the version "12" and the version "13" sample apps fail to start because of a missing `SharedInformer` bean, so the annotation processing (which I dislike anyway) is not working. I guess we need to fix it, but I'd rather just remove the annotations, really.

dsyer:
This fix is masking another problem somewhere else, but at least it gets
the apps working while we investigate.

Fixes #1684 (temporarily at least). Also rebased on #1685.

### DIFF
--- a/examples/examples-release-12/src/main/java/io/kubernetes/client/examples/SpringControllerExample.java
+++ b/examples/examples-release-12/src/main/java/io/kubernetes/client/examples/SpringControllerExample.java
@@ -73,13 +73,6 @@ public class SpringControllerExample {
         SharedInformerFactory sharedInformerFactory, Reconciler reconciler) {
       return new KubernetesControllerFactory(sharedInformerFactory, reconciler);
     }
-
-    // *REQUIRED*
-    // Injecting your SharedInformerFactory class annotated `@KubernetesInformers`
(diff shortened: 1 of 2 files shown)


dsyer:
I understand the problem. It was actually introduced in ce7bc1bba7 when we fixed the post processor. The old version used to instantiate beans (very bad) but then it had the concrete runtime type to inspect. The new version only has the declared bean type, which in these samples is `SharedInformerFactory`, and that has no interesting annotations. The fix in this PR just ensures that the declared type has annotations that can be processed.

yue9944882 (commented):
LGTM, @dsyer can you rebase onto the latest master so we can have a cleaner tree structure (b/c github does non-ff merge by default)

dsyer:
Github can do "rebase on merge". Anyway I rebased manually and pushed it back.

yue9944882 (commented):
> Github can do "rebase on merge". Anyway I rebased manually and pushed it back.

thanks @dsyer, we're merging via the bot which i'm not sure capable of doing that 

/lgtm
/approve

PR #1687, opened 2021-05-06
Removes Kubernetes annotations from Spring integration

dsyer:
I never liked those annotations. The Java API for creating informers
and controllers is pretty straightforward and if you need one, you 
know it, so I don't see the value in declaring them via annotations.

This PR shows what it would look like. I don't think there is more
code in the samples, and what is there is quite explicit about what
it does and why it exists.


yue9944882:
@dsyer the informer annotation is introduced as an experimental functionality in the beginning, the point is to help developers to build their own controllers in a more declarative approach for those having less background-knowledge about the informer internals. i agree that the current informer-annotation looks dense and it doesn't even make much difference between constructing informers via annotation or manual bean instantiation, but i do think it's valuable to make the informer/controller instantiation easier by leveraging spring annotations. w/ that said, i'm leaning on refactoring a more simplified informer-annotation instead of removing the whole lump. developers can always opt-out from informer-annotation and do the manual informer creation, is it?

OTOH in the sense of backward-compatibility, we're supposed to deprecate and leave this for 2-ish release.

dsyer:
> we're supposed to deprecate and leave this for 2-ish release

I thought you told me the Spring features were "experimental" so we can make breaking changes. I don't mind though. If you agree that we can pull the switch on this then I'll refactor to a deprecation. It was only for discussion really.

While it is true that this is an optional feature, I believe it has so little value that it's not really worth keeping (in any form, but please propose something else if you like the general idea). I recommend that we make a clean break and just document though examples the best way to create Kubernetes controllers with Spring. I'm sure there are other ways we can improve the experience - I'm just pretty sure the existing annotations is not it.

yue9944882:
> If you agree that we can pull the switch on this then I'll refactor to a deprecation.

yeah it works for me

yue9944882:
@brendandburns SGTM, i will sort out a roadmap of coming features for developing the spring integration this year.

yue9944882 (commented):
/lgtm
/approve

PR #1689, opened 2021-05-06
Change the env variables implementation for tests.

brendandburns:
SystemLambda appears to be more recently updated than SystemRules. I also hope that it is/will be compatible with JDK16.


brendandburns:
Closing in favor of #1689

Issue #1721, opened 2021-05-20
Periodic resync is not working


ydmao:
**Describe the bug**
`ResyncRunnable.run` always returns false when `shouldResync` is set to `SharedProcessor.shouldResync`, causing the periodic resync to never trigger.

This seems to be a regression introduced by https://github.com/kubernetes-client/java/commit/e63a25645c59834054825eebc1c64bbeb71f4964#diff-50c262394352f8f93b9c7e3a305be21174dc42b68fc44a23dd5cd304837a2539R39

** Client Version **
12.0.0

** Kubernetes Version **
1.18.3

** Java Version **
Java 8

**To Reproduce**

**Expected behavior**

**KubeConfig**
If applicable, add a KubeConfig file with secrets redacted.

**Server (please complete the following information):**
 - OS: [e.g. Linux]
 - Environment [e.g. container]
 - Cloud [e.g. Azure]

**Additional context**

PR #1727, opened 2021-05-24
Fixes duplicated SharedProcessor#shouldResync calls

yue9944882:
Fixes: https://github.com/kubernetes-client/java/issues/1721
@ydmao

### DIFF
--- a/util/src/main/java/io/kubernetes/client/informer/ResyncRunnable.java
+++ b/util/src/main/java/io/kubernetes/client/informer/ResyncRunnable.java
@@ -36,7 +36,7 @@ public class ResyncRunnable implements Runnable {
   public void run() {
     log.debug("ResyncRunnable#resync ticker tick");
 
-    if (shouldResyncFunc == null || (shouldResyncFunc.get() != null && shouldResyncFunc.get())) {
+    if (shouldResyncFunc == null || Boolean.TRUE.equals(shouldResyncFunc.get())) {
       log.debug("ResyncRunnable#force resync");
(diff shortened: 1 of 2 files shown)


ydmao:
this is great - thank you! When will we have the next release with this fix in?

util/src/main/java/io/kubernetes/client/informer/ResyncRunnable.java
```diff
@@ -36,7 +36,7 @@ public ResyncRunnable(DeltaFIFO store, Supplier<Boolean> shouldResyncFunc) {
   public void run() {
     log.debug("ResyncRunnable#resync ticker tick");
 
-    if (shouldResyncFunc == null || (shouldResyncFunc.get() != null && shouldResyncFunc.get())) {
+    if (shouldResyncFunc == null || Boolean.TRUE.equals(shouldResyncFunc.get())) {
```
codeyq, line 39: Mind also adding a regression test in `util/src/test/java/io/kubernetes/client/informer/ResyncRunnableTest.java`?

codeyq (commented):
Thank you for the quick turnaround. Could this PR be back-ported to the 12.0.0 release branch?

brendandburns:
looks good, but adding a unit test makes sense.

util/src/main/java/io/kubernetes/client/informer/ResyncRunnable.java
yue9944882: SGTM

PR #1744, opened 2021-06-09
Revise the way we cache maven builds to include the java version, and get rid of restore-keys

brendandburns:
We've been having problems with broken caches
e.g. #1740 and #1741

I think this will help resolve it.


yue9944882:
@brendandburns hopefully this time it will eventually fix the cache issue 🤞

yue9944882 (commented):
/lgtm

brendandburns:
Looks like that didn't fix it :(

I have no idea what is going wrong, I guess it is not the cache, will try to dig in more.

PR #1769, opened 2021-06-25
Chore: Trimming pom files for fluent modules

yue9944882:
also, this pull adds a new profile named `fluent-gen` to include the fluent-gen module so that we can opt-in to code-gen via `mvn -Pfluent-gen clean compile`


brendandburns:
/lgtm
/approve

Thanks for cleaning this up!

PR #1776, opened 2021-07-06
Chore: Adding `admissionreview` to the submodule list

yue9944882:
cc @brendandburns

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -15,10 +15,11 @@
     <module>extended</module>
     <module>fluent</module>
     <module>spring</module>
-    <module>client-java-contrib/cert-manager</module>
     <module>e2e</module>
-    <module>client-java-contrib/prometheus-operator</module>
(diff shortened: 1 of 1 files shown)


client-java-contrib/admissionreview/pom.xml
```diff
@@ -13,33 +13,22 @@
     </parent>
     <version>12.0.1-SNAPSHOT</version>
 
-    <dependencies>
```
brendandburns, line 16: I think you need to retain these dependencies. The compile is failing in CI/CD.

brendandburns:
This looks good (thanks) but we need to retain the dependencies, I think.

client-java-contrib/admissionreview/pom.xml
yue9944882: fixed the compiling failure

Issue #1223, opened 2020-09-03
[Umbrella Issue] Many tests use Thread.sleep instead of synchronization primitives


brendandburns:
Many of our tests simply have a `Thread.sleep(duration)` when testing multi-threaded code. This is inherently flaky.

Instead we should use synchronization classes (generally `Semaphore`) to synchronized the tests which is both faster in the general case and not flaky in the extreme case.

See for instance #1218 and #1219

brendandburns:
- [ ] util/src/test/java/io/kubernetes/client/informer/cache/ReflectorRunnableTest.java
- [ ] extended/src/test/java/io/kubernetes/client/extended/workqueue/DefaultDelayingQueueTest.java
- [ ] util/src/test/java/io/kubernetes/client/informer/cache/ProcessorListenerTest.java
- [ ] util/src/test/java/io/kubernetes/client/informer/cache/SharedProcessorTest.java
- [ ] util/src/test/java/io/kubernetes/client/CopyTest.java
- [ ] util/src/test/java/io/kubernetes/client/informer/cache/ControllerTest.java
- [ ] extended/src/test/java/io/kubernetes/client/extended/workqueue/DefaultWorkQueueTest.java
- [ ] extended/src/test/java/io/kubernetes/client/extended/leaderelection/LeaderElectionTest.java
- [ ] util/src/test/java/io/kubernetes/client/informer/impl/DefaultSharedIndexInformerTest.java
- [ ] extended/src/test/java/io/kubernetes/client/extended/controller/LeaderElectingControllerTest.java
- [ ] util/src/test/java/io/kubernetes/client/PortForwardTest.java
- [ ] util/src/test/java/io/kubernetes/client/AttachTest.java
- [ ] util/src/test/java/io/kubernetes/client/ExecTest.java
- [ ] extended/src/test/java/io/kubernetes/client/extended/workqueue/ratelimiter/BucketRateLimiterTest.java
- [ ] spring/src/test/java/io/kubernetes/client/spring/extended/controller/KubernetesInformerCreatorTest.java
- [ ] spring/src/test/java/io/kubernetes/client/spring/extended/controller/KubernetesReconcilerCreatorTest.java
- [ ] extended/src/test/java/io/kubernetes/client/extended/event/EventCorrelatorTest.java

yue9944882:
@brendandburns thanks for sorting out these potentially flaking tests.. will fix them later!

/assign

sarveshkaushal:
[...]

PR #1786, opened 2021-07-17
remove some of thread.sleeps in test cases

karthikkondapally:
trying to remove some of thread.sleeps
https://github.com/kubernetes-client/java/issues/1223

### DIFF
--- a/extended/src/test/java/io/kubernetes/client/extended/leaderelection/LeaderElectionTest.java
+++ b/extended/src/test/java/io/kubernetes/client/extended/leaderelection/LeaderElectionTest.java
@@ -285,16 +285,20 @@ public class LeaderElectionTest {
     leaderElectionConfig.setLeaseDuration(Duration.ofMillis(1000));
     leaderElectionConfig.setRetryPeriod(Duration.ofMillis(200));
     leaderElectionConfig.setRenewDeadline(Duration.ofMillis(700));
+    final Semaphore sem = new Semaphore(1);
     LeaderElector leaderElector =
-        new LeaderElector(leaderElectionConfig, (t) -> actualException.set(t));
+        new LeaderElector(leaderElectionConfig, (t) -> { actualException.set(t); sem.release(); });
 
     ExecutorService leaderElectionWorker = Executors.newFixedThreadPool(1);
+    sem.acquire();
     leaderElectionWorker.submit(
         () -> {
           leaderElector.run(() -> {}, () -> {});
         });
-    // TODO: Remove this sleep
-    Thread.sleep(Duration.ofSeconds(2).toMillis());
+    while(!sem.tryAcquire()) {
+      System.out.println("waiting for leaderElectionWorker to throw exception in LeaderElectionTest::testLeaderElectionCaptureException");
+    }
+    
     assertEquals(expectedException, actualException.get().getCause());
   }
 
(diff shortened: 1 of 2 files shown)


karthikkondapally:
@brendandburns @yue9944882

extended/src/test/java/io/kubernetes/client/extended/leaderelection/LeaderElectionTest.java
```diff
@@ -285,16 +285,26 @@ public void testLeaderElectionCaptureException() throws ApiException, Interrupte
...
 
     ExecutorService leaderElectionWorker = Executors.newFixedThreadPool(1);
+    sem.acquire();
     leaderElectionWorker.submit(
         () -> {
           leaderElector.run(() -> {}, () -> {});
         });
-    // TODO: Remove this sleep
-    Thread.sleep(Duration.ofSeconds(2).toMillis());
+    while (!sem.tryAcquire()) {
```
brendandburns, line 303: Is there a reason for this loop? I would remove all this code and just do:

`sem.acquire()` like you did above.

util/src/test/java/io/kubernetes/client/ExecTest.java
```diff
@@ -110,14 +113,17 @@ public void testExecProcess() throws IOException, InterruptedException {
...
+    final Semaphore sem = new Semaphore(2);
+    sem.acquire(2);
+    Thread t1 = asyncCopy(process.getInputStream(), stdout, sem);
+    Thread t2 = asyncCopy(process.getErrorStream(), stderr, sem);
 
     process.getHandler().bytesMessage(makeStream(3, OUTPUT_EXIT0.getBytes(StandardCharsets.UTF_8)));
-    // TODO: Fix this asap!
-    Thread.sleep(1000);
 
+    while (!sem.tryAcquire(2)) {
```
brendandburns, line 123: As above, why not just `sem.acquire(2)`?

brendandburns:
Thanks for the PR! Couple of small comments then I think this is ready to merge.

extended/src/test/java/io/kubernetes/client/extended/leaderelection/LeaderElectionTest.java
karthikkondapally: i wanted to print message on why we are waiting.. 
but i have changed to it to countdown latch, as i thought it would be better  than sempahore in these situations

util/src/test/java/io/kubernetes/client/ExecTest.java
karthikkondapally: wanted to print message on why we are waiting..
but i have changed semaphores to countdown latch, as i thought it would be better than sempahore in these situations

extended/src/test/java/io/kubernetes/client/extended/workqueue/DefaultDelayingQueueTest.java
```diff
@@ -103,7 +103,6 @@ private boolean waitForAdded(DefaultDelayingQueue queue, int size) {
   }
 
   private boolean waitForWaitingQueueToFill(DefaultDelayingQueue queue) {
-    return Wait.poll(
```
brendandburns, line 106: I don't think these are the same test. (and it seems unrelated) can you revert this change?

karthikkondapally: changed it back...

brendandburns:
/lgtm
/approve

Thanks for the PR!

Issue #1659, opened 2021-04-20
[BUG] ModelMapper cannot register any resource classes if packaged in executable fat jar


yue9944882:
**Describe the bug**

if the project is packaged in a whole executable fat jar, e.g:

```shell
$ jar -tf my-project-executable.jar
META-INF/
META-INF/MANIFEST.MF
org/
org/springframework/
org/springframework/boot/
...
BOOT-INF/lib/gson-fire-1.8.5.jar
BOOT-INF/lib/joda-time-2.10.5.jar
BOOT-INF/lib/joda-convert-2.2.1.jar
BOOT-INF/lib/client-java-proto-11.0.1.jar
BOOT-INF/lib/client-java-11.0.1.jar  # NESTED JAR FILE
...
```

the classes under `io.kubernetes.client.openapi.model` will be packaged as a nested jar file inside the fat jar file. when we're trying [scanning the classes](https://github.com/kubernetes-client/java/blob/7e2bf8e0f0c7d5b3b47aa023748bae1e6e1ef711/util/src/main/java/io/kubernetes/client/util/ModelMapper.java#L336-L337) via the class-loader. the URL for the pre-built resource classes will be e.g.:

> file:/home/admin/release/run/target/boot/agent-installer-bootstrap-0.0.1-SNAPSHOT-executable.jar!/BOOT-INF/lib/client-java-api-11.0.1.jar!/io/kubernetes/client/openapi/models

then we will get a wrong URL to the jar file when we're [extracting](https://github.com/kubernetes-client/java/blob/7e2bf8e0f0c7d5b3b47aa023748bae1e6e1ef711/util/src/main/java/io/kubernetes/client/util/ModelMapper.java#L414) file path from the resource path above:

> /home/admin/release/run/target/boot/agent-installer-bootstrap-0.0.1-SNAPSHOT-executable.jar!/BOOT-INF/lib/client-java-api-11.0.1.jar

b/c we cant correctly open the actual jar file using the file path above, no classes will be found in the end.. one feasible solution i can think of is to visit the nested jar file recursively if multiple `!` is found in the URL.

** Client Version **
e.g. `11.0.1

pgalache:
Any advance with this?

yue9944882:
/remove-lifecycle rotten

@pgalache we can workaround by manually reigsterring resource types to ModelMapper, does that work for you?

pgalache:
Yes, that's what I did as a workaround.

[...]

PR #1805, opened 2021-07-29
Chore: Warn if no kubernetes api models were loaded

yue9944882:
this is a temporary remediation for https://github.com/kubernetes-client/java/issues/1659


pgalache:
How does this remediate the problem of loading the resource api model in fatjars?

SxunS:
> How does this remediate the problem of loading the resource api model in fatjars?

The same question.

Issue #1806, opened 2021-07-29
Missing Kubectl scale block for StatefulSets


Paleloser:
**Describe the bug**
At the moment there's no chance of scaling a `V1StatefulSet` using `Kubectl#scale(Class<ApiType> apiTypeClass)`.

I've been digging into the internals of this method and it looks like it generates a `KubectlScale`, which can be executed later on. The _problem_ is that this class' `execute()` method does not include a specific block for scaling a `V1StatefulSet`:

```java
public ApiType execute() throws KubectlException {
  this.validate();
  String jsonPatchStr = String.format("[{\"op\":\"replace\",\"path\":\"/spec/replicas\",\"value\":%d}]", this.replicas);
  AppsV1Api api = new AppsV1Api(this.apiClient);

  try {
    if (this.apiTypeClass.equals(V1Deployment.class)) {
      return (KubernetesObject)PatchUtils.patch(this.apiTypeClass, () -> {
        return api.patchNamespacedDeploymentCall(this.name, this.namespace, new V1Patch(jsonPatchStr), (String)null, (String)null, (String)null, (Boolean)null, (ApiCallback)null);
      }, "application/json-patch+json", this.apiClient);
    } else if (this.apiTypeClass.equals(V1ReplicaSet.class)) {
      return (KubernetesObject)PatchUtils.patch(this.apiTypeClass, () -> {
        return api.patchNamespacedReplicaSetCall(this.name, this.namespace, new V1Patch(jsonPatchStr), (String)null, (String)null, (String)null, (Boolean)null, (ApiCallback)null);
      }, "application/json-patch+json", this.apiClient);
    } else {
      throw new KubectlException("Unsupported class for scale: " + this.apiTypeClass);
    }
  } catch (ApiException var4) {
    throw new KubectlException(var4);
  }
}
```

**Client Version**
io.kubernetes:client-java-extended:12.0.1

**Kubernetes Version**
1.20.2

**Java Version**
openjdk version "16" 2021-03-16
OpenJDK Runtime Environment AdoptOpenJDK (build 16+36)
Eclipse OpenJ9 VM AdoptOpenJDK (build openj9-0.25.0, JRE 16 Mac OS X amd64-64-Bit Compressed References 20210311_66 (JIT enabled, AOT enabled)

**To Reproduce**
Having already instantiated an `ApiClient` named `client` and a StatefulSet with name "name", running the following code will result into a `KubectlException`:

```java
[...]

PR #1809, opened 2021-07-30
Fix/scale for stateful sets

NeverRaR:
fixed :#1806 
this pull let Kubectl scale support StatefulSets


brendandburns:
/lgtm
/approve

Thanks!

recena:
I would like to point out that these changes come from @Paleloser, the person who reported the issue and proposed a fix.

Issue #1664, opened 2021-04-21
Proper way to connect to a GKE cluster in production.


ginokurian87:
**Describe the bug**
I've been running around in circles trying to connect to GKE and get it working properly. Here are the steps that i did for the same.
I'm trying to connect to a GKE cluster from outside of the GCP env. For this i created a service account and downloaded the key.
Set the environment variable ```GOOGLE_APPLICATION_CREDENTIALS```  pointing to the above key.
Also ran the ```gcloud container clusters get-credentials.........``` command locally to generate the config files.
Then wrote a sample java app based in the examples like below:

```
   GoogleCredentials.getApplicationDefault(); //so that the token is refreshed automatically
        KubeConfig.registerAuthenticator(new GCPAuthenticator());
        ApiClient client = Config.defaultClient();
        // if you prefer not to refresh service account token, please use:
        // ApiClient client = ClientBuilder.oldCluster().build();
        Configuration.setDefaultApiClient(client);
        CoreV1Api api = new CoreV1Api();
        V1PodList list =
                api.listPodForAllNamespaces(null, null, null, null, null, null, null, null, null, null);
        for (V1Pod item : list.getItems()) {
            System.out.println(item.getMetadata().getName());
        }

    }
```
When i run the code, for the first time it works well. However after(persumably!) the token refresh, i get the following error :

```
Exception in thread "main" java.lang.IllegalStateException: Unimplemented
	at io.kubernetes.client.util.authenticators.GCPAuthenticator.refresh(GCPAuthenticator.java:61)
	at io.kubernetes.client.util.KubeConfig.getAccessToken(KubeConfig.java:215)
	at io.kubernetes.client.util.credentials.KubeconfigAuthentication.<init>(KubeconfigAuthentication.java:57)
	at io.kubernetes.client.util.ClientBuilder.kubeconfig(ClientBuilder.java:297)
	at io.kubernetes.client.util.ClientBuilder.getClientBuilder(ClientBuilder.java:129)
	at io.kubernetes.client.util.ClientBuilder.standard(ClientBuilder.java:105)
	at io.kubernetes.client.util.ClientBuilder.standard(ClientBuilder.java:100)
```

It works If i do any kubectl operation in my local and then retry.  Is this because of [ISSUE-290](https://github.com/kubernetes-client/java/issues/290)?

All of the above hacks works in the local system, but whet would be the correct approach in prod? Do we have a err...less hacky way of connecting and staying connected to the GKE cluster ?

[...]

PR #1810, opened 2021-08-02
support gcp authentication refresh token

jhbae200:
Related Issues: 
- https://github.com/kubernetes-client/java/issues/290
- https://github.com/kubernetes-client/java/issues/1664

I wrote a "refresh" function because it seems to be using "cmd-path", "cmd-args", "expiry-key", "token-key" to refresh the token.
Need to check that it works on various os platforms.

### DIFF
--- a/util/src/main/java/io/kubernetes/client/util/authenticators/GCPAuthenticator.java
+++ b/util/src/main/java/io/kubernetes/client/util/authenticators/GCPAuthenticator.java
@@ -68,6 +68,10 @@ public class GCPAuthenticator implements Authenticator {
     return (expiry != null && expiry.compareTo(Instant.now()) <= 0);
   }
 
+  public ProcessBuilder getPb() {
+    return new ProcessBuilder();
+  }
+
   @Override
   public Map<String, Object> refresh(Map<String, Object> config) {
     if (!config.containsKey(CMD_ARGS) || !config.containsKey(CMD_PATH))
@@ -76,7 +80,7 @@ public class GCPAuthenticator implements Authenticator {
     String cmdArgs = (String) config.get(CMD_ARGS);
     String fullCmd = cmdPath + cmdArgs;
     try {
-      Process process = new ProcessBuilder().command(Arrays.asList(fullCmd.split(" "))).start();
+      Process process = this.getPb().command(Arrays.asList(fullCmd.split(" "))).start();
       process.waitFor(10, TimeUnit.SECONDS);
       if (process.exitValue() != 0) {
         String stdErr = IOUtils.toString(process.getErrorStream(), StandardCharsets.UTF_8);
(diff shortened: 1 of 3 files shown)


yue9944882:
@jhbae200 thank you for kicking this off! can you resign your commit by properly setting up your git workspace?

```bash
git config user.name ..
git config user.email ...
```

util/src/main/java/io/kubernetes/client/util/authenticators/GCPAuthenticator.java
```diff
@@ -12,10 +12,16 @@
...
 
+import com.google.gson.Gson;
 import io.kubernetes.client.util.KubeConfig;
+
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import java.time.Instant;
-import java.util.Date;
-import java.util.Map;
+import java.util.*;
```
yue9944882, line 21: wildcard import will fail the lint checker

yue9944882 (commented):
@jhbae200 can you run `mvn spotless:apply` before submitting the changes? thanks!

util/src/main/java/io/kubernetes/client/util/authenticators/GCPAuthenticator.java
```diff
@@ -58,6 +70,27 @@ public boolean isExpired(Map<String, Object> config) {
...
+      throw new RuntimeException("Could not refresh token");
+    String cmd_path= (String) config.get(CMD_PATH);
+    String cmd_args = (String) config.get(CMD_ARGS);
+    List<String> commandList = new ArrayList<>(Arrays.asList(cmd_args.split(" ")));
+    commandList.add(0, cmd_path);
+    try {
+      Process process = new ProcessBuilder().command(commandList).start();
+      process.waitFor(10, TimeUnit.SECONDS);
+      if(process.exitValue() != 0) {
+        throw new IOException("Process exit code: "+process.exitValue());
```
brendandburns, line 83: I'm not sure that IOException is right here. `IllegalStateException(...)`?

brendandburns: Also, please make the message more descriptive and include the stderr to make the exception more debuggable.

`throw new IllegalStateException("Failed to get token (" + process.exitValue() + ") " + stderr`

brendandburns:
Some small comments. Generally looks good. Needs linter formatting to conform to the style guide.

jhbae200:
Updated the code. am i right?

jhbae200:
I think need to mock the ProcessBuilder to test the refresh method.
I'm trying to mock that function by adding a getProcessBuilder method to this class. What do you think?

```
public class GCPAuthenticator implements Authenticator {
//somthing code
  public ProcessBuilder getPb() {
    return new ProcessBuilder();
  }

  @Override
  public Map<String, Object> refresh(Map<String, Object> config) {
    //somthing code
      Process process = this.getPb().command(Arrays.asList(fullCmd.split(" "))).start();
    //somthing code
  }
}
```

brendandburns:
Thanks for adding the test!

I think it is simpler if you just make the ProcessBuilder an optional constructor parameter, e.g.

```java
public GCPAuthenticator() {
   this(new ProcessBuilder());
}

public GCPAuthenticator(ProcessBuilder pb) {
   this.pb = pb;
}
....
```

piotom:
Could you tell me when it will be released?

keith-miller:
It took me a while to discover that this existed and was released. Any chance this can be added to the documentation for those coming after me? 

Thanks

PR #1812, opened 2021-08-03
feat : add a selector util for tranforming V1LabelSelector to LabelSelector

NeverRaR:


extended/src/main/java/io/kubernetes/client/extended/kubectl/util/selector/SelectorHelper.java
```diff
@@ -0,0 +1,68 @@
...
+  protected static final String LabelSelectorOpDoesNotExist = "DoesNotExist";
+
+  public static LabelSelector LabelSelectorAsSelector(V1LabelSelector ps) throws Exception {
+    int matcherNum = 0;
+    if (ps.getMatchLabels() != null) matcherNum = ps.getMatchLabels().size();
+    if (ps.getMatchExpressions() != null) matcherNum = ps.getMatchExpressions().size();
+
+    if (matcherNum == 0) {
+      return LabelSelector.empty();
+    }
```
yue9944882, line 38: if i'm reading correctly, it doesnt make a difference if L32-38 are removed, is it?

extended/src/main/java/io/kubernetes/client/extended/kubectl/util/selector/SelectorHelper.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.List;
+import java.util.Map;
+
+public class SelectorHelper {
+  protected static final String LabelSelectorOpIn = "In";
+  protected static final String LabelSelectorOpNotIn = "NotIn";
+  protected static final String LabelSelectorOpExists = "Exists";
+  protected static final String LabelSelectorOpDoesNotExist = "DoesNotExist";
+
+  public static LabelSelector LabelSelectorAsSelector(V1LabelSelector ps) throws Exception {
```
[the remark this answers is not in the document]
yue9944882: the method is required to be camel-cased in convention and i prefer a simpler name such as `parse(..)` or `load(...)`. additionally i'd like to see legit javadocs clarifying the behavior of those public methods. 

```
Extracting a working label selector for matching certain labels from the api model {@link V1LabelSelector}.
```

extended/src/main/java/io/kubernetes/client/extended/kubectl/util/selector/SelectorHelper.java
```diff
@@ -0,0 +1,68 @@
...
+            matchers.add(SetMatcher.notIn(expr.getKey(), expr.getValues().toArray(new String[0])));
+            break;
+          case LabelSelectorOpExists:
+            matchers.add(SetMatcher.exists(expr.getKey()));
+            break;
+          case LabelSelectorOpDoesNotExist:
+            matchers.add(SetMatcher.notExists(expr.getKey()));
+            break;
+          default:
+            throw new Exception(expr.getOperator() + " is not a valid pod selector operator");
```
yue9944882, line 62: throws new IllegalArgumentException(..)

extended/src/main/java/io/kubernetes/client/extended/kubectl/util/selector/SelectorHelper.java
```diff
@@ -0,0 +1,68 @@
...
+import io.kubernetes.client.openapi.models.V1LabelSelectorRequirement;
+import io.kubernetes.client.util.labels.EqualityMatcher;
+import io.kubernetes.client.util.labels.LabelMatcher;
+import io.kubernetes.client.util.labels.LabelSelector;
+import io.kubernetes.client.util.labels.SetMatcher;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+public class SelectorHelper {
```
yue9944882, line 25: wdyt if we move this helper into the existing static helper class `LabelSelector`?

yue9944882 (commented):
thanks! given the fact that [all the conditions in the label selectors are AND'ed](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L1113-L1126), i think flattening both the labels and expressions in a row should work. but i'd like to see separate test into the following cases:

- parsing a `V1LabelSelector` w/ merely label-condition should work as expected
- parsing a `V1LabelSelector` w/ merely expression-condition should work as expected
- parsing a `V1LabelSelector` w/ malformed content should throw an unchecked exception

util/src/main/java/io/kubernetes/client/util/labels/LabelSelector.java
```diff
@@ -46,6 +50,52 @@ public static LabelSelector empty() {
...
+          default:
+            throw new IllegalArgumentException(
+                expr.getOperator() + " is not a valid pod selector operator");
+        }
+      }
+    }
+    return LabelSelector.and(matchers.toArray(matchers.toArray(new LabelMatcher[0])));
+  }
+
+  public static final String LABEL_SELECTOR_OP_IN = "In";
```
yue9944882, line 94: nit: move the public constants to the top

yue9944882 (commented):
last nit otherwise lgtm, thanks!

yue9944882 (commented):
/lgtm
/approve

thanks!

PR #1823, opened 2021-08-15
Optimize DeltaFIFO.combineDeltas

dddddai:
There is no need to allocate a new list when removing the duplicate delta, it's more efficient to modify the list directly.
see https://github.com/kubernetes/kubernetes/commit/238fce9357634502badbcc3704b0655cf8c378c3

Besides, I removed some unnecessary casting.


yue9944882 (commented):
/lgtm
/approve

@dddddai thanks!

Issue #1826, opened 2021-08-17
WebSocketStreamHandler.write() : NegativeArraySizeException when calculating 'remaining' byte count


PredatorVI:
Kubernetes Java Client Version:  13.0.0

In WebSocketStreamHandler.java (line 273), the 'remaining' byte count value is incremented using the cumulative bytesWritten instead of the bufferSize for that iteration causing a java.lang.NegativeArraySizeException.  This issue is likely only seen when copying files larger that 3 times the buffer size (45 MiB).

For testing purposes I changed the min buffer size to 1024 and attempted to copy 4096 bytes.

Initial values:  bytesWritten = 0; remaining = 4096;
Iteration 1:  bytesWritten = 1024; remaining = 3072;  //This is OK.
Iteration 2:  bytesWritten = 2048; remaining = 1024;  //Should be 2048
Iteration 3: bytesWritten = 3072; remaining = -2048; //Should be 1024
Iteration 4: **java.lang.NegativeArraySizeException**

The fix I believe is to decrement remaining by bufferSize:

`        remaining -= bufferSize;`

Issue #1822, opened 2021-08-13
Copy.copyFileToPod() hangs waiting for process to complete


PredatorVI:
Client Version: 13.0.0 
Kubernetes Version:  1.19.12-gke.2100
Java Version:  Java 1.8.0_291
**Server
 - OS: Ubuntu 20.04.1 LTS
 - Environment: GKE Container 
 - Cloud :  GCP/GKE

I have the copyFileFromPod() working, but copyFileToPod() hangs at Copy.java line 459 at proc.waitFor().  I've pulled source, but I don't understand the websockets well enough to know what is going on.  A copy command from the command-line works fine.

I've verified that the 'tar' and 'base64' executables exist in the container:

```
> root@test-gcp-dev-gke-guse4a-0:/tmp# which tar
> /usr/bin/tar
> root@test-gcp-dev-gke-guse4a-0:/tmp# which base64
> /usr/bin/base64
> root@test-gcp-dev-gke-guse4a-0:/tmp# cat /etc/issue
> Ubuntu 20.04.1 LTS \n \l
> 
```

I've not changed the timeouts as I'd expect a simple text file copy would work within the defaults.  

It never seems to timeout or throw an error so I don't have any stack traces.  

Also, I tried API version 12.0.1 (throws SocketTimeout almost immediately) and 11.0.2 (hangs just like 13.0.0).

brendandburns:
For some reason the untar command on the other side isn't completing. It's hard to know why. I don't think that this is a WebSockets issue, I think it is an issue with the process in the container that is unpacking the copied files (or alternately it is an issue with the input data not getting closed properly)

How are you supplying the data that is getting copied? Byte array? Local file path? Is it one file or many?

I will try to repro locally.

PredatorVI:
I've started testing using a simple text file.

The test I'm doing is to first call copyFileFromPod() to grab a config file from the pod (POD:/etc/adduser.conf --> LOCAL:C:/tmp/adduser.conf).  I then turn around and copy that same file using copyFileToPod() (LOCAL:C:/tmp/adduser.conf --> POD:/tmp/adduser.conf).

I am using the java.nio.file.Path variant of the method:

**public void copyFileToPod(String namespace, String pod, String container, Path srcPath, Path destPath)
    throws ApiException, IOException**

I open to suggestions on how to narrow this down more or debug it.  

Thanks!!

PredatorVI:
[...]

PR #1834, opened 2021-08-19
Updated WebSocketStreamHandler.write() to properly calculate 'remaining' bytes

PredatorVI:
Let's try this again! :) 

Fix for #1826: WebSocketStreamHandler.write() to calculate 'remaining' bytes using bufferSize from the current loop iteration instead of the cumulative bytesWritten.

Related to #1822:  Removed the base64 encode/decode steps/streams per comment https://github.com/kubernetes-client/java/issues/1822#issuecomment-900315846 as unnecessary steps to simplify the process.

Also the execCopyToPod() method converts the destPath to a string, but if the client is windows, it sends the path to the 'sh -c **` command using Windows path delimiters.  I added a line to replace any backslashes with forward slashes.


PredatorVI:
Messed up again.

PR #1835, opened 2021-08-19
Updated WebSocketStreamHandler.write() to calculate 'remaining' bytes properly : Fixed destPath normalization for potential Windows delimiters

PredatorVI:
Third time's the charm....

- Fix for #1826: Updated WebSocketStreamHandler.write() to calculate 'remaining' bytes using bufferSize from the current loop iteration instead of the cummulative bytesWritten.

- Related to #1822:  Removed the base64 encode/decode steps/streams per comment at https://github.com/kubernetes-client/java/issues/1822#issuecomment-900315846

- Related to #1822:  Added path normalization to convert potential windows path delimiters to linux delimiters.  The copyFileToPod() variants take a java Path instance for both source and destination.  Since the destination is Linux, when I set destPath = Paths.get("/tmp/foo.txt");, converting that to a string to send to the tar command ends up sending "\tmp" as the parentPath which is not valid for linux.

### DIFF
--- a/util/src/main/java/io/kubernetes/client/Copy.java
+++ b/util/src/main/java/io/kubernetes/client/Copy.java
@@ -34,7 +34,6 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import org.apache.commons.codec.binary.Base64InputStream;
-import org.apache.commons.codec.binary.Base64OutputStream;
 import org.apache.commons.compress.archivers.ArchiveEntry;
 import org.apache.commons.compress.archivers.ArchiveInputStream;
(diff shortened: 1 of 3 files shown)


yue9944882 (commented):
/lgtm

@PredatorVI thank you for the detailed investigation! @brendandburns has the approve

yue9944882:
@PredatorVI can you take a look at the test failure?

PredatorVI:
I believe the Unit Test fix is checked in and should work.  On my windows system I see multiple other failures related to Unix vs Windows path handling or line separators which I didn't address.

yue9944882 (commented):
/lgtm

yue9944882:
getting this in! 

/approve

PR #1849, opened 2021-08-25
Kill the sleep in AttachTest

dddddai:
`InputStream.read()` blocks until input data is available, which makes sure the connection has been established.


yue9944882 (commented):
thanks!

/lgtm
/approve

PR #1851, opened 2021-08-25
Cherry-pick(release-13): Stale watch event should be handled properly

yue9944882:
picking https://github.com/kubernetes-client/java/pull/1832 to release-13

cc @brendandburns @haoming-db 

i will bump a patch version after this gets into 13 branch

### DIFF
--- a/util/src/main/java/io/kubernetes/client/informer/cache/ReflectorRunnable.java
+++ b/util/src/main/java/io/kubernetes/client/informer/cache/ReflectorRunnable.java
@@ -16,6 +16,7 @@ import io.kubernetes.client.common.KubernetesListObject;
 import io.kubernetes.client.common.KubernetesObject;
 import io.kubernetes.client.informer.EventType;
 import io.kubernetes.client.informer.ListerWatcher;
+import io.kubernetes.client.informer.exception.WatchExpiredException;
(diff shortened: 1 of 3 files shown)


util/src/main/java/io/kubernetes/client/informer/cache/ReflectorRunnable.java
```diff
@@ -160,6 +163,8 @@ public void run() {
           closeWatch();
         }
       }
+    } catch (WatchExpiredException e) {
```
haoming-db, line 166: It looks like this code block is unreachable?
`WatchExpiredException` is thrown at line 134 and then caught by line 135, but is never re-thrown.
It hits the `return;` at line 161, which fixes the bug. However, this one looks unused, unless the `exceptionHandler` re-throws the exception.
Please correct me in case I'm wrong.

yue9944882: @haoming-db oh oversight! you're right, the `catch (WatchExpiredException e)` block isnt actually taking effect. but in the `catch(Throwable)` block the thrown `WatchExpiredException` will be finally reaching the `return` statement there. i just opened https://github.com/kubernetes-client/java/pull/1861 and picked it to this pull together. thanks!

PR #1855, opened 2021-08-26
Feat/kubectl rollout undo statefulset

NeverRaR:
feat : support kubectl rollout undo for statefulset

### DIFF
--- a/extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlRollout.java
+++ b/extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlRollout.java
@@ -18,6 +18,7 @@ import io.kubernetes.client.extended.kubectl.exception.KubectlException;
 import io.kubernetes.client.extended.kubectl.util.deployment.DeploymentHelper;
 import io.kubernetes.client.openapi.ApiClient;
 import io.kubernetes.client.openapi.ApiException;
+import io.kubernetes.client.openapi.JSON;
 import io.kubernetes.client.openapi.apis.AppsV1Api;
 import io.kubernetes.client.openapi.models.V1ControllerRevision;
 import io.kubernetes.client.openapi.models.V1ControllerRevisionList;
@@ -29,6 +30,7 @@ import io.kubernetes.client.openapi.models.V1ReplicaSet;
 import io.kubernetes.client.openapi.models.V1StatefulSet;
 import io.kubernetes.client.util.labels.LabelSelector;
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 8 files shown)


extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlRollout.java
```diff
@@ -308,4 +330,241 @@ public V1PodTemplateSpec execute() throws KubectlException {
       }
     }
   }
+
+  class KubectlRolloutUndo extends Kubectl.ResourceBuilder<ApiType, KubectlRolloutUndo>
```
brendandburns, line 334: `static`

extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlRollout.java
```diff
@@ -308,4 +330,241 @@ public V1PodTemplateSpec execute() throws KubectlException {
...
+      implements Kubectl.Executable<ApiType> {
+
+    private final Set<String> annotationsToSkip =
```
brendandburns, line 337: `static`

extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlRollout.java
```diff
@@ -308,4 +330,241 @@ public V1PodTemplateSpec execute() throws KubectlException {
...
+          .getTemplate()
+          .getMetadata()
+          .getLabels()
+          .remove(DeploymentHelper.DEFAULT_DEPLOYMENT_UNIQUE_LABEL_KEY);
+      Map<String, String> annotations = new HashMap<>();
+
+      Map<String, String> deploymentAnnotations = deployment.getMetadata().getAnnotations();
+      Map<String, String> rsAnnotations = rsForRevision.getMetadata().getAnnotations();
+      for (String key : annotationsToSkip) {
+        if (deploymentAnnotations == null) {
```
brendandburns, line 423: Move this check outside of the `for` loop:

```java
if (deploymentAnnotations != null) {
   for ....
}
```

Similar to the one below.

extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlRollout.java
```diff
@@ -308,4 +330,241 @@ public V1PodTemplateSpec execute() throws KubectlException {
...
+
+      rsForRevision
+          .getSpec()
+          .getTemplate()
+          .getMetadata()
+          .getLabels()
+          .remove(DeploymentHelper.DEFAULT_DEPLOYMENT_UNIQUE_LABEL_KEY);
+      Map<String, String> annotations = new HashMap<>();
+
+      Map<String, String> deploymentAnnotations = deployment.getMetadata().getAnnotations();
```
brendandburns, line 420: I feel like this algorithm is complicated. It's easier to do:

```java

Map<String, String> annotations = new Map<>();
annotations.addAll(deployment.getMetadata().getAnnotations());
annotations.addAll(rsForRevision.getMetadata().getAnnotations());
for (String key : annotationsToSkip.keySet()) {
   annotations.remove(key);
}
```

NeverRaR: I don't think the two  algorithms are equivalent.For keys in "annotationsToSkip", we should keep key/value in "annotations" consistent with deployment.getMetadata().getAnnotations() instead of removing them.

extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlRollout.java
NeverRaR: We need to access the apiTypeClass of the external class,so class KubectlRolloutUndo can't be static.

extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlRollout.java
brendandburns: Sorry, I mis-read the algorithm. How about 

```java
Map<String, String> annotations = new Map<>();
annotations.addAll(rsForRevision.getMetadata().getAnnotations());
for (String key : annotationsToSkip.keySet()) {
   if (deploymentAnnotations.contains(key)) {
      annotations.put(key, deploymentAnnotations.get(key));
   } else {
     annotations.remove(key);
   }
}
```

I think that is equivalent (add everything from rs, except the ones to skip)

extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlRollout.java
```diff
@@ -308,4 +330,241 @@ public V1PodTemplateSpec execute() throws KubectlException {
...
+          if (latestRevision < revision) {
+            previousRevision = latestRevision;
+            previousReplicaSet = latestReplicaSet;
+            latestRevision = revision;
+            latestReplicaSet = rs;
+          } else if (previousRevision < revision) {
+            previousRevision = revision;
+            previousReplicaSet = rs;
+          }
+        } else if (toRevision == revision) {
```
brendandburns, line 492: Move this check up to the top?

```java
for (...) {
   long revision = ...
   if (toRevision == revision) {
      return rs;
   }
   ...
}
```

I think that's clearer

brendandburns (changes requested):
A few more comments.

Issue #1843, opened 2021-08-23
PagerTest.testPaginationForNamespaceListWithSuccessThreadSafely() is flaky


NeverRaR:
```
com.github.tomakehurst.wiremock.client.VerificationException: Expected exactly 20 requests matching the following pattern but received 21:
{
  "urlPath" : "/api/v1/namespaces",
  "method" : "GET",
  "queryParameters" : {
    "limit" : {
      "equalTo" : "1"
    }
  }
}

	at com.github.tomakehurst.wiremock.client.WireMock.verifyThat(WireMock.java:538)
	at com.github.tomakehurst.wiremock.client.WireMock.verifyThat(WireMock.java:521)
	at com.github.tomakehurst.wiremock.client.WireMock.verify(WireMock.java:559)
	at io.kubernetes.client.extended.pager.PagerTest.testPaginationForNamespaceListWithSuccessThreadSafely(PagerTest.java:166)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at com.github.tomakehurst.wiremock.junit.WireMockRule$1.evaluate(WireMockRule.java:79)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
[...]

PR #1863, opened 2021-08-31
Bugfix: Move Pager's concurrently read fields to nested class

yue9944882:
Fixes https://github.com/kubernetes-client/java/issues/1843#issuecomment-907548421

cc @NeverRaR  @dddddai @brendandburns


NeverRaR:
lgtm

PR #1865, opened 2021-09-01
Kill some hard-coded sleep in tests

dddddai:
The same as #1849

### DIFF
--- a/util/src/test/java/io/kubernetes/client/CopyTest.java
+++ b/util/src/test/java/io/kubernetes/client/CopyTest.java
@@ -29,6 +29,7 @@ import io.kubernetes.client.util.ClientBuilder;
 import io.kubernetes.client.util.exception.CopyNotSupportedException;
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
 import java.nio.file.Paths;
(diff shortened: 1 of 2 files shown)


util/src/test/java/io/kubernetes/client/CopyTest.java
```diff
@@ -69,13 +70,14 @@ public void testUrl() throws IOException, ApiException, InterruptedException {
                     .withBody("{}")));
 
     try {
-      copy.copyFileFromPod(pod, "container", "/some/path/to/file");
+      InputStream inputStream = copy.copyFileFromPod(pod, "container", "/some/path/to/file");
+      // block until the connection is established
+      inputStream.read();
```
aaruna, line 75: consider adding a timeout to fail the test in case the connection could not be established

dddddai: Thanks for your reply, I suppose if the wiremock connection could not be established, it would throw an exception and fail.
Maybe we could get this one merged and consider adding timeout in another PR.

yue9944882 (commented):
/lgtm
/approve

thanks and apologies for the delay!

PR #1875, opened 2021-09-09
Test: Fixes constant maven download failure by escalating mvn permission

yue9944882:


brendandburns:
This is the same as #1876 which do we want to merge?

Also, I'm wondering why this requires `sudo`?

yue9944882:
closing mine in favor of #1876

/close

PR #1880, opened 2021-09-10
Tweak call for CoreV1 group properly

### LINKED ISSUE #1838: KubectlApply doesn't work with some Rancher clusters.
**Describe the bug**
  I am trying to create an service using Kubectl.apply as per this: https://github.com/kubernetes-client/java/blob/b0c91798c63203f20dc8f2674197e772b2c187ee/docs/kubectl-equivalence-in-java.md#kubectl-apply

  source code:
  ```
  KubectlApply kubectlApply = Kubectl.apply(V1Service.class).fieldManager("kubectl").resource(resource);
  KubectlApply kubectlApplyClient = kubectlApply.apiClient(apiClient);
  kubectlApplyClient.execute();
  ```

  I am getting below error:

  ```
      Caused by: com.google.gson.JsonSyntaxException: java.lang.IllegalStateException: Not a JSON Object: 404
      at com.google.gson.Gson.fromJson(Gson.java:944)
      at com.google.gson.Gson.fromJson(Gson.java:897)
      at com.google.gson.Gson.fromJson(Gson.java:846)
      at io.kubernetes.client.openapi.JSON.deserialize(JSON.java:169)
      at io.kubernetes.client.util.generic.GenericKubernetesApi.patch(GenericKubernetesApi.java:632)
      at io.kubernetes.client.extended.kubectl.KubectlApply.executeServerSideApply(KubectlApply.java:89)
      at io.kubernetes.client.extended.kubectl.KubectlApply.execute(KubectlApply.java:67)
      at chengf.falcon.k8s.learn.ManifestTest.applyManifest(ManifestTest.java:65)
      ... 23 more
      Caused by: java.lang.IllegalStateException: Not a JSON Object: 404
      at com.google.gson.JsonElement.getAsJsonObject(JsonElement.java:91)
      at io.kubernetes.client.gson.V1StatusPreProcessor.isLegitKubernetesStatus(V1StatusPreProcessor.java:39)
      at io.kubernetes.client.gson.V1StatusPreProcessor.preDeserialize(V1StatusPreProcessor.java:26)
      at io.gsonfire.gson.HooksTypeAdapter.runPreDeserialize(HooksTypeAdapter.java:79)
      at io.gsonfire.gson.HooksTypeAdapter.read(HooksTypeAdapter.java:53)
      at com.google.gson.Gson.fromJson(Gson.java:932)
      ... 30 more
  ```

**Client Version**
I've tried the following version：

`11.0.0`、`11.0.2`、`13.0.0`

**Kubernetes Version**
`v1.19.0`

**Java Version**
Java 8

**service.yaml**
```
apiVersion: v1
kind: Service
metadata:
  name: nginx-test
  namespace: appstore
  labels:
    app: nginx-test
spec:
  ports:
    - port: 80
      name: http
      targetPort: 80
      protocol: TCP
  selector:
    app: nginx-test

```

I tried the same resource with kubectl CLI and I don't see any error:
```
kubectl apply --server-side=true --field-manager=kubectl -f service.yaml
```

dddddai:
Fix #1838

If `basePath` contains multiple segments (like using java client with rancher clusters), it would tweak the call incorrectly and break the request url, see https://github.com/kubernetes-client/java/issues/1838#issuecomment-913290787

So it should calculate the segment indexes to be modified instead of hardcoding them to 0 and 1.

### DIFF
--- a/extended/src/test/java/io/kubernetes/client/extended/kubectl/KubectlApplyTest.java
+++ b/extended/src/test/java/io/kubernetes/client/extended/kubectl/KubectlApplyTest.java
@@ -12,9 +12,12 @@ limitations under the License.
 */
 package io.kubernetes.client.extended.kubectl;
 
-import static com.github.tomakehurst.wiremock.client.WireMock.*;
+import static com.github.tomakehurst.wiremock.client.WireMock.aResponse;
+import static com.github.tomakehurst.wiremock.client.WireMock.get;
+import static com.github.tomakehurst.wiremock.client.WireMock.patch;
+import static com.github.tomakehurst.wiremock.client.WireMock.patchRequestedFor;
 import static com.github.tomakehurst.wiremock.client.WireMock.urlPathEqualTo;
-import static org.junit.Assert.*;
+import static org.junit.Assert.assertNotNull;
 
 import com.github.tomakehurst.wiremock.junit.WireMockRule;
 import com.github.tomakehurst.wiremock.matching.EqualToPattern;
--- a/util/src/main/java/io/kubernetes/client/util/generic/GenericKubernetesApi.java
+++ b/util/src/main/java/io/kubernetes/client/util/generic/GenericKubernetesApi.java
@@ -824,7 +824,17 @@ public class GenericKubernetesApi<
       return call;
     }
     HttpUrl url = call.request().url();
-    HttpUrl tweakedUrl = url.newBuilder().removePathSegment(1).setPathSegment(0, "api").build();
+    String basePath = customObjectsApi.getApiClient().getBasePath();
+    // count segments in the basePath
+    int offset = 0;
+    for (int i = basePath.indexOf("://") + 3; i < basePath.length(); ++i) {
+      if (basePath.charAt(i) == '/') {
+        ++offset;
+      }
+    }
+    HttpUrl tweakedUrl =
+        url.newBuilder().removePathSegment(offset + 1).setPathSegment(offset, "api").build();
+
     return this.customObjectsApi
         .getApiClient()
         .getHttpClient()
--- a/util/src/test/java/io/kubernetes/client/util/generic/GenericKubernetesApiForCoreApiTest.java
+++ b/util/src/test/java/io/kubernetes/client/util/generic/GenericKubernetesApiForCoreApiTest.java
@@ -170,6 +170,36 @@ public class GenericKubernetesApiForCoreApiTest {
   }
 
   @Test
+  public void patchNamespacedPodWithApiPrefix() {
+    V1Patch v1Patch = new V1Patch("{}");
+    V1Pod foo1 =
+        new V1Pod().kind("Pod").metadata(new V1ObjectMeta().namespace("default").name("foo1"));
+    // add api prefix
+    String prefix = "/k8s/clusters/c-7q988";
+    stubFor(
+        patch(urlEqualTo(prefix + "/api/v1/namespaces/default/pods/foo1"))
+            .withHeader("Content-Type", containing(V1Patch.PATCH_FORMAT_STRATEGIC_MERGE_PATCH))
+            .willReturn(aResponse().withStatus(200).withBody(new Gson().toJson(foo1))));
+
+    GenericKubernetesApi<V1Pod, V1PodList> rancherPodClient =
+        new GenericKubernetesApi<>(
+            V1Pod.class,
+            V1PodList.class,
+            "",
+            "v1",
+            "pods",
+            new ClientBuilder().setBasePath("http://localhost:" + 8181 + prefix).build());
+    KubernetesApiResponse<V1Pod> podPatchResp =
+        rancherPodClient.patch(
+            "default", "foo1", V1Patch.PATCH_FORMAT_STRATEGIC_MERGE_PATCH, v1Patch);
+
+    assertTrue(podPatchResp.isSuccess());
+    assertEquals(foo1, podPatchResp.getObject());
+    assertNull(podPatchResp.getStatus());
+    verify(1, patchRequestedFor(urlPathEqualTo(prefix + "/api/v1/namespaces/default/pods/foo1")));
+  }
+
+  @Test
   public void testReadTimeoutShouldThrowException() {
     ApiClient apiClient = new ClientBuilder().setBasePath("http://localhost:" + 8181).build();
     apiClient.setHttpClient(


util/src/main/java/io/kubernetes/client/util/generic/GenericKubernetesApi.java
```diff
@@ -824,7 +824,17 @@ private Call tweakCallForCoreV1Group(Call call) {
       return call;
     }
     HttpUrl url = call.request().url();
-    HttpUrl tweakedUrl = url.newBuilder().removePathSegment(1).setPathSegment(0, "api").build();
+    String basePath = customObjectsApi.getApiClient().getBasePath();
+    // count segments in the basePath
+    int offset = 0;
```
brendandburns, line 829: I don't think this will work if there is a trailing `/` in basePath (e.g. `https://foo.bar/`) and I think it probably should, can you add a test case for that and validate

extended/src/test/java/io/kubernetes/client/extended/kubectl/KubectlApplyTest.java
```diff
@@ -98,4 +106,62 @@ public void testApplyConfigMap() throws KubectlException, IOException {
         1, patchRequestedFor(urlPathEqualTo("/api/v1/namespaces/foo/configmaps/bar")));
     assertNotNull(configMap);
   }
+
+  @Test
+  public void testApplyServiceInRancher() throws KubectlException, IOException {
```
yue9944882, line 111: @dddddai this repo will stay vendor-neutral so i dont think it's a good idea to put such a test that verifying the behavior for rancher. 

a better way to do this is to have an additional setter e.g. `setAPIPrefix` for the class `GenericKubernetesApi` so that users can tweak api path freely.

util/src/main/java/io/kubernetes/client/util/generic/GenericKubernetesApi.java
dddddai: Thanks for your review, actually the trailing `/` is removed in `ClientBuilder.build`, so basePath would not have a trailing `/` (unless the user create a `ApiClient` directly instead of using `ClientBuilder.build`)

https://github.com/kubernetes-client/java/blob/f20788272291c0e79a8c831d8d5a7dd94d96d2de/util/src/main/java/io/kubernetes/client/util/ClientBuilder.java#L432-L437

extended/src/test/java/io/kubernetes/client/extended/kubectl/KubectlApplyTest.java
dddddai: @yue9944882 Thanks for your review

> this repo will stay vendor-neutral so i dont think it's a good idea to put such a test that verifying the behavior for rancher.

Actually this test just makes sure `tweakCallForCoreV1Group` work fine with api prefix, so I guess the function name should be changed to something like `testApplyServiceWithApiPrefix`?

> a better way to do this is to have an additional setter e.g. setAPIPrefix for the class GenericKubernetesApi so that users can tweak api path freely.

But users could not see `GenericKubernetesApi` when using `Kubectl`, for example:
```java
V1Service service =
    Kubectl.apply(V1Service.class)
        .apiClient(new ClientBuilder().setBasePath("http://localhost:" + 8384 + prefix).build())
        .resource(
            new V1Service()
                .kind("Service")
                .apiVersion("v1")
                .metadata(
                    new V1ObjectMeta()
                        .namespace("foo")
                        .name("bar")
                        .labels(Collections.singletonMap("app", "bar")))
                .spec(spec))
        .execute();
```
In this case, users can only set api prefix on `ApiClient`, shall I add the setter in `ApiClient`?

yue9944882: @dddddai IIUC this pull doesnt need to involve changes in the `KubectlApply` related sources. making `GenericKuberentesApi` able to respect the prefix from the `ApiClient` should be sufficient. so we can remove this test, are we?

util/src/test/java/io/kubernetes/client/util/generic/GenericKubernetesApiForCoreApiTest.java
```diff
@@ -169,6 +169,36 @@ public void patchNamespacedPodReturningObject() {
...
+            .willReturn(aResponse().withStatus(200).withBody(new Gson().toJson(foo1))));
+
+    GenericKubernetesApi<V1Pod, V1PodList> rancherPodClient =
+        new GenericKubernetesApi<>(
+            V1Pod.class,
+            V1PodList.class,
+            "",
+            "v1",
+            "pods",
+            new ClientBuilder().setBasePath("http://localhost:" + 8181 + prefix).build());
```
yue9944882, line 191: on the other hand, i think the usage of `setBasePath` here is error-proning b/c the setter is doing an in-place modification upon the `ApiClient` instance. there will be thread-safety issue if multiple generic api modifying the basepath concurrently. ideally, what i expected is a function to provide an immutable shallow copy of `ApiClient` before we tweak the prefix, such as:

```
ApiClient shallowCopied = apiClient.newBuilder().setBasePath("<path>" + prefix).build()
```

we can do this in a follow-up anyway

yue9944882 (commented):
thanks, this pull will provide the extensibility for switching between multiple clusters. plz rename the test to get rid of the name like `rancher`

util/src/test/java/io/kubernetes/client/util/generic/GenericKubernetesApiForCoreApiTest.java
dddddai: > i think the usage of `setBasePath` here is error-proning b/c the setter is doing an in-place modification upon the `ApiClient` instance. there will be thread-safety issue if multiple generic api modifying the basepath concurrently

Sorry I didn't get your point, there is no method for `GenericKuberentesApi` to modify the basepath, each `GenericKuberentesApi` has its own `ApiClient`. Please correct me if I'm wrong.

Do you mean multiple `GenericKuberentesApi` sharing one `ApiClient`?

yue9944882: oh oversight, i took `ClientBuilder` as `ApiClient`.

yue9944882 (commented):
/lgtm
/approve

thanks!

Issue #1886, opened 2021-09-16
Java Client throws java.lang.NullPointerException: null at io.kubernetes.client.Discovery.lambda$null$5(Discovery.java:119)


saparaj:
Hi @yue9944882 and @brendanburns,

We are seeing this null pointer exception in Model.refresh() call - We are trying to do simple apply manifest 
```
apiVersion: v1
kind: ServiceAccount
metadata:
  name: sa-test
  labels:
    mytest: sa-test
```

Full Stack Trace
```
java.lang.NullPointerException: null
at io.kubernetes.client.Discovery.lambda$null$5(Discovery.java:119)
at java.util.Optional.ifPresent(Optional.java:159)
at io.kubernetes.client.Discovery.lambda$manageRelationFromResourceToSubResources$6(Discovery.java:117)
at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1384)
at java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:580)
at io.kubernetes.client.Discovery.manageRelationFromResourceToSubResources(Discovery.java:114)
at io.kubernetes.client.Discovery.groupResourcesByName(Discovery.java:97)
at io.kubernetes.client.Discovery.findAll(Discovery.java:72)
at io.kubernetes.client.Discovery.findAll(Discovery.java:65)
at io.kubernetes.client.Discovery.findAll(Discovery.java:55)
at io.kubernetes.client.util.ModelMapper.refresh(ModelMapper.java:239)
at io.kubernetes.client.util.ModelMapper.refresh(ModelMapper.java:219)
```
** Client Version **
12.0.0

** Kubernetes Version **
1.20.8

** Java Version **
Java 8

Please let us know any pointers on how to fix this or what is causing this?

brendandburns:
Something in your discovery APIs (the meta-api that lists the API is unexpected, can you attach the output of curl to `/api` and also to `/apis/<group>` for all of the APIs on your cluster.

We probably should add some defense in depth for a null-pointer in that case too.

saparaj:
@brendanburns Here is the output of curl 
![1- kubectl api-versions](https://user-images.githubusercontent.com/62569721/133737820-ed63a1de-b5a8-41f4-85bb-def5196329f9.png)
![2a-kubectl api-resources](https://user-images.githubusercontent.com/62569721/133737840-dfccc84a-ea78-4674-90cf-7ca8829bb84c.png)
[...]

PR #1894, opened 2021-09-21
Guard against NPE in manageRelationFromResourceToSubResources from Discovery class

saparaj:
Issue: https://github.com/kubernetes-client/java/issues/1886

This code is added to have defense against null pointer exception in manageRelationFromResourceToSubResources method in discovery class


yue9944882 (commented):
/lgtm
/approve

Issue #1897, opened 2021-09-23
Issues with building Dockerfiles in client-java-contrib


anirudha-ptr:
**Describe the bug**
Couple of issues with Dockerfiles in client-java-contrib:

- The base image is a private image in Dockerfile.gen. It should take the base image as an argument with default values.
- Dockerfile.gen-base is using variables and doesnt define them as arguments:
  - ${OPENAPI_GENERATOR_USER_ORG}
  - OPENAPI_GENERATOR_COMMIT
  Because of this the Dockerfile doesn't build.

We're building these images and adding these customizations on top of it. 

** Client Version **
12.0.0

** Kubernetes Version **
NA

** Java Version **
11

**To Reproduce**
Build the dockerfiles in client-contrib-java

**Expected behavior**
Dockerfile.gen-base should build, and we should be able to override the base image in Dockerfile.gen

brendandburns:
Thanks for sending the PR!

PR #1899, opened 2021-09-24
Fix arguments to dockerfiles

anirudha-ptr:
Couple of issues with Dockerfiles in client-java-contrib:

The base image is a private image in Dockerfile.gen. It should take the base image as an argument with default values.
Dockerfile.gen-base is using variables and doesnt define them as arguments:
${OPENAPI_GENERATOR_USER_ORG}
OPENAPI_GENERATOR_COMMIT
Because of this the Dockerfile doesn't build.

https://github.com/kubernetes-client/java/issues/1897


anirudha-ptr:
I signed it

brendandburns:
cc @yue9944882

PR #1963, opened 2021-11-01
Return a default APIClient if we run into an exception

### LINKED ISSUE #1946: Create A Default Client In Spring Autoconfiguration
Currently if an `ApiClient` cannot be created when Spring AutoConfiguration is run the application will fail to start.  This is common when running an app with `client-java-spring-integration` on the classpath but maybe starting the app outside of Kubernetes.  In Spring Cloud Kubernetes we have done something like this.  
https://github.com/spring-cloud/spring-cloud-kubernetes/blob/main/spring-cloud-kubernetes-client-autoconfig/src/main/java/org/springframework/cloud/kubernetes/client/KubernetesClientUtils.java#L34

I wonder if we could maybe do the same thing in the auto configuration classes in the Kuberentes Java Client?

ryanjbaxter:
Fixes #1946

### DIFF
--- a/spring/src/main/java/io/kubernetes/client/spring/extended/controller/config/KubernetesInformerAutoConfiguration.java
+++ b/spring/src/main/java/io/kubernetes/client/spring/extended/controller/config/KubernetesInformerAutoConfiguration.java
@@ -17,6 +17,8 @@ import io.kubernetes.client.openapi.ApiClient;
 import io.kubernetes.client.spring.extended.controller.KubernetesInformerFactoryProcessor;
 import io.kubernetes.client.util.ClientBuilder;
 import java.io.IOException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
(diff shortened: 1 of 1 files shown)


ryanjbaxter:
I signed it

spring/src/main/java/io/kubernetes/client/spring/extended/controller/config/KubernetesInformerAutoConfiguration.java
```diff
@@ -25,10 +28,32 @@
 @ConditionalOnKubernetesInformerEnabled
 public class KubernetesInformerAutoConfiguration {
 
+  private static final Log LOG = LogFactory.getLog(KubernetesInformerAutoConfiguration.class);
+
   @Bean
   @ConditionalOnMissingBean
   public ApiClient defaultApiClient() throws IOException {
-    return ClientBuilder.defaultClient();
+    try {
```
brendandburns, line 36: The current [`defaultClient()`](https://github.com/kubernetes-client/java/blob/master/util/src/main/java/io/kubernetes/client/util/ClientBuilder.java#L103) tries to load from a local config in the following order:
* `$KUBECONFIG` if it exists
* `$HOME/.kube/config` if it exists
* in-cluster configuration if it exists
* localhost:8080 if all else fails

I think we should maintain that order of selection.

What are the specific exceptions that you are seeing? They're not listed in the issue.

I'd prefer we fixed the specific exceptions while maintaining the same load order.

brendandburns: And perhaps it is better to add the exception handling to the `defaultClient()` function instead of the Spring loader.

spring/src/main/java/io/kubernetes/client/spring/extended/controller/config/KubernetesInformerAutoConfiguration.java
```diff
@@ -25,10 +28,32 @@
 @ConditionalOnKubernetesInformerEnabled
 public class KubernetesInformerAutoConfiguration {
 
+  private static final Log LOG = LogFactory.getLog(KubernetesInformerAutoConfiguration.class);
```
yue9944882, line 31: nit: conventionally we use slf4j as the logging interface

yue9944882 (commented):
plz run `mvn spotless:apply` to fix the format error

ryanjbaxter:
@yue9944882 @brendandburns sorry for the delay in getting back to this.

I actually think it might be simpler if we just try and create the ApiClient and if that fails we fall back to creating one that points to localhost.  Let me know what you think.

ryanjbaxter:
@yue9944882 I believe I fixed the formatting issues, sorry about that

yue9944882 (commented):
/lgtm
/approve

@ryanjbaxter thanks a bunch!

PR #1969, opened 2021-11-05
Update proto files to the latest definitions. Add some new versions.

brendandburns:
cc @yue9944882 

I'm not sure anyone uses the proto definitions, but they hadn't been updated in a _long_ time. I'm amazed it still worked :P


yue9944882:
@brendandburns great! can you run the formatter so that we can proceed to merge?

yue9944882 (commented):
/lgtm
/approve

PR #1976, opened 2021-11-12
Release 14

yue9944882:


yue9944882:
@brendandburns uploaded a new release 14.0.0 to the central, hopefully this will work 🤞btw i found that during the release the `release-14` and the tag `client-java-parent-14.0.0` will be directly pushed to this repo w/o any PR, is that correct?

brendandburns:
@yue9944882 yeah, it does that. Not sure if there is a way to adjust it so that it pushes to a personal fork instead...

/lgtm
/approve

PR #1981, opened 2021-11-14
Update ReflectorRunnable.java

shriyasankhyan:
Semaphore instead of thread.sleep


util/src/main/java/io/kubernetes/client/informer/cache/ReflectorRunnable.java
```diff
@@ -1,7 +1,7 @@
 /*
 Copyright 2020 The Kubernetes Authors.
 Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
+You may not use this file except in compliance with the License.
```
yue9944882, line 4: this will fail the licence formatter, lets move this to another pull

Issue #1802, opened 2021-07-26
throw NullPointerException when Config.defaultClient()


lusson-luo:
**Describe the bug**
throw NullPointerException when Config.defaultClient()

** Client Version **
`8.0.2`

** Kubernetes Version **
`v1.18.10`

** Java Version **
Java 8

**To Reproduce**
use this kubeconfig, then run 
```
Config.defaultClient();
```

**Expected behavior**
 runs normally or If it's a Config format error, tell it's a Config format error, not a NullPointerException

**KubeConfig**

```
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: xxxxxxxxxx
    server: https://xx.xx.xx.xx/k8s/clusters/c-mzvvs
  name: dev03
contexts: null
current-context: dev03
kind: Config
preferences: {}
users:
- name: dev03
  user:
    token: xx.xx
```

**Server (please complete the following information):**
 - Linux
 - container

**Additional context**

**cause**

kubeconfig parse currentCluster is null，The code does not make a null pointer judgment

![image](https://user-images.githubusercontent.com/19859400/127026182-b916bfa4-ad47-41ce-b837-4b682fce8410.png)

**error log**

> 
> Caused by: org.springframework.beans.BeanInstantiationException: Failed to instantiate [net.xx.gateway.autoload.AutoLoadConfig$$EnhancerBySpringCGLIB$$27d9f98a]: Constructor threw exception; nested exception is java.lang.NullPointerException
> 	at org.springframework.beans.BeanUtils.instantiateClass(BeanUtils.java:217)
> 	at org.springframework.beans.factory.support.SimpleInstantiationStrategy.instantiate(SimpleInstantiationStrategy.java:87)
> 	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.instantiateBean(AbstractAutowireCapableBeanFactory.java:1310)
> 	... 16 common frames omitted
> Caused by: java.lang.NullPointerException: null
> 	at io.kubernetes.client.util.ClientBuilder.kubeconfig(ClientBuilder.java:245)
> 	at io.kubernetes.client.util.ClientBuilder.getClientBuilder(ClientBuilder.java:114)
> 	at io.kubernetes.client.util.ClientBuilder.standard(ClientBuilder.java:93)
> 	at io.kubernetes.client.util.ClientBuilder.standard(ClientBuilder.java:85)
> 	at io.kubernetes.client.util.Config.defaultClient(Config.java:113)
[...]

PR #1983, opened 2021-11-14
#1802 Check if server is available from kubeconfig.

cbuschka:
Added null check with IllegalArgException and added a test case to ClientBuilderTest.


brendandburns:
Looks good, but you need to run `mvn spotless:apply` to format the code.

brendandburns:
@cbuschka apologies but we're apparently in the middle of transitioning CLA bots, you'll need to sign the older CLA for now. Sorry for the noise.

cbuschka:
@brendandburns No worries, I have signed two CLAs. One with DocuSign and one with HelloSign. Can you point me to a place I can recheck if something is still missing?

brendandburns:
Thanks for the patience. merging.

PR #1991, opened 2021-11-16
Regenerate fluent apis

yue9944882:
cc @brendandburns


brendandburns:
Are these in the `14.0.0` release? If not we should probably cut a `14.0.1`

/lgtm
/approve

yue9944882:
> Are these in the 14.0.0 release? If not we should probably cut a 14.0.1

@brendandburns the module `fluent-gen` is never published, so i guess it's fine. or we can separate the versioning of this module from the parent starting from a `v1.0.0` b/c it's irrelevant to the parent's versioning, wdyt?

brendandburns:
Sorry, I wasn't clear: what I was wondering was: "did you regenerate the fluent bindings for the 14.0.0 release?"

And then I realized that _I_ actually had regenerated them in #1936, so I was just confused.

Apologies for the noise.

Issue #1629, opened 2021-04-03
Difficult to extend exception handler


domainname:
@yue9944882 @brendandburns 

It's difficult to extend exception handler in ReflectorRunnable. Controller does not expose a property for developer to inject a customized exception handler. Nor does SharedInformerFactory. One probably has to write his/her own SharedInformerFactory and Controller, in order to inject a customized exception handler.

brendandburns:
Can you give an example of the code you would like to write or the function signature you are looking for?

We can definitely look to improve things.

Thanks!

domainname:
Hi @brendanburns. This is a typical way to write a informer. There is nowhere to inject an exception handler.

```java
        SharedIndexInformer<V1Deployment> deploymentInformer =
            factory.sharedIndexInformerFor(
                (CallGeneratorParams params) -> api.listDeploymentForAllNamespacesCall(
                    null, null, null, null, null,
                    null, params.resourceVersion, null, params.timeoutSeconds, params.watch,
                    null),
                V1Deployment.class, V1DeploymentList.class);
        deploymentInformer.addEventHandler(deploymentEventHandler);
```

The function signature is 

```java
public class SharedInformerFactory {
  ...
  public synchronized <ApiType extends KubernetesObject, ApiListType extends KubernetesListObject>
      SharedIndexInformer<ApiType> sharedIndexInformerFor(
          CallGenerator callGenerator,
          Class<ApiType> apiTypeClass,
          Class<ApiListType> apiListTypeClass);
  ...
}
```

I hope to see a signature like this:

```java
  public synchronized <ApiType extends KubernetesObject, ApiListType extends KubernetesListObject>
      SharedIndexInformer<ApiType> sharedIndexInformerFor(
          CallGenerator callGenerator,
          Class<ApiType> apiTypeClass,
          Class<ApiListType> apiListTypeClass,
          BiConsumer<Class<ApiType>, Throwable> exceptionHandler);
```

brendandburns:
I think I'm still confused. The `DefaultSharedIndexInformer` doesn't have any `catch` blocks in it, the ReflectorRunnable has a parameter where you can pass in the handler.

Are you talking about the `Controller` class, it is true that doesn't make it easy to pass in an exception handler, we could certainly add a parameter there.

Thanks!

domainname:
[...]

PR #1992, opened 2021-11-16
Add exception handler to SharedInformerFactory

cbuschka:


cbuschka:
Posted too early, sorry.

cbuschka:
Relates to issue #1629

util/src/main/java/io/kubernetes/client/informer/cache/Controller.java
```diff
@@ -65,19 +66,34 @@
...
-    this.queue = queue;
+    this(apiTypeClass, queue, listerWatcher, processFunc, resyncFunc, fullResyncPeriod, null);
+  }
+    public Controller(
+          Class<ApiType> apiTypeClass,
+          DeltaFIFO queue,
+          ListerWatcher<ApiType, ApiListType> listerWatcher,
+          Consumer<Deque<MutablePair<DeltaFIFO.DeltaType, KubernetesObject>>> processFunc,
+          Supplier<Boolean> resyncFunc,
+      long fullResyncPeriod,
```
brendandburns, line 86: nit: indentation.  Please run `mvn spotless:apply` to apply formatting.

util/src/main/java/io/kubernetes/client/informer/impl/DefaultSharedIndexInformer.java
```diff
@@ -82,16 +83,47 @@ public DefaultSharedIndexInformer(
...
+    public DefaultSharedIndexInformer(
+        Class<ApiType> apiTypeClass,
+        ListerWatcher<ApiType, ApiListType> listerWatcher,
+        long resyncPeriod,
+        Cache<ApiType> cache,
+        BiConsumer<Class<ApiType>, Throwable> exceptionHandler) {
+        this(apiTypeClass, listerWatcher, resyncPeriod,
+            // down-casting should be safe here because one delta FIFO instance only serves one
+            // resource
+            // type
```
brendandburns, line 100: nit pull this comment up to the previous line.

brendandburns:
This looks reasonable to me. Can you add unit tests to ensure we don't regress in the future?

Thanks!

cbuschka:
@brendandburns Sorry for the confusion. Could you please link this pr to issue #1629. In https://github.com/kubernetes-client/java/issues/1629#issuecomment-970681857 I wanted @domainname to confirm that this would solve his problem first. Sure it lacks reasonable default value handling, tests, and a little love ;)

util/src/main/java/io/kubernetes/client/informer/cache/Controller.java
```diff
@@ -113,7 +129,7 @@ public void run() {
 
     synchronized (this) {
       // TODO(yue9944882): proper naming for reflector
-      reflector = new ReflectorRunnable<ApiType, ApiListType>(apiTypeClass, listerWatcher, queue);
+      reflector = newReflector();
```
cbuschka, line 132: @brendandburns Thought about injecting a ReflectorFactory to improve testability. But then I chose this way so we do not add another class and we still can choose this option later.

util/src/test/java/io/kubernetes/client/informer/impl/DefaultSharedIndexInformerTest.java
```diff
@@ -12,366 +12,47 @@
 */
 package io.kubernetes.client.informer.impl;
```
cbuschka, line 13: Renamed DefautSharedInformterTest to DefaultSharedIndexInformerWireMockTest and added a pure unit test DefautSharedInformterTest. (It is not as big as it looks like :) )

util/src/main/java/io/kubernetes/client/informer/cache/Controller.java
```diff
@@ -65,19 +66,34 @@
 
   private ScheduledFuture reflectorFuture;
 
+  /* visible for testing */ BiConsumer<Class<ApiType>, Throwable> exceptionHandler;
+
   public Controller(
```
cbuschka, line 71: I am not sure if these constructor chains are a good strategy long term. The best option I currently see is
- add a builder for the class
- use the all params constructor from the builder
- deprecate all constructors with partial param sets

Also I would try to get rid of the many new calls here via factories. This could improve testability and introduce entry points for people who want to influence behaviour. A larger change but could be done in baby steps. What do you think @brendandburns ?

cbuschka:
@brendandburns Now it is ready for review. Could you take a look into it, please. I have added some comments. But they are more kind of a discussion. Is there a place to discuss issues?

util/src/main/java/io/kubernetes/client/informer/cache/Controller.java
brendandburns: Can you file this as an issue so we can have a broader discussion? I want to get @yue9944882 to weigh in as well.

util/src/test/java/io/kubernetes/client/informer/cache/ReflectorRunnableTest.java
```diff
@@ -12,8 +12,7 @@
 */
 package io.kubernetes.client.informer.cache;
 
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.*;
```
brendandburns, line 15: style is no wildcard imports.

brendandburns:
This looks good to me, but would like @yue9944882 to look at the test refactor to make sure we didn't lose any coverage.

util/src/test/java/io/kubernetes/client/informer/cache/ReflectorRunnableTest.java
cbuschka: Fixed

util/src/main/java/io/kubernetes/client/informer/impl/DefaultSharedIndexInformer.java
cbuschka: Fixed.

util/src/main/java/io/kubernetes/client/informer/cache/Controller.java
cbuschka: Yay, have created issue #1995 for discussion. Looking forward to if :)

yue9944882 (commented):
/lgtm
/approve

PR #2040, opened 2021-12-22
gh-action: Upload snapshot as post-commit job

yue9944882:
@brendandburns i added my jira account credentials and gpg keys in the repo secrets to push snapshots. github claims that these secrets are perfectly protected 🤞

if this works well, we can move the maven release job as another job triggered by simple git tags. so we don't need to deal with the fiddly release process manually anymore!

### DIFF
--- a/.github/workflows/snapshot.yml
+++ b/.github/workflows/snapshot.yml
@@ -1,0 +1,29 @@
+name: Upload Snapshot
+
+on:
+  push:
+    branches:
+      - master
+      - release-*
+  workflow_dispatch: {}
+
+jobs:
+  verify:
+    name: latest-images
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Setup Java
+        uses: actions/setup-java@v2
+        with:
+          distribution: 'temurin'
+          java-version: 8.0.x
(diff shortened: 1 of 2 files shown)


brendandburns:
Thanks for doing this. Is there an example of where these snapshots end-up? e.g. what is the version?

yue9944882:
@brendandburns they're pushed to sonatype oss snapshot repository, so developers can get quick access to the master branch package by adding the following repository to their pom.xml:

```xml
  <repositories>
    <!-- Added to get the Atmosphere 1.1.0-SNAPSHOT, can be removed when 1.1.0 is released -->
    <repository>
      <id>oss.sonatype.org-snapshot</id>
      <url>http://oss.sonatype.org/content/repositories/snapshots</url>
      <releases>
        <enabled>false</enabled>
      </releases>
      <snapshots>
        <enabled>true</enabled>
      </snapshots>
    </repository>
  </repositories>
```

> e.g. what is the version?

the snapshots are built from branch `master`,  i added a badge to show the latest snapshot version.
Sonatype Nexus (Snapshots)

brendandburns:
/lgtm
/approve

Thanks for starting us towards GH action based automation!

brendandburns:
(we should probably expand this to the release branches also)

PR #2050, opened 2021-12-24
Release: Fixes snapshot job env var name

yue9944882:
s/MAVEN_GPG_PASSPHRASE/GPG_PASSPHRASE/


brendandburns:
Let's separate the fix and the deletion of some of the examples.

PR #2056, opened 2021-12-31
Sync automated release 10.0.2

yue9944882:
@brendandburns i was testing automating the release via github action on release-10. as a result the `10.0.2` release is successfully pushed to nexus central. but the experiment was on my fork so i'm syncing the release commits to the upstream repo.

Issue #2041, opened 2021-12-22
Release 11.0.4


ryanjbaxter:
Can we release 11.0.4?

I would like to take advantage of this fix
https://github.com/kubernetes-client/java/pull/1963

yue9944882:
@ryanjbaxter for sure, i'm working moving the release infra to github action from snapshots #2040 so that we can easily publish releases by pushing tags. i will send a 11.0.4 on the new release infra

/assign

btw how do you think about picking the commit to our major branch?

ryanjbaxter:
Great thanks!

> btw how do you think about picking the commit to our major branch?

Sounds good would you like me to do that?

yue9944882:
> Sounds good would you like me to do that?

yea by any chance can you pick it to master branch?

ryanjbaxter:
@yue9944882 Here is the PR

yue9944882:
https://github.com/kubernetes-client/java/runs/4734270953?check_suite_focus=true

@ryanjbaxter 11.0.4 release is published. @brendandburns the sad news is that our release job failed in the last step after the artifact is published. 

> remote: error: GH006: Protected branch update failed for refs/heads/release-11.        
remote: error: Required status check "cla/linuxfoundation" is expected.        
To https://github.com/kubernetes-client/java.git
 ! [remote rejected] release-11 -> release-11 (protected branch hook declined)
error: failed to push some refs to 'https://github.com/kubernetes-client/java.git'

i opened https://github.com/kubernetes-client/java/pull/2069 to catch up with the automated release

ryanjbaxter:
Great thanks!  We will update next week!

yue9944882:
/close

the 11.0.4 is available

PR #2057, opened 2021-12-31
Adding an manual release job that applies to release branches publishing maven releases

yue9944882:
> Signed-off-by: yue9944882 <291271447@qq.com>

@brendandburns i also clarified the usage of the release job in the doc.

Screen Shot 2021-12-31 at 4 17 00 PM 1

https://github.com/kubernetes-client/java/pull/2056 is an example of the generated commits by the job. as the next step, we can cherry-pick this release job to the recent release branches and send a release `11.0.4` for https://github.com/kubernetes-client/java/issues/2041

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,84 @@
+name: Maven Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      releaseVersion:
+        type: string
+        required: true
+        description: The POM release version of this release.
+      nextDevelopmentVersion:
+        type: string
+        required: true
+        description: The next POM development version with suffix "-SNAPSHOT", will be the next development version after the release is done.
+      dry-run:
+        type: boolean
+        required: true
+        description: Dry run, will not push branches or upload the artifacts.
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Validate Input
+        run: |
+          echo "${{ github.ref_type }}" | perl -ne 'die unless m/^branch$/'
+          echo "${{ github.ref_name }}" | perl -ne 'die unless m/^release-\d+$/'
+          echo "${{ github.event.inputs.releaseVersion }}" | perl -ne 'die unless m/^\d+\.\d+\.\d+$/'
+          echo "${{ github.event.inputs.nextDevelopmentVersion }}" | perl -ne 'die unless m/^\d+\.\d+\.\d+-SNAPSHOT$/'
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Check Actor
+        run: |
+          # Release actor should be in the OWNER list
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,88 @@
...
+          echo "${{ github.event.inputs.nextDevelopmentVersion }}" | perl -ne 'die unless m/^\d+\.\d+\.\d+-SNAPSHOT$/'
+      - name: Check Actor
+        run: |
+          # Release actor should be in the OWNER list
+          cat OWNERS | grep ${{ github.actor }}
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Import GPG key
+        id: import_gpg
+        uses: crazy-max/ghaction-import-gpg@v4
```
brendandburns, line 37: Is there a way that we could use a more official github action for this? I'm a little worried about GPG key exfiltration from a random github action that could get compromised.

brendandburns:
One comment about the GPG key handling....

.github/workflows/release.yml
yue9944882: @brendandburns i didnt find an official action for importing the gpg keys, i tried importing via command lines but it's failing on missing tty device or sth. https://github.com/crazy-max/ghaction-import-gpg this action seems actively maintained and accumulated a few followers. we can switch to official if there is one in the future, what do you think?

brendandburns: I don't know. I think I'm super paranoid about supply chain security these days, and handing our signing keys over to a GH action that could get compromised seems like a bad idea.

I will ping the GH people that I know and see if there is a better option. If not we'll see what we can do.

Apologies for the delay!

brendandburns: Can you look at this GH Action here:

https://github.com/microsoft/gctoolkit/actions/runs/1604755958/workflow#L32

It appears that you can use the `export GPG_TTY=$(tty)` line to setup the tty correctly 
https://github.com/microsoft/gctoolkit/actions/runs/1604755958/workflow#L42

Let me know if that doesn't work. The Microsoft/GitHub Java folks claim it should work.

Thanks!

yue9944882: @brendandburns removed the third-party gpg action. alongside `export GPG_TTY=$(tty)` i also added `--pinentry-mode loopback --passphrase ***` to make the --edit-key works. the setup-java action works by editing the settings.xml profile on the worker node but it doesn't directly set up gpg for us. 

and in our case, we need to do some additional work editing the key's trust level b/c the key is currently expired. am new to the gpg key renewing process so i stick to the expired key in the release job. we can renew one as a follow-up in the future, after that our release job can be simplified further.

brendandburns:
Thanks! You can just extend the expiration date for the key, I think, see for example:

https://superuser.com/questions/813421/can-you-extend-the-expiration-date-of-an-already-expired-gpg-key

Issue #1638, opened 2021-04-08
Client is not compatible with Java 16


holgerstolzenberg:
**Describe the bug**
I tried to upgrade a Spring Boot app from Java 14 to 16. The app also includes Spring Cloud which under the hood uses the Kubernetes client library. The app fails to start up do to an illegal access of [KubernetesFromYamlProcessor](https://github.com/kubernetes-client/java/blob/master/spring/src/main/java/io/kubernetes/client/spring/extended/manifests/KubernetesFromYamlProcessor.java). 

Here is the full stacktrace:
```
Caused by: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'echoesService' defined in class path resource [de/dehst/platform/playground/service/modules/echoes/EchoesService.class]: Unsatisfied dependency expressed through constructor parameter 0; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'echoesRepository' defined in de.dehst.platform.playground.service.modules.echoes.EchoesRepository defined in @EnableJpaRepositories declared on Application: Cannot create inner bean '(inner bean)#389a9ff6' of type [org.springframework.orm.jpa.SharedEntityManagerCreator] while setting bean property 'entityManager'; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name '(inner bean)#389a9ff6': Cannot resolve reference to bean 'entityManagerFactory' while setting constructor argument; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'entityManagerFactory': Post-processing of FactoryBean's singleton object failed; nested exception is java.lang.reflect.InaccessibleObjectException: Unable to make field private static final java.lang.reflect.Method jdk.proxy2.$Proxy166.m0 accessible: module jdk.proxy2 does not "opens jdk.proxy2" to unnamed module @324e4822
        at org.springframework.beans.factory.support.ConstructorResolver.createArgumentArray(ConstructorResolver.java:800) ~[spring-beans-5.3.5.jar:5.3.5]
        at org.springframework.beans.factory.support.ConstructorResolver.autowireConstructor(ConstructorResolver.java:229) ~[spring-beans-5.3.5.jar:5.3.5]
        at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.autowireConstructor(AbstractAutowireCapableBeanFactory.java:1354) ~[spring-beans-5.3.5.jar:5.3.5]
[...]

PR #2078, opened 2022-01-10
Fixes java 16 compability

yue9944882:
1. replacing `com.github.stefanbirkner/system-lambda` w/ `uk.org.webcompere/system-stubs-core`, which is a dropin replacement for the former library that fixes the environment issue.

2. adding a default contructor for `Watch#Response` to avoid gson's reflective dynamic constructor which violates the jdk16 restriction.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -262,6 +262,12 @@
         <scope>test</scope>
       </dependency>
       <dependency>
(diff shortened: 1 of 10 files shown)


yue9944882:
Fixes https://github.com/kubernetes-client/java/issues/1638

PR #2080, opened 2022-01-11
Improve descriptions a little bit with examples.

brendandburns:
When I did the release, I was a little confused with what was being requested, so I updated the descriptions with some examples.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,11 +6,11 @@ on:
       releaseVersion:
         type: string
         required: true
-        description: The POM release version of this release.
+        description: The POM release version of this release. Must be a semantic version of the form X.Y.Z.
       nextDevelopmentVersion:
         type: string
         required: true
-        description: The next POM development version with suffix "-SNAPSHOT", will be the next development version after the release is done.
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -6,11 +6,11 @@ on:
       releaseVersion:
         type: string
         required: true
-        description: The POM release version of this release.
+        description: The POM release version of this release. Must be a semantic version of the form X.Y.Z.
       nextDevelopmentVersion:
         type: string
         required: true
-        description: The next POM development version with suffix "-SNAPSHOT", will be the next development version after the release is done.
+        description: The next POM development version after the release is done. Must be of the form X.Y.${Z+1}-SNAPSHOT
```
yue9944882, line 13: @brendandburns actually it is the default behavior of the maven release plugin but i had this manual option on purpose in case we will bump a mid/minor version of the release sometime in the future. so for clarification it should be either `X.${Y+1}.0-SNAPSHOT` or `X.Y.${Z+1}-SNAPSHOT`, wdyt?

yue9944882:
@brendandburns bump ^^^^

yue9944882 (commented):
/lgtm

getting this in, let me clarify the detail in a follow-up

PR #2116, opened 2022-01-28
Add Deployment rollout restart example

ddubson:
This is a client-java v15.x compatible example of a Deployment resource rollout restart function which involves several indirect/non-trivial steps, which I thought could be quite useful for anyone seeking to have greater programmatic control over their Kubernetes deployments.

### DIFF
--- a/examples/examples-release-15/src/main/java/io/kubernetes/client/examples/DeployRolloutRestartExample.java
+++ b/examples/examples-release-15/src/main/java/io/kubernetes/client/examples/DeployRolloutRestartExample.java
@@ -1,0 +1,140 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


examples/examples-release-15/src/main/java/io/kubernetes/client/examples/DeployRolloutRestartExample.java
```diff
@@ -0,0 +1,97 @@
...
+import io.kubernetes.client.util.wait.Wait;
+
+public class DeployRolloutRestartExample {
+	public static void main(String[] args) throws IOException, ApiException {
+		ApiClient client = Config.defaultClient();
+		Configuration.setDefaultApiClient(client);
+		AppsV1Api appsV1Api = new AppsV1Api(client);
+
+		String deploymentName = "example-nginx";
+		String imageName = "nginx:latest";
```
brendandburns, line 44: fwiw, using the `latest` tag is an anti-pattern maybe make this a specific semantic version?

brendandburns:
Thanks for the PR! One minor nit, otherwise LGTM.

yue9944882 (commented):
plz run the formatter `mvn spotless:apply`, otherwise LGTM

examples/examples-release-15/src/main/java/io/kubernetes/client/examples/DeployRolloutRestartExample.java
ddubson: 👍🏼 Thanks for catching that! I applied a specific tag `1.21.6` instead of `latest`.

ddubson:
@yue9944882 ah I was looking for the style formatter, thanks for guiding me to it. I applied it and re-pushed my changes.

PR #2148, opened 2022-02-23
Add support for transforming objects in shared informers

### LINKED ISSUE #2146: Support for transforming objects in shared informers
Recently k8s client-go has introduced a transform func that allows users to customize objects in informers, I think we should follow it as it can be really beneficial (e.g. reducing memory usage)

For more info please refer to:
https://github.com/kubernetes/kubernetes/pull/104300
https://github.com/kubernetes/kubernetes/pull/107507

dddddai:
Fixes #2146

### DIFF
--- a/util/src/main/java/io/kubernetes/client/informer/SharedInformer.java
+++ b/util/src/main/java/io/kubernetes/client/informer/SharedInformer.java
@@ -54,4 +54,19 @@ public interface SharedInformer<ApiType extends KubernetesObject> {
   // store. The value returned is not synchronized with access to the underlying store and is not
   // thread-safe.
   String lastSyncResourceVersion();
+
+  /**
+   * The TransformFunc is called for each object which is about to be stored. This function is
+   * intended for you to take the opportunity to remove, transform, or normalize fields. One use
+   * case is to strip unused metadata fields out of objects to save on RAM cost.
+   *
+   * <p>Must be set before starting the informer.
+   *
+   * <p>Note: Since the object given to the handler may be already shared with other goroutines, it
+   * is advisable to copy the object being transform before mutating it at all and returning the
+   * copy to prevent data races.
+   *
+   * @param transformFunc the transform function
+   */
+  void setTransform(TransformFunc transformFunc);
 }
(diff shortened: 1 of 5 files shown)


util/src/main/java/io/kubernetes/client/informer/SharedInformer.java
```diff
@@ -54,4 +55,19 @@
...
+   *
+   * <p>Must be set before starting the informer.
+   *
+   * <p>Note: Since the object given to the handler may be already shared with other goroutines, it
+   * is advisable to copy the object being transform before mutating it at all and returning the
+   * copy to prevent data races.
+   *
+   * @param transformFunc the transform function
+   */
+  void setTransformFunc(Function<KubernetesObject, KubernetesObject> transformFunc);
```
yue9944882, line 72: lets define a functional interface named `TransformFunc` to align w/ the upsteam, also the original go function returns an error upon failure, so i think we should declare a checked exception e.g. `throws ObjectTransformException`. the unit test should also cover the failure path when the checked exception is thrown.

yue9944882 (commented):
@dddddai thanks! one outstanding comment otherwise lgtm

util/src/main/java/io/kubernetes/client/informer/SharedInformer.java
dddddai: > also the original go function returns an error upon failure, so i think we should declare a checked exception e.g. `throws ObjectTransformException`.

The exception handling references `addIndexers`, which returns an error in client-go while in client-java it doesn't declare an exception
https://github.com/kubernetes/kubernetes/blob/v1.23.4/staging/src/k8s.io/client-go/tools/cache/shared_informer.go#L189
https://github.com/kubernetes-client/java/blob/a5a7ce94abab72391b6425436e7e5af01a4c5854/util/src/main/java/io/kubernetes/client/informer/impl/DefaultSharedIndexInformer.java#L287-L292
Do you mean something like this?
```java
public interface TransformFunc {

  /**
   * @param the original object to be transformed
   * @return the transformed object
   */
  KubernetesObject transform(KubernetesObject object) throws ObjectTransformException;
}
```

yue9944882: > Do you mean something like this?

yes, thats exactly what i thought, as for `addIndexers`, i think we can keep it, it's the legacy code and it's working.

dddddai: OK, fixed

yue9944882 (commented):
/lgtm
/approve

PR #2187, opened 2022-03-28
don't use internal gson package when JDK provides alternative

rotty3000:
Signed-off-by: Raymond Augé <raymond.auge@liferay.com>


brendandburns:
fwiw, this file is code generated, but we set it to ignore a while ago and so we're not currently generating it.

For the time being, I'm ok merging this, but ultimately we should go back to generating this code.

rotty3000:
Please let me know if there is something else to be done here. I don't want to make further PRs without at least the first one fully processed :)

thx for your consideration.

yue9944882:
the edited file is already listed in the ignore list, so it's fine to append more edits directly. as a follow-up we should submit it the code-generater community.

/lgtm

PR #2197, opened 2022-04-02
Adding a gh-action for generated java job

yue9944882:
@brendandburns this automated workflow can help us to do the code generation via manually running workflow, in the end,a PR will be opened. see an example at https://github.com/yue9944882/java/pull/23/files

### DIFF
--- a/.github/workflows/generate.yml
+++ b/.github/workflows/generate.yml
@@ -1,0 +1,64 @@
+name: Generate
+
+on:
+  workflow_dispatch:
+    inputs:
+      kubernetesBranch:
+        type: string
+        required: true
+        description: 'The remote kubernetes release branch to fetch openapi spec. .e.g. "release-1.23"'
+
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Java
+        uses: actions/checkout@v3
+      - name: Checkout Gen
+        uses: actions/checkout@v3
+        with:
+          path: gen
+          repository: kubernetes-client/gen
+          token: ${{ secrets.PAT_TOKEN }}
+      - name: Generate Branch Name
+        run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/generate.yml
```diff
@@ -0,0 +1,64 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Java
+        uses: actions/checkout@v3
+      - name: Checkout Gen
+        uses: actions/checkout@v3
+        with:
+          path: gen
+          repository: kubernetes-client/gen
+          token: ${{ secrets.PAT_TOKEN }}
```
brendandburns, line 23: I don't think we need a token do we? This should be fully public.

brendandburns:
Generally looks good. One small comment.

We should consider creating a shared robot account so that we don't have to use our own personal access tokens.

brendandburns:
Also, fwiw, this won't regenerate the fluent APIs so this is only going to work for patch updates (for now).

.github/workflows/generate.yml
yue9944882: the default token doesnt allow any cross-repo read/write even tho the repo is public. so we have to use a PAT token here, i granted the token merely read access

yue9944882:
> We should consider creating a shared robot account so that we don't have to use our own personal access tokens.

yeah the configured PAT token in this repo is current from my account, but the good news is that the permission is limitted (read only). ideally we should replace my personal token w/ a robot token.

> Also, fwiw, this won't regenerate the fluent APIs so this is only going to work for patch updates (for now).

@brendandburns nice catch, now i added fluent generation to the job.

yue9944882:
@brendandburns also we need to pick this to release-15 branch after it's merged

brendandburns:
/lgtm
/approve

Thanks!

PR #2215, opened 2022-04-16
refactor: Use `Collections#isEmpty()` instead of comparing `size()`

yeikel:
Non functional change to improve the code quality. See https://rules.sonarsource.com/java/RSPEC-1155

This is the result of running the recipe `org.openrewrite.java.cleanup.IsEmptyCallOnCollections` from the OpenRewrite project

Read more : https://docs.openrewrite.org/reference/recipes/java/cleanup/isemptycalloncollections

Created with: [Moderne](https://app.moderne.io)

### DIFF
--- a/extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlDrain.java
+++ b/extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlDrain.java
@@ -88,7 +88,7 @@ public class KubectlDrain extends KubectlCordon {
     for (V1Pod pod : pods) {
       if (pod.getMetadata().getOwnerReferences() == null) continue;
 
-      if (!force && pod.getMetadata().getOwnerReferences().size() == 0) {
+      if (!force && pod.getMetadata().getOwnerReferences().isEmpty()) {
         throw new KubectlException("Pods unmanaged by a controller are present on the node");
       }
       // Throw if there are daemon set pods and ignore daemon set is false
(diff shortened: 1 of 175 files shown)


brendandburns:
I personally don't see any difference between `isEmpty()` and `size() == 0` they both are visually the same to me.

@yue9944882 wdyt?

Also needs formatting.

yeikel:
> I personally don't see any difference between `isEmpty()` and `size() == 0` they both are visually the same to me.
> 
> @yue9944882 wdyt?
> 
> Also needs formatting.

What formatting are you referring to?

They're the same but not idiomatic. Using a  `size` check is not necessary when there is a dedicated method to do the same

yue9944882:
@brendandburns the distinction between the two methods is that `Collections#isEmpty()` will also handle the null collection object, overall i agree this is a positive refactor. @yeikel can you re-format by running `mvn spotless:apply`?

brendandburns:
I thought about this some more.

I don't think that one-off refactors are useful for style. If we want to change our coding style, it needs to be enforced via automation or else it will just drift back to where it is.

Currently this project uses the spotless plugin with GoogleJavaFormat as it's style enforcer, plus a few additional rules about how imports are formatted. This is specified in the `pom.xml`

I personally don't see enough value in the code change to force people to use `isEmpty()` instead of `size() == 0` and if we're not forcing it via spotless, then I don't see the value in a refactor PR (see above).

yeikel:
> I thought about this some more.
> 
> I don't think that one-off refactors are useful for style. If we want to change our coding style, it needs to be enforced via automation or else it will just drift back to where it is.
> 
> Currently this project uses the spotless plugin with GoogleJavaFormat as it's style enforcer, plus a few additional rules about how imports are formatted. This is specified in the `pom.xml`
> 
> I personally don't see enough value in the code change to force people to use `isEmpty()` instead of `size() == 0` and if we're not forcing it via spotless, then I don't see the value in a refactor PR (see above).
> 
> To even be considered for merging, this PR will need to include changes to `pom.xml` that enable automation and require the formatting, but again, I'm not sure that requiring this style adds value to this project.

Rather than requiring, would you be in favor of adding automation to apply it automatically? 

We can onboard this project to OpenRewrite to do this plus many of the other recipes available

Issue #2318, opened 2022-07-21
Unexpected unknown resource type exception while Kubectl.apply


fbi007130:
**Describe the bug**
I am trying to apply a yaml file with Kubectl according to https://github.com/kubernetes-client/java/blob/b0c91798c63203f20dc8f2674197e772b2c187ee/docs/kubectl-equivalence-in-java.md#kubectl-apply .But there is a exception as below.The yaml file can be execute successfully via kubectl command in ssh.
Exception:
`io.kubernetes.client.extended.kubectl.exception.KubectlException: Unexpected unknown resource type: class io.kubernetes.client.util.generic.dynamic.DynamicKubernetesObject
	at io.kubernetes.client.extended.kubectl.Kubectl$ApiClientBuilder.getGenericApi(Kubectl.java:268)
	at io.kubernetes.client.extended.kubectl.Kubectl$ResourceBuilder.getGenericApi(Kubectl.java:309)
	at io.kubernetes.client.extended.kubectl.KubectlApply.executeServerSideApply(KubectlApply.java:73)
	at io.kubernetes.client.extended.kubectl.KubectlApply.execute(KubectlApply.java:67)
	at com.ChaoBladeStepdefs.deletePod(ChaoBladeStepdefs.java:47)
	at ✽.Delete pod(file:///C:/Users//Test.feature:4)`

Yaml file:

`apiVersion: chaosblade.io/v1alpha1
kind: ChaosBlade
metadata:
  name: delete-pod-by-labels
spec:
  experiments:
  - scope: pod
    target: pod
    action: delete
    desc: "delete pod by labels"
    matchers:
    - name: labels
      value:
      - "app.kubernetes.io/instance=valuevalue"
    - name: namespace
      value:
      - "idm"
    - name: evict-percent
      value:
      - "100"
`

Java code:
`String yamlStr=new String(Files.readAllBytes(Paths.get("/tmp/deletepod.yaml")));
DynamicKubernetesObject dynamicK8sObject=Dynamics.newFromYaml(yamlStr);
Kubectl.apply(DynamicKubernetesObject.class).resource(dynamicK8sObject).execute();
`

**Client Version**
`16.0.0`

**Kubernetes Version**
 `1.23.3`

**Java Version**
Java 8

**To Reproduce**

**Expected behavior**
delete-pod-by-labels should be create.

**KubeConfig**
If applicable, add a KubeConfig file with secrets redacted.

**Server (please complete the following information):**
[...]

PR #2321, opened 2022-07-22
Add support for dynamic objects to kubectl apply.

brendandburns:
Addresses #2318 (I think)

@yue9944882 this is a sketch, let me know if you think it looks ok and I will add tests.

### DIFF
--- a/extended/src/main/java/io/kubernetes/client/extended/kubectl/Kubectl.java
+++ b/extended/src/main/java/io/kubernetes/client/extended/kubectl/Kubectl.java
@@ -22,6 +22,9 @@ import io.kubernetes.client.openapi.ApiException;
 import io.kubernetes.client.openapi.Configuration;
 import io.kubernetes.client.util.ModelMapper;
 import io.kubernetes.client.util.generic.GenericKubernetesApi;
+import io.kubernetes.client.util.generic.dynamic.DynamicKubernetesObject;
+import io.kubernetes.client.util.generic.dynamic.DynamicKubernetesListObject;
+
 
 /**
  * Kubectl provides a set of helper functions that has the same functionalities as corresponding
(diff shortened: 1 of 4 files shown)


extended/src/main/java/io/kubernetes/client/extended/kubectl/Kubectl.java
```diff
@@ -308,6 +311,26 @@ protected GenericKubernetesApi<ApiType, KubernetesListObject> getGenericApi()
...
     }
+
+    protected GenericKubernetesApi<? extends KubernetesObject, ? extends KubernetesListObject> getGenericApi(KubernetesObject targetObj)
+        throws KubectlException {
+      if (targetObj instanceof DynamicKubernetesObject) {
+        DynamicKubernetesObject obj = (DynamicKubernetesObject)targetObj;
+        String gv = obj.getApiVersion();
+        int ix = gv.indexOf("/");
+        String group = ix == -1 ? "" : gv.substring(0, ix);
+        String version = gv.substring(ix + 1);
```
yue9944882, line 322: https://github.com/kubernetes-client/java/blob/master/util/src/main/java/io/kubernetes/client/util/ModelMapper.java#L180-L190

i think we can promote the above code from `ModelMapper` to a utility helper function like `ParsedAPIVersion ModelMapper#parseAPIVersion(...)`, so we can reuse that here, wdyt?

brendandburns: Done.

brendandburns:
@yue9944882 updated per your comment. Please take another look. Once this looks good, I can add tests.

yue9944882:
@brendandburns overall looks good, lets add some tests. (and fix the format)

brendandburns:
Ok, I wrote tests and fixed formatting, this is ready for final review.

Thanks!

yue9944882 (commented):
/lgtm
/approve

PR #2354, opened 2022-08-30
Allow using non default context

omarkhammassi:
This PR adds this ability to specify which context to try to create configuration from.

### DIFF
--- a/util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
+++ b/util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
@@ -33,7 +33,6 @@ import io.kubernetes.client.util.exception.CSRNotApprovedException;
 import java.io.BufferedReader;
 import java.io.ByteArrayInputStream;
 import java.io.File;
-import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
@@ -97,20 +96,13 @@ public class ClientBuilder {
    *     cannot be read.
    */
   public static ClientBuilder standard() throws IOException {
-    return standard(true, null);
+    return standard(true);
   }
 
-  public static ClientBuilder standard(String context) throws IOException {
-    return standard(true, context);
-  }
-
-  public static ClientBuilder standard(boolean persistConfig, String context) throws IOException {
-    final File kubeConfig = findConfigFromEnv();
-    ClientBuilder clientBuilderEnv = getClientBuilder(persistConfig, kubeConfig, context);
-    if (clientBuilderEnv != null) return clientBuilderEnv;
-    final File config = findConfigInHomeDir();
-    ClientBuilder clientBuilderHomeDir = getClientBuilder(persistConfig, config, context);
-    if (clientBuilderHomeDir != null) return clientBuilderHomeDir;
+  public static ClientBuilder context(String context) throws IOException {
+    final File kubeConfig = getKubeConfigFile();
+    ClientBuilder clientBuilder = getClientBuilder(true, kubeConfig, context);
+    if (clientBuilder != null) return clientBuilder;
     final File clusterCa = new File(SERVICEACCOUNT_CA_PATH);
     if (clusterCa.exists()) {
       return cluster();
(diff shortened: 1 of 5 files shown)


util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
```diff
@@ -97,15 +97,19 @@ public static ApiClient defaultClient() throws IOException {
...
-    return standard(true);
+    return standard(true, null);
   }
 
-  public static ClientBuilder standard(boolean persistConfig) throws IOException {
+  public static ClientBuilder standard(String context) throws IOException {
+    return standard(true, context);
+  }
+
+  public static ClientBuilder standard(boolean persistConfig, String context) throws IOException {
```
brendandburns, line 107: Rather than change this signature, I think we should add a `context(String context)` method to `ClientBuilder`

util/src/main/java/io/kubernetes/client/util/KubeConfig.java
```diff
@@ -81,20 +81,36 @@ public static void registerAuthenticator(Authenticator auth) {
...
     ArrayList<Object> users = (ArrayList<Object>) configMap.get("users");
     Object preferences = configMap.get("preferences");
 
     KubeConfig kubeConfig = new KubeConfig(contexts, clusters, users);
-    kubeConfig.setContext(currentContext);
+
+    if (context != null) {
+      kubeConfig.setContext(context);
+    } else {
+      String currentContext = (String) configMap.get("current-context");
```
yue9944882, line 110: nit: promote `"currrent-context"` to a static final variable in this class

util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
```diff
@@ -114,14 +118,14 @@ public static ClientBuilder standard(boolean persistConfig) throws IOException {
...
-  private static ClientBuilder getClientBuilder(boolean persistConfig, File kubeConfig)
+  private static ClientBuilder getClientBuilder(boolean persistConfig, File kubeConfig, String context)
       throws IOException {
     if (kubeConfig != null) {
       try (BufferedReader kubeConfigReader =
           new BufferedReader(
               new InputStreamReader(
                   new FileInputStream(kubeConfig), StandardCharsets.UTF_8.name()))) {
-        KubeConfig kc = KubeConfig.loadKubeConfig(kubeConfigReader);
+        KubeConfig kc = KubeConfig.loadKubeConfig(kubeConfigReader, context);
```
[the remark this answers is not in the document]
yue9944882: +1 keep the parameter list minimal and neat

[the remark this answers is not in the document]
omarkhammassi: Done.

util/src/main/java/io/kubernetes/client/util/KubeConfig.java
omarkhammassi: Done.

util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
omarkhammassi: Done.

util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
```diff
@@ -100,28 +99,46 @@ public static ClientBuilder standard() throws IOException {
     return standard(true);
   }
 
+  public static ClientBuilder context(String context) throws IOException {
+    final File kubeConfig = getKubeConfigFile();
+    ClientBuilder clientBuilder = getClientBuilder(true, kubeConfig, context);
```
brendandburns, line 104: Don't change the signature here, just call `kubeconfig.setContext(context)` before you call `getClientBuilder()`

util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
```diff
@@ -100,28 +99,46 @@ public static ClientBuilder standard() throws IOException {
...
+    if (clientBuilder != null) return clientBuilder;
     final File clusterCa = new File(SERVICEACCOUNT_CA_PATH);
     if (clusterCa.exists()) {
       return cluster();
     }
     return new ClientBuilder();
   }
 
-  private static ClientBuilder getClientBuilder(boolean persistConfig, File kubeConfig)
+  private static File getKubeConfigFile() {
```
brendandburns, line 124: This is messy. We should really refactor the whole `ClientBuilder` class so that it defers all of this construction until the `build()` method is called.

That's beyond the scope of this PR though (unless you want to :)) but please add a `TODO: defer construction to build() method` here.

util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
```diff
@@ -100,28 +99,46 @@ public static ClientBuilder standard() throws IOException {
     return standard(true);
   }
 
+  public static ClientBuilder context(String context) throws IOException {
+    final File kubeConfig = getKubeConfigFile();
```
brendandburns, line 103: This isn't quite what I was expecting, I think that we should support `ClientBuilder.standard().context("foo").build()` as well as `ClientBuilder.kubeconfig(someFile).context("foo").build()`

I think what this means is that we need to save off the KubeConfig object in the builder rather than constructing it new each time.

util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
omarkhammassi: `kubeConfig` is of type `File` , `setContext(context)` can only be set at a later stage when building the client

brendandburns:
Closing this PR as stale, please feel free to re-open it if you plan to work on it further.

Issue #2361, opened 2022-09-07
Do not print exception stack trace through System.err 


dani8art:
This library makes use of `ex.printStackTrace()` at some point in its code where propagation of an exception is not possible, there is not surface where to place the exception or it can't be consumed. We understand sometimes it is needed and it helps debugging but this method uses `System.err` as default `PrintWriter` which is causing issues because dependant apps can't mute these lines in their logs.

The three main points where `ex.printStackTrace()` is used are the following
- `/util/src/main/java/io/kubernetes/client/Exec.java` https://github.com/kubernetes-client/java/blob/62d6c626ec1a2fd72f2da7e17a822f405b8445ac/util/src/main/java/io/kubernetes/client/Exec.java#L521
- `/extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlExec.java` https://github.com/kubernetes-client/java/blob/f20788272291c0e79a8c831d8d5a7dd94d96d2de/extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlExec.java#L76
- `/extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlPortForward.java` https://github.com/kubernetes-client/java/blob/f20788272291c0e79a8c831d8d5a7dd94d96d2de/extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlPortForward.java#L107

**Replacement proposal**

If we use `slf4j` api, we will be able to handle the visibility of these lines from outside the library. That way dependant apps may eventually hide these traces from their logs. We have made some tests and it is totally safe to replace `ex.printStackTrace()` by the following

```java
  @Test
  public void stacktrace() {
    try {
      throw new RuntimeException("error");
    } catch (RuntimeException e) {
      log.debug(e.getMessage(), e);
    }
  }
```

Resulting in the following lines that in this case can be hidden if we use a proper configuration in our apps.

```
10:08:10.355 [main] DEBUG  i.k.client.util.StackTraceTest - java.lang.RuntimeException: error
java.lang.RuntimeException: error
	at io.kubernetes.client.util.StackTraceTest.stacktrace(StackTraceTest.java:19)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
[...]

PR #2366, opened 2022-09-12
Add a consumer for unhandled errors to Exec.java

dani8art:
Signed-off-by: Daniel Arteaga Barba <darteaga@vmware.com>

This PR will try to implement an acceptable approach for the following proposal #2361

See an example of use.

```java
  @Override
  public Process exec(String[] args, boolean stdin, boolean tty) {
    try {
      Exec kubeExec = new Exec(client);
      kubeExec.setOnUnhandledError(this::logUnhandledExceptions);

      return kubeExec.exec(namespace, host, args, null, stdin, tty);
    } catch (ApiException | IOException e) {
      throw Errors.internal("Error executing command on the remote host", e).get();
    }
  }

  private void logUnhandledExceptions(final Throwable throwable) {
    log.debug(throwable.getMessage(), throwable);
  }
```

### DIFF
--- a/extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlExec.java
+++ b/extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlExec.java
@@ -21,6 +21,7 @@ import io.kubernetes.client.util.Streams;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.util.Optional;
 import java.util.function.Consumer;
 
 public class KubectlExec extends Kubectl.ResourceAndContainerBuilder<V1Pod, KubectlExec>
@@ -82,7 +83,7 @@ public class KubectlExec extends Kubectl.ResourceAndContainerBuilder<V1Pod, Kube
                 try {
                   Streams.copy(in, out);
                 } catch (IOException ex) {
-                  onUnhandledError.accept(ex);
+                  Optional.ofNullable(onUnhandledError).orElse(Throwable::printStackTrace).accept(ex);
                 }
               }
             });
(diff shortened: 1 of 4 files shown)


extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlExec.java
```diff
@@ -47,33 +49,40 @@ public KubectlExec tty(boolean tty) {
...
+  protected static Thread copyAsync(InputStream in, OutputStream out, Consumer<Throwable> onUnhandledError) {
     Thread t =
         new Thread(
             new Runnable() {
               public void run() {
                 try {
                   Streams.copy(in, out);
                 } catch (IOException ex) {
-                  ex.printStackTrace();
+                  onUnhandledError.accept(ex);
```
brendandburns, line 85:
```java
if (onUnhandledError != null) {
   ...
} else {
  ex.printStackTrace();
}
```

extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlPortForward.java
```diff
@@ -98,13 +105,13 @@ public void run() {
...
-                    Thread t1 = copyAsync(sock.getInputStream(), out);
-                    Thread t2 = copyAsync(in, sock.getOutputStream());
+                    Thread t1 = copyAsync(sock.getInputStream(), out, onUnhandledError);
+                    Thread t2 = copyAsync(in, sock.getOutputStream(), onUnhandledError);
 
                     t1.join();
                     t2.join();
                   } catch (InterruptedException | IOException ex) {
-                    ex.printStackTrace();
+                    onUnhandledError.accept(ex);
```
brendandburns, line 114: Same here as above.

extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlExec.java
dani8art: You're right! thanks!

extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlPortForward.java
dani8art: Done!

Issue #2396, opened 2022-09-29
DefaultRateLimitQueueTest is flaky


brendandburns:
```sh
Error:  Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.507 s <<< FAILURE! - in io.kubernetes.client.extended.workqueue.DefaultRateLimitQueueTest
Error:  testSimpleRateLimitQueue(io.kubernetes.client.extended.workqueue.DefaultRateLimitQueueTest)  Time elapsed: 0.507 s  <<< FAILURE!
java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:87)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.junit.Assert.assertTrue(Assert.java:53)
```

e.g. https://github.com/kubernetes-client/java/actions/runs/3148147767/jobs/5118376488

brendandburns:
I haven't seen this flake in a while, I'm closing it for now and hoping that 2397 fixed it :)

brendandburns:
Nevermind, it just flaked :(

```sh
Error:  Failures: 
Error:    DefaultRateLimitQueueTest.testSimpleRateLimitQueue:57 Unexpected time: 490385000 vs 500000000
Error:  Tests run: 100, Failures: 1, Errors: 0, Skipped: 0
Error:  Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.22.2:test (default-test) on project client-java-extended: There are test failures.
Error:  
Error:  Please refer to D:\a\java\java\extended\target\surefire-reports for the individual test results.
Error:  Please refer to dump files (if any exist) [date].dump, [date]-jvmRun[N].dump and [date].dumpstream.
Error:  -> [Help 1]
Error:  
Error:  To see the full stack trace of the errors, re-run Maven with the -e switch.
Error:  Re-run Maven using the -X switch to enable full debug logging.
Error:  
Error:  For more information about the errors and possible solutions, please read the following articles:
Error:  [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoFailureException
Error:  
Error:  After correcting the problems, you can resume the build with the command
Error:    mvn <args> -rf :client-java-extended
Error: Process completed with exit code 1.
```

brendandburns:
I'm going to close this again via #2435 we'll see if it reoccurs.

brendandburns:
Arg, it flaked again:

```
Error:  Failures: 
[...]

PR #2397, opened 2022-09-29
Improve the test to better debug flakiness (and maybe fix it)

brendandburns:
part of fixing #2396


yue9944882 (commented):
/lgtm
/approve

Issue #2398, opened 2022-09-29
please add a spring boot 3 aot support to enable graalvm use cases


joshlong:
hi 

tl;dr: my name is josh long. i work on the spring team at tanzu, a division of vmware. i would like to talk to you about contributing spring boot (3) AOT support, expanding the existing `spring` module.  i built something like this for the older 'spring native' project and i could easily turn it into a spring boot 3 aot module.  

longer story:

## spring framework 6, spring boot 3, and ahead-of-time compilation 
spring boot 3 (built on spring framework 6) is due sometime in november / december 2022. this year.  it integrates and greatly expands/refines a lot of the ahead-of-time processing that we prototyped in a separate experimental project called spring native. 

spring framework 6 has a new ahead-of-time compilation capability. it's basically a way to write code - regular components in your production spring code - that gets involved at _compile-time_ and can contribute files that get passed into the graalvm native image compiler.  spring doesn't run your full application at compile time. instead, after all the `.java` files have been compiled, the spring boot build plugin starts up a java process containing all the spring component configuration (such as comes from java configuration, component scanning, xml, etc.) and canonicalizes everything into the spring metamodel of `BeanDefinition`s, which are the things that - eventually - get turned into real 'beans' at runtime. it's at this stage, at compile-time, that spring involves your aot components. think of them like compile-time equivalents to `BeanFactoryPostProcessor` and `BeanPostProcessor`s. 

## graalvm and spring boot 3 aot

you can get an app turned into a graalvm native image thusly: `mvn -Pnative clean package` or `mvn -Pnative -DskipTests clean package`. this should work for many classes of applications out-of-the-box, assuming the 

graalvm requires configuration files (`META-INF/native-image/*json`) whenever you do anything at runtime that involves dynamic runtime behavior such as reflection, serialization, jdk proxies, jni, loading resources from the classpath, etc. 

we use this aot capability to transform a typical spring boot application and write out configuration files for graalvm native image compilation. this automatic mechanism works for all the things over which spring has control-any beans you create, for example. but third party libraries will need to ship their own configuration. you can contribute `.json` files directly in the library, which get picked up by the graalvm native image compiler. 

the trouble with those is that:
 * you need to make sure the string-y `.json` files stay in sync with the types in your codebase
 * there's no way to specify, as in a framework or library usecase, that all subclasses of a given type `T` should have certain configuration. 
 
[...]

PR #2402, opened 2022-10-01
add support for Spring Framework 6 AOT engine 

joshlong:
Hi 

Please find a proof-of-concept introduction of AOT support in Spring Boot 3, per #2398 .

I've tested the proposed changes and they work with both Spring Boot 3 and Spring Boot 2. 

Spring Boot 3 is not yet GA, however. So, i've had to add `snapshot` and `milestone` repositories. Obviously, this won't do in a library like the Kubernetes Java client which needs to be releasable all the time. I'm not sure how you want to handle this. Maybe we create a separate module that doesn't get released and that has the appropriate repositories? Maybe we just stash this PR until [November 24th](https://github.com/spring-projects/spring-boot/milestone/240), when Spring Boot 3 is due to go GA? (maybe a little later than that, who knows...) 
 
Also, for my contribution to work, I need to add a dependency on an external [library called Reflections](https://www.baeldung.com/reflections-library). This is used to capture all the code-generated types for CRDs, for example, in the Kubernetes API _and_ in the user's auto-configured packages in a Spring Boot application. This library is _only_ used at compile time, and _only_ when the user is using Spring Boot 3 AOT compilation. that's not the default. I don't know how to handle that, either. I suppose you could make the package `optional` and then make sure that the interested developer who wnats to use AOT processing gets a warning or something when they try it without that class on the classpath? I don't like that approach very much. It'd be nice if the library was just there and it worked for anybody who wants to build an AOT/ GraalVM native image, out-of-the-box.  

Is there a Maven `scope` i don't know about that limits the lifetime of the binary _explicitly_ to the time the AOT engine runs? 

I tried this new support out on [a sample Kubernetes CRD+Controller demo](https://github.com/kubernetes-native-java/controllers-101) and it works just fine, better than what I was able to do before with Spring Native, actually. From ac consumer's perspective, I didn't have to do anything to make code from the Kubernetes Java client do its magic. I was able to delete code, even. 

thanks for your time and consideration

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -40,6 +40,7 @@
     <maven-plugin-version>1.0.0</maven-plugin-version>
     <exec-maven-plugin.version>1.6.0</exec-maven-plugin.version>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+
     <javax.annotation.version>1.3.2</javax.annotation.version>
     <snakeyaml.version>1.33</snakeyaml.version>
     <slf4j.version>2.0.3</slf4j.version>
@@ -329,7 +330,16 @@
       <layout>default</layout>
     </repository>
   </distributionManagement>
-
+  <repositories>
+    <repository>
+      <id>spring-milestones</id>
+      <name>Spring Milestones</name>
+      <url>https://repo.spring.io/milestone</url>
+      <snapshots>
+        <enabled>false</enabled>
+      </snapshots>
+    </repository>
+  </repositories>
   <build>
     <pluginManagement>
       <plugins>
@@ -530,7 +540,6 @@
       </plugin>
     </plugins>
   </build>
-
   <profiles>
     <profile>
       <id>release-sign-artifacts</id>
--- a/pom.xml
+++ b/pom.xml
@@ -40,7 +40,7 @@
     <maven-plugin-version>1.0.0</maven-plugin-version>
     <exec-maven-plugin.version>1.6.0</exec-maven-plugin.version>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-
+    <reflections.version>0.10.2</reflections.version>
     <javax.annotation.version>1.3.2</javax.annotation.version>
     <snakeyaml.version>1.33</snakeyaml.version>
     <slf4j.version>2.0.3</slf4j.version>
@@ -89,6 +89,11 @@
 
   <dependencyManagement>
     <dependencies>
+      <dependency>
+        <groupId>org.reflections</groupId>
+        <artifactId>reflections</artifactId>
+        <version>${reflections.version}</version>
+      </dependency>
       <dependency>
         <groupId>org.apache.commons</groupId>
         <artifactId>commons-lang3</artifactId>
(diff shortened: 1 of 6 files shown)


pom.xml
```diff
@@ -482,49 +487,49 @@
               <include>src/test/groovy/**/*.groovy</include>
             </includes>
 
-            <importOrder>  <!-- or a custom ordering -->
```
brendandburns, line 485: Please remove these whitespace changes.

brendandburns, line 485: Please remove these whitespace changes.

spring/src/main/java/io/kubernetes/client/spring/extended/controller/config/aot/KubernetesBeanFactoryInitializationAotProcessor.java
```diff
@@ -0,0 +1,81 @@
...
+
+    private final MemberCategory[] allMemberCategories = MemberCategory.values();
+
+    @Override
+    public BeanFactoryInitializationAotContribution processAheadOfTime(
+            @NotNull ConfigurableListableBeanFactory beanFactory) {
+        return (generationContext, beanFactoryInitializationCode) -> {
+            RuntimeHints hints = generationContext.getRuntimeHints();
+            String[] classNames = new String[]{
+                    "com.google.gson.JsonElement",//
```
brendandburns, line 38: Is there a reason why there are empty comments after each of these lines?

brendandburns:
I took a look at the code changes and other than the whitespace changes which need to be reverted, this looks fine to me.

I will let @yue9944882 take a look at the Spring side and comment on how we want to sequence this into the released client.

pom.xml
joshlong: i believe this has now been addressed.

joshlong: i believe this has now been addressed.

joshlong:
Hi, thanks for the quick replies. somehow, the biggest change of my original PR - the fact that it needs a snapshot or milestone (non-Maven central) repository was not in the original `pom.xml`. Don't ask me how. So I took the code on `main`, applied it over mine, _re_-applied the changes without accidentally invoking my IDE's code reformatter, and here we are. 

I've updated the PR

brendandburns:
We already have a spring component (spring-boot-actuator) marked as `optional` https://github.com/kubernetes-client/java/blob/master/pom.xml#L177 so I think that for the reflections dependency, it makes sense mark it optional also, although I don't feel super strongly about this, but we've definitely gotten feedback about not wanting this library to pull in lots and lots of dependencies.

Regarding the milestone dependency, you're definitely right that we don't want to pull this in until it is released. So it feels like we should leave this PR open until it is released. There's not too much overlap w/ this PR so it shouldn't be too hard to keep it rebased.

joshlong:
Awesome! Both of those points - delaying until Boot 3 GA and marking Reflections as `optional` - are  reasonable and I appreciate them. I'll follow up sometime after Spring Boot 3 is GA. 

Maybe we could do something where, if you try to run the code and do an AOT compilation, it can check that reflections is on the classpath and if it isn't then warn the user to add it to their classpath? That shouldn't be too hard. Spring has a `boolean ClassUtils.isPresent (String className, ClassLoader)` we can put the warning behind. We'd have to update that one class in my PR. 

I'm pretty pleased how far this has come in just a week. Thanks for a great project, and I appreciate the fast turnarounds. Hopefully, people interested in doing so will soon (before 2023, maybe?) be able to build Kubernetes clients and controllers that take 50mb of RAM and startup In no time at all soon enough.

brendandburns:
Adding that warning seems totally fine to me.

Thanks

yue9944882:
> Error:  COMPILATION ERROR : 
Error:  /home/runner/work/java/java/spring/src/main/java/io/kubernetes/client/spring/extended/controller/config/KubernetesInformerProperties.java:[17,51] cannot access org.springframework.boot.context.properties.ConfigurationProperties
  bad class file: /home/runner/.m2/repository/org/springframework/boot/spring-boot/3.0.0-M5/spring-boot-3.0.0-M5.jar(org/springframework/boot/context/properties/ConfigurationProperties.class)
    class file has wrong version 61.0, should be 52.0
    
@joshlong the idea SGTM, but i noticed the error messages above in the build log. is spring-boot 3 still compatible with java 8? there's still a distinct portion of users actively using java 8, so the compatibility should be preserved as well.

joshlong:
> > Error:  COMPILATION ERROR : 
> 
> Error:  /home/runner/work/java/java/spring/src/main/java/io/kubernetes/client/spring/extended/controller/config/KubernetesInformerProperties.java:[17,51] cannot access org.springframework.boot.context.properties.ConfigurationProperties
> 
>   bad class file: /home/runner/.m2/repository/org/springframework/boot/spring-boot/3.0.0-M5/spring-boot-3.0.0-M5.jar(org/springframework/boot/context/properties/ConfigurationProperties.class)
> 
>     class file has wrong version 61.0, should be 52.0
> 
>     
> 
> @joshlong the idea SGTM, but i noticed the error messages above in the build log. is spring-boot 3 still compatible with java 8? there's still a distinct portion of users actively using java 8, so the compatibility should be preserved as well.

Hi - no, spring boot 3 requires a Java 17 baseline.

brendandburns:
I filed an issue so we can discuss how we want to handle Spring 6 and it's dependency on Java 17

https://github.com/kubernetes-client/java/issues/2452

yue9944882:
/close

closing it b/c #2457 was merged

OlgaMaciaszek:
Hi @brendandburns, when is a release containing these changes scheduled for?

PR #2435, opened 2022-11-02
Fix the default delay queue to use a fixed time source.

brendandburns:
The previous code uses `Instant.now()` which is subject to time changes from NTP or `adjtime(...)`

By switching to `nanoTime` which is only used for measuring duration we eliminate these potential issues.

This _may_ be why #2396 is still happening.

Note that this bug is still present in the upstream `client-go` I will send a fix there also.


yue9944882 (commented):
/lgtm

Issue #2440, opened 2022-11-04
Improve SSLUtils Private Key Loading


exceptionfactory:
**Describe the bug**
The `io.kubernetes.client.util.SSLUtils` class contains several `loadKey` methods that use a variety of approaches to read private keys from PEM-encoded bytes. The current implementation relies on the client key algorithm parsed from the PEM header to determine the parsing strategy, using the Bouncy Castle `PEMParser` for ECDSA, custom DER parsing for PKCS1 RSA, and attempting to use the Java `PKCS8EncodedKeySpec` for RSA or ECDSA encoded using PKCS8.

**Client Version**
`16.0.2`

**Kubernetes Version**
`1.24.0`

**Java Version**
Java 8

**To Reproduce**
NA

**Expected behavior**
The Bouncy Castle `PEMParser` supports both PKCS1 and PKCS8 encoding for all of private key algorithms listed, so the implementation could be streamlined to eliminate custom DER parsing and avoid the current try-catch approach for RSA and ECDSA PKCS8.

**KubeConfig**
NA

**Server (please complete the following information):**
NA

**Additional context**
I am interesting in contributing a pull request with these improvements.

exceptionfactory:
Closing based on merging of the linked pull requests

Issue #1724, opened 2021-05-22
when Load private key with  RSA, try PKCS1 failed，but not to try PKCS8;


watermelon-zai:
**Describe the bug**
package io.kubernetes.client.util.SSLUtils.class

private static PrivateKey loadKey(InputStream keyInputStream, String clientKeyAlgo){...}

In this method, when it is' RSA ', try to use pkcs1, but if it fails to use pkcs1 and throws an exception, it will not try to use pkcs8, but the method ends directly; Is it necessary try catch or other？

** Client Version **
10.0.0

** Kubernetes Version **
 `v1.18.3`

** Java Version **
Java 8

yue9944882:
https://en.wikipedia.org/wiki/PKCS_8

not sure i'm getting it correctly, there're two possible headers for PKCS8 format:

```
-----BEGIN PRIVATE KEY-----
```
or
```
-----BEGIN ENCRYPTED PRIVATE KEY-----
```

and none of them contains the word "RSA", isnt it? @watermelon-zai

PR #2441, opened 2022-11-04
Refactor SSLUtils.loadKey using PEMParser

exceptionfactory:
This pull request implements private key loading refactoring proposed in #2440.

The implementation refactors the `SSLUtils.loadKey` method to use the Bouncy Castle `PEMParser` for all types of private keys. The existing implementation uses `PEMParser` for `ECDSA` Private Keys, and the proposed changes eliminate the need for custom DER and PKCS8 parsing. Changes include updates to the SSLUtils test class, providing a reusable test method and also asserting expected algorithm values for parsed Private Keys.

### DIFF
--- a/util/src/main/java/io/kubernetes/client/util/SSLUtils.java
+++ b/util/src/main/java/io/kubernetes/client/util/SSLUtils.java
@@ -129,7 +129,8 @@ public class SSLUtils {
     try (final PEMParser pemParser = new PEMParser(new InputStreamReader(keyInputStream))) {
       final Object pemObject = pemParser.readObject();
       if (pemObject == null) {
-        final String message = String.format("PEM Private Key Algorithm [%s] not parsed", clientKeyAlgo);
+        final String message =
+            String.format("PEM Private Key Algorithm [%s] not parsed", clientKeyAlgo);
         throw new InvalidKeySpecException(message);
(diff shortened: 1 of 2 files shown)


brendandburns:
Thanks for the PR. A couple of questions:
First, I'm curious about the motivation for the PR. It looks like a general cleanup and that's appreciated, but did you run into problems with the other code?

Second can I request that this is split into two PRs:

1. the refactor to SSLUtils
2. the refactor to the unit tests

The reason for this is that I want to make sure that the existing tests pass with the refactored SSL code. I worry that in refactoring both the implementation and the test at the same time it is possible for behaviors to regress.

Thanks!

exceptionfactory:
Thanks for the quick reply @brendandburns!

I was reviewing the client code for its approach to loading private keys and came across issue #1724. I also noted the code comments with the try-catch approach to handling different algorithms with PKCS8. I haven't noticed any runtime issues, but I have worked on projects such as [sshj](https://github.com/hierynomus/sshj) to improve key loading.

I understand the concern about refactoring both the implementation and tests together. I will update this PR to remove the unit test updates and create a new PR for just those changes.

brendandburns:
/lgtm
/approve

(assuming tests pass :)

PR #2457, opened 2022-11-22
add a module for Spring Boot 3 AOT

joshlong:
this supercedes my last PR. in this example I've preserved Spring Framework 5 / Spring Boot 2 / Java 8 support in the `client-java-spring-integration` module, adding only one text file for Spring Boot 3 support. 

I've also created a new module, `client-java-spring-aot-integration`, that contributes GraalVm hints using Spring Boot 3's AOT engine. This module depends on Spring Framework 6 / Spring Boot 3 / Java 17 / the third party Reflections library. 

I've built a trivial controller here https://github.com/kubernetes-native-java/controllers-101 and linked it against the `-aot` module in this PR, and it works :)


spring-aot/src/main/java/io/kubernetes/client/spring/aot/KubernetesBeanFactoryInitializationAotProcessor.java
```diff
@@ -0,0 +1,91 @@
...
+
+    private final MemberCategory[] allMemberCategories = MemberCategory.values();
+
+    @Override
+    public BeanFactoryInitializationAotContribution processAheadOfTime(
+            @NotNull ConfigurableListableBeanFactory beanFactory) {
+        return (generationContext, beanFactoryInitializationCode) -> {
+            RuntimeHints hints = generationContext.getRuntimeHints();
+            String[] classNames = new String[]{
+                    "com.google.gson.JsonElement",//
```
yue9944882, line 46: these class names need to be secured by unit-test, in case the linked classes are renamed/moved in the future. we can do this in a follow-up.

spring-aot/pom.xml
```diff
@@ -0,0 +1,136 @@
...
+                    <source>17</source>
+                    <target>17</target>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+    <!--
+     todo :   we don't need this after 24 november, 2022
+    -->
+    <repositories>
```
yue9944882, line 99: nit:how about we lift/centralize these repository declarations to the parent pom?

spring-aot/pom.xml
```diff
@@ -0,0 +1,136 @@
...
+        <artifactId>client-java-parent</artifactId>
+        <groupId>io.kubernetes</groupId>
+        <version>16.0.0-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+    <dependencies>
+        <dependency>
+            <groupId>org.reflections</groupId>
+            <artifactId>reflections</artifactId>
+            <version>0.10.2</version>
```
yue9944882, line 20: nit: move the versioned dependency to the parent pom and let the children modules inherit.

yue9944882:
@joshlong the new module will block the compilation of [test jobs](https://github.com/kubernetes-client/java/blob/9c09e7c2184599f67c688dca44477b4379e59248/.github/workflows/maven.yml#L30) running on platforms older than java17, i think we need to explicitly exclude `spring-aot` from these compatibility tests (e.g. either by setting `--pl "!spring-aot"` when running maven cmd or declaring a new profile in the parent pom). and then we need to copy/paste a separate test job for covering`spring-aot` on java17 platform from the existing ones. i opened https://github.com/kubernetes-client/java/pull/2459 to show the expected edits on the github action configurations, can you incorporate/rebase the pull?

joshlong:
> can you run the formatter so that `Verify Source Format` job can pass?

what formatted?

spring-aot/pom.xml
joshlong: they won't be required after the 24th. no need to add them at all, afaict

spring-aot/pom.xml
joshlong: good idea

joshlong:
ive addressed most of the issues, i think. i hope this works :) I think you mean `mvn spotless:apply` for the formatted, but I'm not sure. i ran that, tho.

.github/workflows/maven.yml
```diff
@@ -22,12 +22,34 @@ jobs:
           java-version: 8.0.x
       - name: Verify Format and License
         run: mvn spotless:check
+  legacy-build:
```
brendandburns, line 25: I'd really rather not maintain two different build targets, it's a recipe for drift and cut/paste errors.

Is there a way that we can combine them together with a conditional depending on the java version, that should be possible.

pom.xml
```diff
@@ -561,5 +562,21 @@ limitations under the License.
         <module>fluent-gen</module>
       </modules>
     </profile>
+    <profile>
```
brendandburns, line 565: I don't know if it is possible in a pom, but is there a way to inherit this? I'd really prefer to avoid duplication if we can avoid it.

spring-aot/pom.xml
```diff
@@ -0,0 +1,136 @@
...
+            <artifactId>spring-test</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>org.springframework.boot</groupId>
+                <artifactId>spring-boot-dependencies</artifactId>
+                <version>3.0.0-SNAPSHOT</version>
```
brendandburns, line 52: Can we wait until this is released, I'd rather not depend on a snapshot version.

spring-aot/pom.xml
```diff
@@ -0,0 +1,136 @@
...
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.jacoco</groupId>
+                <artifactId>jacoco-maven-plugin</artifactId>
+                <version>0.8.8</version>
```
brendandburns, line 64: pull this version up to the parent pom (this may be a problem in all of our poms...)

spring-aot/pom.xml
brendandburns: Let's wait until after the 24th and get these removed.

yue9944882:
> I think you mean mvn spotless:apply for the formatted

@joshlong exactly, looks like all the tests passed now

spring-aot/pom.xml
joshlong: Absolutely ;) that'll be in about 24 hours...

.github/workflows/maven.yml
yue9944882: sure they can be merged into one conditional target, i updated https://github.com/kubernetes-client/java/pull/2459 to show how it works. @joshlong by any chance can you rebase again onto my latest commit?

pom.xml
yue9944882: i'm not sure about that, now it's already declared in the root-level pom. as far as i know, it's not inheriting from any other project/places.

spring-aot/pom.xml
yue9944882: let's switch to non-snapshot release after that

joshlong:
Hi - i just updated the PR to use Spring Boot 3 GA! Happy day! 

The PR removes all mentions of snapshots and snapshot repositories. I also avoid redundantly declaring the Spring Boot BOM in the child module, overriding the `spring-boot.version` property instead. 

Also, I am not sure if this was intentional or if I missed something in my earlier PR, but I noticed that the `spring-aot` module wasn't among those enumerated in the `<modules>` section in the top-most `pom.xml` so I added that.

.github/workflows/maven.yml
```diff
@@ -65,16 +43,23 @@ jobs:
           path: ~/.m2/repository
           key: ${{ runner.os }}-maven-${{ matrix.java }}-${{ hashFiles('pom.xml', '**/pom.xml') }}
       - name: Build with Maven
+        shell: bash
         run: |
-          mvn clean test -q -B --define=org.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn
+          if [ $(grep -E '^(8|11)\.' <<< '${{ matrix.java }}') ]; then
+            # some module doesn't compile on java platform lower than 17, need to skip them by specifying a profile
+            MODS_OVERRIDES='-pl !spring-aot'
```
yue9944882, line 50: excluding `spring-aot` for java8 and java11 build ^^^

yue9944882:
@brendandburns @joshlong i appended another commit to this thread so that the test workflow can pass after we bump java version to 17. the java 8/11 compatibility is meanwhile assured by https://github.com/kubernetes-client/java/pull/2457#discussion_r1033150820

joshlong:
whats the next step?

yue9944882:
@joshlong it looks like https://github.com/kubernetes-client/java/pull/2457/files#r1029896501 hasn't been resolved yet, can you move the plugin declaration to the root?

joshlong:
> @joshlong it looks like https://github.com/kubernetes-client/java/pull/2457/files#r1029896501 hasn't been resolved yet, can you move the plugin declaration to the root?

Hi - what version? The `spring.boot.version`? We don't want the global version to be 3.0, right? we need 2.x for everything _except_ this module. (or am I missing something?)

and the rest of the XML in that `pom.xml` is just a copy and paste of whatever is in your other `pom.xml`. I'm happy that this PR is contributing to the discussion around healthy refactorings to your build structure, but it's above my paygrade :) please do whatever you need to do to the build so that you're happy, I just want to see this thing merged so I can stop worrying about it. it's been many months, after all, since this discussion started.

yue9944882:
> and the rest of the XML in that pom.xml is just a copy and paste of whatever is in your other pom.xml

i think @brendandburns is talking about the version of the jacoco plugin, but yeah there're some other `pom.xml` duplicating the plugin declaration. we can lift them to the parent pom through out the project in a follow-up.

brendandburns:
/lgtm
/approve

We can do the pom cleanup in a different PR.

@joshlong it would be great if we could get your example controller merged into the examples directory.

joshlong:
I'll get the example controller cleaned up and send a separate PR. 

It was a lot of fun working with y'all so far. Thanks for allowing me the opportunity to contribute, even if ever so slightly, to your much-appreciated efforts.

PR #2465, opened 2022-11-24
Delete label

### LINKED ISSUE #2422: Kubectl delete label support
**Describe the bug**
Kubectl.label() does not have a delete option as far as I have seen. Normally in k8s, a label can be deleted via CLI by putting "-" at the end of the label.

Ex:
```kubectl label node <nodename> <labelname>=labelvalue``` adds the label, and
```kubectl label node <nodename> <labelname>-``` deletes the label

But in Kubectl.label() function cannot have "-" at the end of the label name due to some restrictions. Can it be achieved? 

By the way, I found a getaround with the jsonpatches and created a patch to the resource in order to achieve this functionality:

```
String patchString = "[{ \"op\": \"remove\", \"path\": \"/labelpath\" }]";
                V1Patch patch = new V1Patch(patchString);
                Kubectl.patch(V1Node.class).apiClient(apiClient).name(nodeName).patchContent(patch).execute();
```

shubhbapna:
Resolves #2422
- added `deleteLabel` method in the `KubectlLabel` class

### DIFF
--- a/extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlLabel.java
+++ b/extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlLabel.java
@@ -38,7 +38,7 @@ public class KubectlLabel<ApiType extends KubernetesObject>
   }
 
   public KubectlLabel<ApiType> deleteLabel(String key) {
-    this.addingLabels.put(key, null);
+    this.addingLabels.put(key, "null");
     return this;
   }
 
--- a/extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlLabel.java
+++ b/extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlLabel.java
@@ -37,6 +37,11 @@ public class KubectlLabel<ApiType extends KubernetesObject>
     return this;
   }
 
+  public KubectlLabel<ApiType> deleteLabel(String key) {
+    this.addingLabels.put(key, null);
+    return this;
+  }
+
   @Override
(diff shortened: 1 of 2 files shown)


brendandburns:
Thanks for the PR.

I don't actually think that the tests in this test (either the tests that you added, or the tests that were in the file) actually test anything other than the fact that network calls are being made.

Specifically it seems like we should be checking that the body of the PUT request looks correct (in all cases) and we don't do that.

If you wanted to fix these tests, that would be awesome, but I don't necessarily want to gate this PR on fixing tests that were not something you added, so I'm ok to merge this PR and then file an issue to get these tests fixed in a different PR.

Let me know what you think.

brendandburns:
cc @yue9944882

brendandburns:
I went ahead and fixed the existing tests in https://github.com/kubernetes-client/java/pull/2471

@shubhbapna can you add similar tests to this PR?

shubhbapna:
Yep will do!

shubhbapna:
Should I wait for #2471 to be merged, pull those changes and then fix the tests? Or should I fix them right now and resolve any merge conflicts later?

yue9944882:
@shubhbapna overall LGTM, and https://github.com/kubernetes-client/java/pull/2471 has been merged, can you rebase onto the latest?

yue9944882:
@shubhbapna can you run the formatter by `mvn spotless:apply` so that the test can pass?

shubhbapna:
@yue9944882 @brendandburns Rebased, fixed the tests and ran the formatter

PR #2471, opened 2022-11-28
Add validation that the request body is as expected in kubectl label.

brendandburns:
This fixes tests that weren't really testing much previously.


yue9944882 (commented):
/lgtm
/approve

PR #2483, opened 2022-12-05
Update release.yml

brendandburns:
Exclude the recent spring-aot work.

cc @yue9944882 @joshlong 

Ultimately we'll need to add another build target for the Java17 dependent module...

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -54,6 +54,7 @@ jobs:
           mvn -q \
             -Dexec.executable=echo \
             -Dexec.args='${project.version}' \
+            -pl !spring-aot \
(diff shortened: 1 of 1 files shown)


yue9944882:
@brendandburns i noticed the same failure in the uploading-snapshot workflow after we merged #2457 in. how about we fix the snapshot first progressively to verify the change and compatibility?

yue9944882:
@brendandburns @joshlong i opened https://github.com/kubernetes-client/java/pull/2484 as an alternative solution which bumps compiling JDK to 17 instead of skipping the new module explicitly

brendandburns:
Closing in favor of #2487

PR #2484, opened 2022-12-06
Workflow: Bumping snapshot workflow to JDK 17

yue9944882:


brendandburns:
I think that we need to cherry-pick this into the `release-17` branch.

PR #2505, opened 2022-12-22
Fix path for CRD GHA in docs

ddymko:
Going through the CRD docs I noticed that the link was invalid

### DIFF
--- a/docs/generate-model-from-third-party-resources.md
+++ b/docs/generate-model-from-third-party-resources.md
@@ -10,7 +10,7 @@ and [KubernetesListObject](https://github.com/kubernetes-client/java/blob/master
 ### 1. Fork Upstream Repo
 
 Fork the repository [kubernetes-client/java](https://github.com/kubernetes-client/java)
-so that you can run the github action ["CRD Java Model Generate"](./.github/workflows/generate-crd.yml)
+so that you can run the github action ["CRD Java Model Generate"](/.github/workflows/generate-crd.yml)
 in your forked repo. (Running github action job manually requires "collaborator" access to the repo).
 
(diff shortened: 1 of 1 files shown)


docs/generate-model-from-third-party-resources.md
```diff
@@ -10,7 +10,7 @@ and [KubernetesListObject](https://github.com/kubernetes-client/java/blob/master
 ### 1. Fork Upstream Repo
 
 Fork the repository [kubernetes-client/java](https://github.com/kubernetes-client/java)
-so that you can run the github action ["CRD Java Model Generate"](./.github/workflows/generate-crd.yml)
+so that you can run the github action ["CRD Java Model Generate"](/.github/workflows/generate-crd.yml)
```
brendandburns, line 13: I don't think this actually works correctly. I think the right link is `/kubernetes-client/java/blob/master/.github/workflows/generate-crd.yml`

ddymko: This is working fine on my fork/branch - it looks like it appends the link root path of the project.

https://github.com/ddymko/kubernetes-client-java/blob/ddymko/fix-doc-crd-path/docs/generate-model-from-third-party-resources.md

I did try appending `/kubernetes-client/java/blob/master/.github/workflows/generate-crd.yml` and this will append the a path that includes this.

`https://github.com/ddymko/kubernetes-client-java/blob/ddymko/fix-doc-crd-path-b/kubernetes-client/java/blob/master/.github/workflows/generate-crd.yml`

brendandburns: ok, the markdown -> href must be doing something funky. I will merge this and we'll hope it works :)

PR #2524, opened 2023-01-16
fix  client's timeout  when vip failover

coderYii:
fix when vip failover(base on keepalived+haproxy)，the informer watch data is not updated.

### DIFF
--- a/util/src/main/java/io/kubernetes/client/informer/SharedInformerFactory.java
+++ b/util/src/main/java/io/kubernetes/client/informer/SharedInformerFactory.java
@@ -76,9 +76,6 @@ public class SharedInformerFactory {
    * @param threadPool specified thread pool
    */
   public SharedInformerFactory(ApiClient client, ExecutorService threadPool) {
-    if (client.getReadTimeout() != 0) {
-      throw new IllegalArgumentException("read timeout of ApiClient must be zero");
-    }
 
     apiClient = client;
     informerExecutor = threadPool;


brendandburns:
cc @yue9944882 we've been back and forth on this a few times. I'm not opposed to this change necessarily, but the idea is that a watch should persist as long as possible even if no data is flowing.

yue9944882:
i think we can make this configurable so users can opt-out to their own read timeout instead of just removing the defensive timeout check.

brendandburns:
I think if we remove this check, it will be configurable, since it will be whatever is passed in with the `apiClient` or is there an alternative approach that you'd prefer?

yue9944882:
the default read timeout for apiClient is `0`, so now the assertion will fail if a user overrides a non-zero read timeout here. if i remember correctly, the point of having a zero assertion here is to prevent users from restarting watch connection too frequently e.g. a 30s read timeout means that the informer will refresh the watch connection and the local cache every 30s which can be a big burden for the client memory. due to the limitation of okhttp library, the apiClient cannot set different read timeouts for short-running and long-running api calls, generally, a watch call is expected to last about 5-10 minutes, so as an alternative, we can relax the assertion to `readTimeout >= 10min`, @coderYii do you think that will work for your case?

coderYii:
@yue9944882  now i solve it by rewriting the constructor. If the startup log can increase the timeout warning, i think programers should decide how to set the timeout according to the production environment.

PR #2540, opened 2023-01-27
Add asynchronous methods to GenericKubernetesApi

rjeberhard:
This change updates GenericKuberentesApi to have asynchronous methods similar to each of the existing synchronous methods. I've chosen to have the callback be an implementation of Consumer<KubernetesApiResonse<ApiType>> and have each method return Future<KubernetesApiResponse<ApiType>> to continue the pattern of not exposing symbols from OkHttp.

Second, and much more minor, I was having a lot of trouble building locally because of "port already in use" messages from WireMock. I saw that the wireMockConfig().dynamicPort() pattern was already in use elsewhere in the code base so I've updated more test classes.

### DIFF
--- a/util/src/main/java/io/kubernetes/client/util/generic/GenericKubernetesApi.java
+++ b/util/src/main/java/io/kubernetes/client/util/generic/GenericKubernetesApi.java
@@ -49,6 +49,7 @@ import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 import java.util.function.Function;
+import java.util.function.UnaryOperator;
 import javax.annotation.Nonnull;
 import okhttp3.Call;
 import okhttp3.HttpUrl;
@@ -161,6 +162,11 @@ public class GenericKubernetesApi<
     this.apiTypeClass = apiTypeClass;
     this.apiListTypeClass = apiListTypeClass;
     this.customObjectsApi = customObjectsApi;
+  }
+
+  // For unit-test
+  ApiClient getApiClient() {
+    return customObjectsApi.getApiClient();
   }
 
   /**
@@ -1258,13 +1264,32 @@ public class GenericKubernetesApi<
     }
   }
 
+  // For unit-test
+  interface CallAdapter extends UnaryOperator<Call> {
+
+  }
+
+  private <DataType extends KubernetesType> Call prepareCall(CallBuilder callBuilder,
+                                                             Consumer<KubernetesApiResponse<DataType>> callback)
+      throws ApiException {
+    Call call = prepareCall(callBuilder);
+
+    // For unit-test
+    if (callback instanceof CallAdapter) {
+      call = ((CallAdapter) callback).apply(call);
+    }
+
+    return call;
+  }
+
   private <DataType extends KubernetesType> Future<KubernetesApiResponse<DataType>> executeCallAsync(
       ApiClient apiClient, Class<DataType> dataClass, CallBuilder callBuilder,
       Consumer<KubernetesApiResponse<DataType>> callback) {
     try {
-      Call call = prepareCall(callBuilder);
+      Call call = prepareCall(callBuilder, callback);
       CountDownLatch latch = new CountDownLatch(1);
       AtomicReference<KubernetesApiResponse<DataType>> result = new AtomicReference<>();
+
       apiClient.executeAsync(call, JsonElement.class, new ApiCallback<JsonElement>() {
         @Override
         public void onFailure(ApiException e, int statusCode, Map<String, List<String>> responseHeaders) {
--- a/util/src/main/java/io/kubernetes/client/util/generic/GenericKubernetesApi.java
+++ b/util/src/main/java/io/kubernetes/client/util/generic/GenericKubernetesApi.java
@@ -198,7 +198,7 @@ public class GenericKubernetesApi<
    * @return future for the kubernetes api response
    */
   public Future<KubernetesApiResponse<ApiType>> getAsync(String name,
-                                                         Consumer<KubernetesApiResponse<ApiType>> callback) {
+                                                         @Nonnull Consumer<KubernetesApiResponse<ApiType>> callback) {
     return getAsync(name, new GetOptions(), callback);
   }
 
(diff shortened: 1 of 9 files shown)


rjeberhard:
Some of the integration tests appear to have intermittent failures. How can I trigger another run?

rjeberhard:
@yue9944882, I realize that this is a big change. I've tried to leave as much of the synchronous behavior unchanged. For instance, I moved the behavior that generates a CallBuilder to a method so that it could be shared by the asynchronous variants, but didn't otherwise change the synchronous methods (or tests). The exception was the patch methods. These were previously using PatchUtils, but I updated the method to match the pattern of the other operations.

If it would be better to do a live code review then I'm happy to schedule a Zoom or Teams meeting.

util/src/main/java/io/kubernetes/client/util/generic/GenericKubernetesApi.java
```diff
@@ -789,30 +1277,161 @@ KubernetesApiResponse<DataType> getKubernetesApiResponse(
...
+          });
+
+      return new Future<KubernetesApiResponse<DataType>>() {
+        @Override
+        public boolean cancel(boolean mayInterruptIfRunning) {
+          if (isDone()) {
+            return false;
+          }
+          call.cancel();
+          boolean isCanceled = call.isCanceled();
```
yue9944882, line 1393: i'm not sure if `call#cancel()` is a synchronous/blocking process. it's not, then calling `call#isCancel()` right after `cancel()` will mostly always return false, wdyt?

util/src/main/java/io/kubernetes/client/util/generic/GenericKubernetesApi.java
```diff
@@ -789,30 +1277,161 @@ KubernetesApiResponse<DataType> getKubernetesApiResponse(
...
+  // For unit-test
+  interface CallAdapter extends UnaryOperator<Call> {}
+
+  private <DataType extends KubernetesType> Call prepareCall(
+      CallBuilder callBuilder, @Nonnull Consumer<KubernetesApiResponse<DataType>> callback)
+      throws ApiException {
+    Call call = prepareCall(callBuilder);
+
+    // For unit-test
+    if (callback instanceof CallAdapter) {
```
yue9944882, line 1330: i always have a complex feeling about having these unit-test-only code in the main path. as for this `CallAdapter` interface, the mere usage i found is a `TestCallback` which is for retrieving response data from the client, and if i understand correctly they can be replaced by `Future#get()`, so we dont `CallAdapter` here, is that correct?

yue9944882 (commented):
thanks for adding this feature and fixing those details! the critical changes are actually smaller than this pull looks like, so overall i think LGTM, but it looks like some test-only code can be pruned from the main path.

util/src/main/java/io/kubernetes/client/util/generic/GenericKubernetesApi.java
rjeberhard: The unit-test code (GenericKubernetesApiForCoreApiTest, line 552) needs to modify the Call before it is executed. I did not want to expose the Call type (or any of the OkHttp types) through the GenericKubernetesApi API. 

Because the Call needs to be modified before it is executed or enqueued, this call can't be replaced with Future#get().

If you would prefer, I could expose a supported API to modify the Call before it is used. I've had good success with unit-test specific code that is package private.

util/src/main/java/io/kubernetes/client/util/generic/GenericKubernetesApi.java
rjeberhard: Oh, I understand what you are saying... My expectation is that `cancel()` is synchronous and that the Call is really canceled if it can be. The implementation of the concrete class is in Kotlin and looks like this:
```
  /**
   * Immediately closes the socket connection if it's currently held. Use this to interrupt an
   * in-flight request from any thread. It's the caller's responsibility to close the request body
   * and response body streams; otherwise resources may be leaked.
   *
   * This method is safe to be called concurrently, but provides limited guarantees. If a transport
   * layer connection has been established (such as a HTTP/2 stream) that is terminated. Otherwise
   * if a socket connection is being established, that is terminated.
   */
  override fun cancel() {
    if (canceled) return // Already canceled.

    canceled = true
    exchange?.cancel()
    connectionToCancel?.cancel()

    eventListener.canceled(this)
  }
```

rjeberhard: With this implementation, `isCanceled()` will always return `true`. I was more concerned with catching the rare case (different implementation class) where the call could not be canceled rather than a race-condition with an asynchronous version of `cancel()`.

I don't feel that strongly about this, so if you prefer, I can just have the method return `true` once `cancel()` is called.

rjeberhard:
@yue9944882, I've modified the implementation of the Future to use an AtomicBoolean rather than calling Call.isCanceled(). 

I need to keep the ability for the unit-test to modify the Call, but we can discuss a different mechanism. The goal of this test relies on being able to call one of the async methods and know that the request has not yet been made so that the cancel and done related fields and methods can be tested. Presently, this is done by giving the test class a chance to modify the generated Call object before it is enqueued. The Call is modified to add a tag which can then be read by the interceptor. Now, the interceptor can block on a latch that is shared with the test code. The test code can then absolutely know the the request has not yet been made.

I can imagine an alternate API where any caller (not just a unit-test) is given a mechanism to inspect and modify the Call instance before the request is made.  Is that okay or can you think of a different approach?

util/src/main/java/io/kubernetes/client/util/generic/GenericKubernetesApi.java
yue9944882: thanks for clarifying the behavior of `RealCall#cancel()`. the current implementation should work if that is a synchronous function.

util/src/main/java/io/kubernetes/client/util/generic/GenericKubernetesApi.java
yue9944882: > The unit-test code (GenericKubernetesApiForCoreApiTest, line 552) needs to modify the Call before it is executed.

oh i missed that unit-test. ideally it would be nice if there're other workaround to remove those test-only code. but we can continue to refactor them in follow-ups when we have a better idea.

yue9944882:
> I can imagine an alternate API where any caller (not just a unit-test) is given a mechanism to inspect and modify the Call instance before the request is made. Is that okay or can you think of a different approach?

maybe we can add a new member-variable to apiClient, e.g. a chain of call modifiers, then in each of those unit-test, we can instantiate a new apiClient and insert a test-scoped call modifier for the test purpose. anyway, i'm getting this pull in. thanks for the contribution!

/lgtm
/approve

PR #2550, opened 2023-02-03
Support PartialObjectMetadata requests

rjeberhard:
The API server supports requests that efficiently check resources by only returning the `metadata` stanza. These PartialObjectMetadata requests are described here: https://github.com/kubernetes/website/issues/37448.

This change adds support for PartialObjectMetadata to the GenericKubernetesApi.


util/src/main/java/io/kubernetes/client/util/generic/options/CreateOptions.java
```diff
@@ -16,6 +16,11 @@ public class CreateOptions {
   private String dryRun;
   private String fieldManager;
 
+  public CreateOptions dryRun(String dryRun) {
```
brendandburns, line 19: This change feels like it should be a different PR?

kubernetes/src/main/java/io/kubernetes/client/common/PartialObjectMetadata.java
```diff
@@ -0,0 +1,108 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package io.kubernetes.client.common;
+
+import io.kubernetes.client.openapi.models.V1ObjectMeta;
+import java.util.Objects;
+import javax.annotation.Nullable;
+
+public class PartialObjectMetadata implements KubernetesObject {
```
brendandburns, line 19: Should we templatize this for clarity? so that someone can declare a:

`PartialObjectMetadata<V1Pod>` or somesuch, so that you can really easily convert it into it's actual object?

util/src/main/java/io/kubernetes/client/util/generic/options/CreateOptions.java
rjeberhard: I can do that... The reason I did this is that I was introducing methods to GetOptions and ListOptions and realized that none of them had builder-style methods and so I added them to all of the *Options classes.

brendandburns:
Thanks for the PR! Can you pull out the various *Options.java changes into a separate PR? Much of that change feels orthogonal to adding PartialObjectMetadata.

rjeberhard:
@brendandburns, PR created https://github.com/kubernetes-client/java/pull/2551

kubernetes/src/main/java/io/kubernetes/client/common/PartialObjectMetadata.java
brendandburns: Yeah, but I was imagining a world where you got the metadata but wanted to pass it off to something that knew how to print Pods or somesuch, it's definitely not required, but I kind of like the idea of knowing what kind of thing I have gotten the Metadata for from a type-safety perspective...

brendandburns:
I thought about this some more. Why do we even need the PartialObjectMetadata and PartialObjectMetadataList objects?

I feel like you could do `GenericKubernetesApi<V1Pod>.get(...)` with GetOptions that request partial object metadata and it should work correctly, right?

rjeberhard:
> I thought about this some more. Why do we even need the PartialObjectMetadata and PartialObjectMetadataList objects?

Okay, that makes sense. I was thinking to be most explicit but you make a very good point. Let me try that experiment and I'll update the PR.

rjeberhard:
@brendandburns, yes, I can remove the PartialObjectMetadata and *List classes and use the specific types.  For example, the following works correctly:
```
    ApiClient apiClient = ClientBuilder.standard().build();
    GenericKubernetesApi<V1CustomResourceDefinition, V1CustomResourceDefinitionList> crdClient =
        new GenericKubernetesApi<>(V1CustomResourceDefinition.class, V1CustomResourceDefinitionList.class,
            "apiextensions.k8s.io", "v1", "customresourcedefinitions", apiClient);

    V1CustomResourceDefinitionList crds =
        crdClient.list(new ListOptions().isPartialObjectMetadataListRequest(true)).throwsApiException().getObject();
```

The only "oddity" is that the kind fields are set to "PartialObjectMetadata" and "PartialObjectMetadataList", respectively. I think that's fine.

kubernetes/src/main/java/io/kubernetes/client/common/PartialObjectMetadata.java
rjeberhard: Done... I've removed POM and POML classes. You are right that you can just use the original, specific classes.

brendandburns:
/lgtm
/approve

Thanks for the PR!

rjeberhard:
@brendandburns, what's the time frame for 18.0.0 or may I backport my recent changes to the maintenance branch for 17.0.x (whichever that branch is) and request a 17.0.2? Thanks.

PR #2551, opened 2023-02-03
Builder methods for *Options classes

rjeberhard:


brendandburns:
/lgtm
/approve

Thanks for the quick turn around. One nit: Does Java really require the underscore for `_continue()` in a method name? I'm sort of surprised that is true I thought the compiler was smarter.

rjeberhard:
> Thanks for the quick turn around. One nit: Does Java really require the underscore for _continue() in a method name? I'm sort of surprised that is true I thought the compiler was smarter.

Yes, it's really required :(

PR #2553, opened 2023-02-08
Fix `DynamicKubernetesTypeAdaptorFactory` when it does not apply.

eamonnmcmanus:
When a `TypeAdapterFactory` does not handle the type in its `create` method, the protocol is for it to return `null`. Calling `gson.getDelegateAdapter` is only appropriate for factories that wrap the results of other factories.

See https://github.com/google/gson/issues/2312.


brendandburns:
/lgtm
/approve

Relevant docs are here:
https://javadoc.io/doc/com.google.code.gson/gson/latest/com.google.gson/com/google/gson/TypeAdapterFactory.html

Issue #2509, opened 2022-12-31
Controllers and LeaderElector uses non-daemon threads by default


Prototik:
**Describe the bug**
I noticed my app won't stop after `main` is exited, thread dump says there are live threads: `leader-elector-*`.
I searched this problem via issues and it surprised me as it got fixed long time ago: #873
But was broken in 3f542d9c50c2864f7581dde847e44ec150eff3b8.
I don't know what @dsyer means with comment `Daemon status inherited from default`, but docs for [`Executors.defaultThreadFactory`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/Executors.html#defaultThreadFactory()) explicitly says it creates non-daemon threads by default:
`Each new thread is created as a non-daemon thread...`

**Client Version**
`17.0.0`

**Java Version**
Java 17

brendandburns:
Seems like we should provide the ability for a user to inject their own thread factory if they want to so that you can chose what kind of threads you want.

We'd be happy to take a PR to add this customizability to the `Threads` class

nervoussidd:
Hey can you assign me this issue? i would love to contribute.

Sajiyah-Salat:
Can someone explain me this pr. I would love to work on  that?

swastik959:
hi i am new to open source and want to contribute to this issue can you guide me through it

swastik959:
@brendandburns hi I have submitted a pr please review it

Aman254:
Hey Can anyone please invite me to the kubernetes slack work space 

My  Email:  utsav.kumar.3639@gmail.com

mahehsblogs:
Hey, I am new to Open Source. Is this issue still open?

PR #2557, opened 2023-02-12
added the ability for users to set their own thread

swastik959:
**added the ability for users to set their own thread**

I have made a function to set the thread set by user 

solves #2509 

*if any discrepancy is there feel free to point out*

### DIFF
--- a/util/src/main/java/io/kubernetes/client/util/Threads.java
+++ b/util/src/main/java/io/kubernetes/client/util/Threads.java
@@ -24,10 +24,9 @@ public class Threads {
     public static void setDefaultThreadFactory(ThreadFactory factory) {
       defaultFactory = factory;
   }
-    final ThreadFactory factory = defaultFactory;
     final AtomicInteger threadNumber = new AtomicInteger(1);
     return r -> {
-      Thread thread = factory.newThread(r);
+      Thread thread = defaultFactory.newThread(r);
       // Daemon status inherited from default
       thread.setName(String.format(format, threadNumber.getAndIncrement()));
       return thread;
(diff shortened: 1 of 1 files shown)


util/src/main/java/io/kubernetes/client/util/Threads.java
```diff
@@ -20,9 +20,14 @@ public class Threads {
 
   public static ThreadFactory threadFactory(String format) {
     final ThreadFactory defaultFactory = Executors.defaultThreadFactory();
+    // Users can inject their own thread 
+    public static void setDefaultThreadFactory(ThreadFactory factory) {
+      defaultFactory = factory;
```
brendandburns, line 25: this doesn't work correctly because you use `factory` below.

If you remove the change on line 27 (`final ThreadFactory factory ...`) and also the change on line 30 (`Thread thread = factory.newThread(r);`) then I think this code will work correctly.

However, please add a unit test to confirm the behavior is working as intended.

brendandburns:
Thank you for the PR! I have made a couple of comments.

swastik959:
@brendandburns can you tell me the location of the test file please

brendandburns:
Looks like we don't have a test file for this class :( 

I won't make you add one for such a small change, but if you want to send a second PR to add a test, it would be `ThreadsTest.java` in this directory: 

https://github.com/kubernetes-client/java/tree/master/util/src/test/java/io/kubernetes/client/util

brendandburns:
Looks like CI/CD is failing, can you take a look at the tests?

swastik959:
@brendandburns I have opened a new pr you can review it and have also made a pr for the test cases

PR #2588, opened 2023-03-05
Update spotless testing to use JDK 11

brendandburns:
Spotless has moved to only support JDK 11+, see #2579


yue9944882 (commented):
/lgtm
/approve

PR #2598, opened 2023-03-20
Use synchronization classes instead of Thread.sleep in test cases

niteshs7:
Use synchronization classes instead of Thread.sleep in test cases

PR is to fix one of the issue under this PR umbrella #1223


brendandburns:
/lgtm
/approve

Thank you for this PR!

PR #2601, opened 2023-03-21
Controllers and LeaderElector uses non-daemon threads by default #2509

mahehsblogs:
I tried to solve this issue "Controllers and LeaderElector uses non-daemon threads by default #2509". Explicitly set daemon status to false.


brendandburns:
@mahehsblogs thanks for your interest and sending a PR, but actually a previous PR added the ability to inject your own thread factory which solves the problem. We should have closed issue #2509 that was an oversight. Apologies for the confusion!

We would love for you to contribute to this client with a different issue.

Issue #2436, opened 2022-11-02
DefaultSharedIndexInformerWireMockTest is flaky


brendandburns:
```
Error:  Tests run: 5, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 14.683 s <<< FAILURE! - in io.kubernetes.client.informer.impl.DefaultSharedIndexInformerWireMockTest
Error:  testNamespacedPodInformerNormalBehavior(io.kubernetes.client.informer.impl.DefaultSharedIndexInformerWireMockTest)  Time elapsed: 2.702 s  <<< FAILURE!
java.lang.AssertionError: expected:<true> but was:<false>
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotEquals(Assert.java:835)
	at org.junit.Assert.assertEquals(Assert.java:120)
	at org.junit.Assert.assertEquals(Assert.java:146)
	at io.kubernetes.client.informer.impl.DefaultSharedIndexInformerWireMockTest.testNamespacedPodInformerNormalBehavior(DefaultSharedIndexInformerWireMockTest.java:153)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
```

abhishekS14:
Hi! Is anyone investigating this? I can check it otherwise.

PR #2608, opened 2023-03-24
Use synchronization classes instead of Thread.sleep and fix flaky test

niteshs7:
This PR removes thread.sleep and uses synchronization classes similar to fix made for issues under umbrella #1223 
This should also fix flaky test mentioned in #2436


brendandburns:
/lgtm
/approve

Thanks for the PR! (and special thanks for a flaky test fix!)

PR #2616, opened 2023-04-02
changed the annotation from javax to jakarta

Mohdcode:
i have made changes in all file in which I've changed the namespace  as it was mentioned for only annotation ->
 from javax.annotation to jakarta.annotation


brendandburns:
Unfortunately, all of this code is generated. If we want to make the switch we need to make it in the upstream code generator here: https://github.com/kubernetes-client/gen/blob/master/openapi/java.xml#L47 by adding a `javaxPackage` field and regenerating the code.

fwiw, this may not be as easy a "first issue" as I thought.

If you want, you can just send a PR for all of the references that are _not_ in the `kubernetes` directory. That code is hand-written and therefore a PR to switch them can just be merged.

Mohdcode:
sorry it didn't help.

PR #2617, opened 2023-04-03
tweak graceful shutdown of LeaderElector

maediy:
I'm going to update LeaderElector not to emit error logs while graceful shutdown.

I got the error log below when I restart an application using LeaderElector even if I gracefully shutdown it.
https://github.com/kubernetes-client/java/blob/39a29b92345698a91c8717776dbcc5ad4e7acf86/extended/src/main/java/io/kubernetes/client/extended/leaderelection/LeaderElector.java#L209

The reason is that executors were closed by `shutdownNow` method.
So I updated codes to use `shutdown` method instead and await termination for a while.


brendandburns:
cc @yue9944882

yue9944882 (commented):
/lgtm
/approve

PR #2623, opened 2023-04-07
Lobocode patch 1

lobocode:
This pull request includes a series of changes to address the issues and suggestions pointed out by the SonarLint analysis. Each change has been committed with an appropriate commit. These changes include improvements such as code simplification, use of Optional instead of null, more informative Javadoc tags, among others.

The refactoring of the findApiResourceByGroupVersionKind method involved creating a new HashMap to store API resources by their kind and group. This new map allows for faster lookup of resources during the method's execution, improving the overall performance. The previous implementation required iterating through all resources until a match was found, resulting in a slower search. The new implementation leverages the HashMap data structure, which provides faster access to resources by key. This refactoring improves the code's maintainability, readability, and efficiency.

### DIFF
--- a/util/src/main/java/io/kubernetes/client/util/ModelMapper.java
+++ b/util/src/main/java/io/kubernetes/client/util/ModelMapper.java
@@ -30,6 +30,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
 
+import org.apache.commons.lang3.tuple.ImmutablePair;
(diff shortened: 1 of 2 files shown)


brendandburns:
This PR looks like it overlaps with your other PR. Can you either separate them out, or we can wait until the other PR merges and then review this. Your call.

One thing about linting like this, I think it is only truly useful if we set it up in the CI/CD system so that we stay consistent going forward.

We'll need to take a look at SonarLint (which I don't have any experience with) and see if it makes sense to make it a permanent part of this project.

PR #2626, opened 2023-04-10
Use synchronization classes instead of Thread.sleep in test cases

tmrpavan:
Use synchronization classes instead of Thread.sleep in test cases

[Umbrella Issue] Many tests use Thread.sleep instead of synchronization primitives https://github.com/kubernetes-client/java/issues/1223

### DIFF
--- a/util/src/test/java/io/kubernetes/client/informer/cache/SharedProcessorTest.java
+++ b/util/src/test/java/io/kubernetes/client/informer/cache/SharedProcessorTest.java
@@ -12,16 +12,17 @@ limitations under the License.
 */
 package io.kubernetes.client.informer.cache;
 
-import static org.junit.Assert.*;
-
 import io.kubernetes.client.common.KubernetesObject;
 import io.kubernetes.client.informer.ResourceEventHandler;
 import io.kubernetes.client.openapi.models.V1ObjectMeta;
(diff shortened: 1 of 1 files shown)


tmrpavan:
@brendandburns  Can you review my PR

brendandburns:
In this context, I'm not sure that sleep is a problem. We need to wait for a period of time, but we expect to be interrupted. Adding the `lock.wait(10s)` instead of `sleep(10s)` doesn't really change the expectation of an interrupt and I don't think it makes this test less flaky.

You could, in theory, make this test less flaky by blocking the thread until just before shutdown is called, and _then_ waiting for a period of time to get interrupted by shutdown, but I don't think it's worth it.

brendandburns:
(oh and many thanks for taking on these type of PRs, they're generally useful and hard to get contributions, so thanks! it's just that this one may not be quite as useful as others)

tmrpavan:
@brendandburns Thanks for your comments. I will work on this

tmrpavan:
@brendandburns Can you please review this PR

brendandburns:
I don't think that this change is really any different than the code that is already there. I think in this case, Thread.sleep(...) is actually the right thing to do in this test.

Let me know if I'm missing something.

Thanks!

PR #2628, opened 2023-04-10
Lobocode refactor02

lobocode:
In this WebSocketStreamHandler refactored version, I extracted several methods to break down the complexity of the original write() method. Each method has a specific responsibility, making the code easier to understand and maintain. 

 I've wrapped the waiting logic inside a while loop that checks whether the WebSocketStreamHandler.this.socket is still null. This change ensures that the method handles spurious wakeups correctly.  I've used computeIfAbsent() to create and return a new WebSocketOutputStream if the key (stream) is not present in the output map. If the key is already present, the method returns the corresponding value.   I've wrapped the waiting logic inside a while loop that checks whether the state is still State.UNINITIALIZED. This ensures that the method handles spurious wakeups correctly, as recommended by the documentation. Additionally, I've added Thread.currentThread().interrupt(); to set the interrupt status of the current thread before re-throwing the InterruptedException. This ensures that the interruption is not ignored and can be properly handled by the caller.

 I've added Thread.currentThread().interrupt(); to set the interrupt status of the current thread when the InterruptedException is caught. Additionally, I've thrown an IOException with a message indicating that the flush operation was interrupted. This ensures that the interruption is not ignored and can be properly handled by the caller.

This some changes to the findHomeDir() function in ClientBuilder class to make it easier to read and work with. The code has been split into smaller methods which are easier to understand and maintain. This is super important because when code gets too complex it can be really hard to test and fix when things go wrong. By making the findHomeDir() function less complicated, we can improve the overall quality of the code and make it more reliable in the long run.

And refactored code in KubeConfig class in the getCredentials() method to improve its readability, maintainability and reduce its complexity. Replaced nested if-statements with early returns and used guard clauses to simplify the code flow. Also extracted the nested credentialsViaExecCredential method for better readability and maintainability

### DIFF
--- a/util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
+++ b/util/src/main/java/io/kubernetes/client/util/ClientBuilder.java
@@ -175,26 +175,39 @@ public class ClientBuilder {
         return config;
       }
     }
-    if (System.getProperty("os.name").toLowerCase().startsWith("windows")) {
-      String homeDrive = System.getenv("HOMEDRIVE");
-      String homePath = System.getenv("HOMEPATH");
-      if (homeDrive != null
-          && homeDrive.length() > 0
-          && homePath != null
-          && homePath.length() > 0) {
-        File homeDir = new File(new File(homeDrive), homePath);
-        if (homeDir.exists()) {
-          return homeDir;
-        }
-      }
-      String userProfile = System.getenv("USERPROFILE");
-      if (userProfile != null && userProfile.length() > 0) {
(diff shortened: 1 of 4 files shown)


brendandburns:
It looks to me like the unit tests locked up (they all timed out @ 360 minutes) so I think that there is something broken by this refactor.

lobocode:
> It looks to me like the unit tests locked up (they all timed out @ 360 minutes) so I think that there is something broken by this refactor.

You're right, I'll take a closer look here and test it on my fork to avoid these issues.

PR #2641, opened 2023-04-21
Fix resource lifecycle issues in KubectlPortForward

dsyer:
Without this change only one instance of PortForward is ever created
and so when its resources and closed they cannot be re-used. It is
better to create a new instance for each local socket, and then
second and subsequent client connections will succeed instead of
fail.


extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlPortForward.java
```diff
@@ -78,47 +78,44 @@ public void shutdown() {
...
+                InputStream in = result.getInputStream(targetPort);
+                OutputStream out = result.getOutboundStream(targetPort);
+                Thread t1 = copyAsync(sock.getInputStream(), out, onUnhandledError);
+                Thread t2 = copyAsync(in, sock.getOutputStream(), onUnhandledError);
+
+                t1.join();
+                in.close();
+                t2.join();
+              } catch (Exception ex) {
+                onUnhandledError.accept(ex);
```
brendandburns, line 114: `onUnhandledError` can be null (I think) so can you retain the `Optional` code from the original?

dsyer: It can only be null if the user intentionally sets it to null, right? Why would I want to do that?

dsyer: ... but whatever.

brendandburns:
/lgtm
/approve

Thanks!

PR #2642, opened 2023-04-23
migrate the javax.annotation.* to jakarta.annotation

tmrpavan:


yue9944882:
@tmrpavan can you elaborate what is the purpose of this? `javax.*` packages is deemed to be more universal than`jakarta.*`, isnt it?

brendandburns:
@yue9944882 wdyt? I'm personally ok to merge this, but it has implications for all users of this library (a new dependency on jakarta)

yue9944882:
@brendandburns sounds good. but the model generation is supposed to be done by the workflow. if we want this, we should start by merging https://github.com/kubernetes-client/gen/pull/241 then trigger a new run of the workflow.

brendandburns:
@yue9944882 agreed, I just wanted to make sure that folks were onboard with this before we merged the gen PR. I will get that process started.

dermot-hardy:
@brendandburns Are there any plans to merge this?  I notice that https://github.com/kubernetes-client/gen/pull/241 was merged back in May but that the `io.kubernetes:client-java:19.0.0` released a couple of months ago is still using `javax.annotation.*` rather than `jakarta.annotation.*`?

brendandburns:
Closing this because we now have a different approach to handle this (see #3027)

Issue #2638, opened 2023-04-20
KubectlLabel does not delete labels


patrick-schlipf:
**Describe the bug**
In `KubectlLabel`-class, when issuing `deleteLabel(String key)`, the label is not deleted but the value is set to the string `"null"`. This leads to the following issues:
- If the label exists, its value is overwritten with the string `"null"`.
- If the label does not exist, this call creates a new label with the string `"null"` as value.

**Client Version**
`18.0.0`

**Kubernetes Version**
`1.24.1`

**Java Version**
`openjdk version "17.0.6" 2023-01-17`
`IBM Semeru Runtime Open Edition 17.0.6.0 (build 17.0.6+10)`

**To Reproduce**
Case: Delete existing label:
  
1. Add label.
```
➜  ~ kubectl label node node1 test/delete-label=OK
node/node1 labeled
```

2.  Check that the label exist.
```
➜  ~ kubectl get node -l=test/delete-label -L=test/delete-label
NAME       STATUS   ROLES    AGE    VERSION   DELETE-LABEL
node1      Ready    <none>   327d   v1.24.1   OK
```

3. Delete label:
```
Kubectl.label(V1Node.class)
        .name("node1")
        .deleteLabel("test/delete-label")
        .execute();
```

4. Check for the label again:
```
➜  ~ kubectl get node -l=test/delete-label -L=test/delete-label
NAME       STATUS   ROLES    AGE    VERSION   DELETE-LABEL
node1      Ready    <none>   327d   v1.24.1   null
```

Case: Delete non-existing label:
  
1.  Check that the label does not exist.
```
➜  ~ kubectl get nodes -l=test/delete-label -L=test/delete-label
No resources found
```

2. Delete label:
```
Kubectl.label(V1Node.class)
        .name("node1")
        .deleteLabel("test/delete-label")
        .execute();
```

3. Check for the label again:
```
➜  ~ kubectl get nodes -l=test/delete-label -L=test/delete-label
NAME       STATUS   ROLES    AGE    VERSION   DELETE-LABEL
node1      Ready    <none>   327d   v1.24.1   null
```

**Expected behavior**
Label is deleted.

**KubeConfig**
N/A

**Server (please complete the following information):**
N/A

**Additional context**
N/A

AsfarHorani:
hi @patrick-schlipf is this issue still open? I am willing to contribute

patrick-schlipf:
[...]

PR #2645, opened 2023-04-24
Fixed bug: label being set as null instead of deleted

v1ku:
Updated KubectlLabel class to fix bug #2638 where the label was not deleted properly. The deleteLabel() method now removes the label key from the addingLabels map, and the resulting JSON object in the PUT request correctly omits the label key instead of setting it to null. Also updated the corresponding unit test to reflect the changes.


yue9944882 (commented):
/lgtm
/approve

PR #2662, opened 2023-05-16
Fixed field names for explicit substitution

telyatnikov:
For substitution we have to use field names, for example `xKubernetesEmbeddedResource` instead of `x-kubernetes-embedded-resource`.

### DIFF
--- a/util/src/main/java/io/kubernetes/client/util/Yaml.java
+++ b/util/src/main/java/io/kubernetes/client/util/Yaml.java
@@ -53,6 +53,8 @@ import org.yaml.snakeyaml.nodes.Tag;
 import org.yaml.snakeyaml.representer.Represent;
 import org.yaml.snakeyaml.representer.Representer;
 
+import javax.lang.model.SourceVersion;
+
 /** The type Yaml. */
 public class Yaml {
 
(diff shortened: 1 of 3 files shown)


brendandburns:
Thank you for the PR! Can you add a unit test that demonstrates the broken behavior and validates that this PR fixes it (and so that it stays fixed in the future)

telyatnikov:
I've included a unit test that demonstrates the absence of the added property in TypeDescription.

brendandburns:
Cool thank you! fwiw, I didn't mean that you had to revert your fix and prove to me that it would fail, but I appreciate that you did!

You can add the fix back in and we can get this merged.

Thank you!

telyatnikov:
I had to figure out why the tests failed after my 'fix' and how PropertySubstitute from snakeyaml actually works. Well, my fix proved to be incorrect. 

But maybe my attempt will be useful in some way - I've found that one of the extensions - `x-kubernetes-validations` wasn't included. I added the relevant unit test and fix. I aimed to make the search for required substitutions more flexible and get rid of hardcoded annotations. 

I'm really sorry for wasting your time and jumping into things without testing or thinking it through. Thanks for your understanding!

PR #2682, opened 2023-05-26
Hitansu

hitansupanda:
I have tried adding semaphore in copytext.java


brendandburns:
This is a good test to fix, but this isn't the right way to do it.

What you want to do is to create a Semaphore with one permit `new Semaphore(1);` and then acquire it _before_ the thread is started. Then release once the thread Runnable has made the call, and also acquire _again_ after the thread has started.

And then delete the sleep.

That way you are guaranteed that the thread has finished executing before the validation code runs.

PR #2693, opened 2023-06-05
Shutdown resync and reflect executors now in Controller.shutdown

mindgamer:


util/src/main/java/io/kubernetes/client/informer/cache/Controller.java
```diff
@@ -158,8 +158,8 @@ public void stop() {
         reflectorFuture.cancel(true);
       }
     }
-    reflectExecutor.shutdown();
-    resyncExecutor.shutdown();
+    reflectExecutor.shutdownNow();
```
yue9944882, line 161: can you elaborate why we need to forcibly shutdown the executors here?

mindgamer: We have implemented the Shared Informer Index to cache Spark Job Objects across a vast number of K8s clusters and namespaces. The eviction of caches occurs when they remain unused for over an hour. Upon cache eviction, we initiate the shutdown of controllers with the expectation that both the reSync and reflectExecutor will be properly terminated. However, in a high throughput environment characterized by hundreds of clusters, thousands of namespaces, and millions of jobs, frequent cache creation and evictions prolong the shutdown process for the reSync and reflectExecutor, resulting in thread leaks.

To mitigate this issue, we have updated the controller shutdown code. Instead of utilizing the "shutdown" method, we now employ "shutdownNow" to promptly terminate the reSync and reflectExecutor, ensuring a more efficient shutdown right away.

mindgamer: @yue9944882 Hi Min Kim,

I would greatly appreciate it if you could kindly review the comment above and provide feedback. If everything appears to be in order, could you please sign-off on the PR?

Based on the test results, we have observed that the resync and reflectExecutor threads remain idle even after the cache has been evicted.

yue9944882 (commented):
/lgtm
/approve

yue9944882:
@mindgamer do you want to backport this to any past release?

mindgamer:
> @mindgamer do you want to backport this to any past release?

Yes, Would it be possible to backport to 18.0.0?

PR #2695, opened 2023-06-06
Update generate.yml

brendandburns:
This fixes spotless formatting so that we can regenerate the client for 1.27.x


yue9944882 (commented):
/lgtm
/approve

PR #2712, opened 2023-06-20
cherry pick of #2693: Shutdown resync and reflect executors now in

mindgamer:
Cherry pick of #2693 on version 18.0.0

#2693: Shutdown resync and reflect executors now in


Huang-Wei:
cc @yue9944882 for approval 🙏

PR #2716, opened 2023-06-22
Update codeql-analysis.yml

brendandburns:
Lock to a specific semantic version b/c 2.13.4 appears to break us.


brendandburns:
force LGTM b/c our build is broken.

PR #2760, opened 2023-08-15
( WIP ) Issue 2606 - java client sdk with jakarta annotation

dasarathirout:
( WIP ) Issue 2606 - java client sdk , jakarta namespaces built with open api tool v6.6.0


dasarathirout:
Generated client sdk with fix  https://github.com/kubernetes-client/gen/issues/246 open-api-tool v6.6.0

brendandburns:
Thanks for taking this on!

I noticed that this added a number of CI/CD files like a github actions and a travis CI build. I don't think we want to add those files.

I'll review the gen change first and see if we can get that to a good state, and then we can go from there

PR #2768, opened 2023-08-21
Update dependencies except for Spring

rjeberhard:
We're seeing OWASP scan reports about okio, which is included by okhttp3. I'm curious about why Dependabot didn't update this dependency? Whatever the answer, here are all of the missing updates except for those from Spring, which seem to require other, compensating changes.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -233,7 +233,7 @@
       <dependency>
         <groupId>ch.qos.logback</groupId>
         <artifactId>logback-classic</artifactId>
-        <version>1.4.11</version>
+        <version>1.3.11</version>
         <scope>test</scope>
       </dependency>
       <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -41,14 +41,14 @@
     <javax.annotation.version>1.3.2</javax.annotation.version>
     <snakeyaml.version>2.1</snakeyaml.version>
     <slf4j.version>2.0.7</slf4j.version>
-    <caffeine.version>2.9.3</caffeine.version>
-    <protobuf.version>3.24.0</protobuf.version>
+    <caffeine.version>3.1.8</caffeine.version>
+    <protobuf.version>3.24.1</protobuf.version>
     <junit.version>4.13</junit.version>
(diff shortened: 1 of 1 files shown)


brendandburns:
Thanks for the PR. I'm wondering if dependabot doesn't do this b/c there are explicit versions in okhttp or somesuch?

Anyway, assuming tests pass:

/lgtm
/approve

brendandburns:
fwiw, dependabot did send one of them:

https://github.com/kubernetes-client/java/pull/2767

rjeberhard:
> fwiw, dependabot did send one of them:
> 
> #2767

Yes, it's good, of course, that Dependabot found this update. That does leave a mystery of why it never offered PR's for the other updates. I looked at the configuration and it's the simple, typical configuration found in many other projects. Because I could see that the okhttp update had not occurred, I ran `mvn versions:display-dependency-updates` and reviewed the output.

brendandburns:
Yeah, it's odd it looks like dependabot has never offered an update for `okhttp` all of the PRs that I can find that reference okhttp are manual:

https://github.com/search?q=repo%3Akubernetes-client%2Fjava+okhttp&type=pullrequests

brendandburns:
@rjeberhard looks like the logback dependency doesn't support Java8?

rjeberhard:
> @rjeberhard looks like the logback dependency doesn't support Java8?

Yeah, I was too hopeful that I'd be able to do all of the updates. I think that the best plan would be to have specific PR's for each dependency similar to what Dependabot should have created.

brendandburns:
Ok, individual targetted PRs makes sense.

I also wonder if we need to move java8 support out into a different fork or some other approach to continuing support for Java8...

brendandburns:
I looked into this more I think logback should move to 1.3.11 not 1.4.11 they claim to be identical except 1.3 supports Java 8...

rjeberhard:
Since this PR is down to just 2 changes, this seems close enough to the goal of having individual PR's. In the future, I'll create individual PR's if Dependabot doesn't generate the necessary updates.

PR #2785, opened 2023-09-08
Update PortForwardTest.java

peeplika:
fixed the test to use semaphore class instead of thread.sleep() method.

### DIFF
--- a/util/src/test/java/io/kubernetes/client/PortForwardTest.java
+++ b/util/src/test/java/io/kubernetes/client/PortForwardTest.java
@@ -184,7 +184,6 @@ public class PortForwardTest {
             });
     synchronized (block) {
       t.start();
-      Thread.sleep(2000);
       handler.close();
       block.wait();
     }


brendandburns:
I don't think that this is the correct fix. You need to move the Semaphore create outside of the Thread, and then block on the semaphore instead of sleeping.

peeplika:
@brendandburns I have tried again , can you  review it again ? I am sorry for the inconvenience , I have just started with java but I wanted to try .

util/src/test/java/io/kubernetes/client/PortForwardTest.java
```diff
@@ -110,6 +111,12 @@ public void testPortForwardResult() throws IOException, InterruptedException {
         makeStream(new byte[] {2, 32, 3}, msgData2.getBytes(StandardCharsets.UTF_8)));
 
     final Object block = new Object();
+    final Semaphore s= new Semaphore(1);
+    try{
```
brendandburns, line 115: I don't think you need the try/catch block. This function can throw InterruptedException already (though it won't in practice)

util/src/test/java/io/kubernetes/client/PortForwardTest.java
```diff
@@ -121,13 +128,13 @@ public void run() {
...
                   ex.printStackTrace();
-                }
-              }
+                }}
             });
     synchronized (block) {
       t.start();
       block.wait();
     }
+    s.release();
```
brendandburns, line 137: the release should happen inside the Thread. The idea is that you reserve the Semaphore before the thread starts, and then release it inside the thread. That way you know for sure that the Thread has executed before moving on to the test.

That can replace the sleep which was trying to do the same thing.

brendandburns:
I apologize, I hadn't looked at the code closely. The necessary synchronization is already there with the `Object block`

I think that the only thing that is necessary is removing the `sleep` in the second test. I think it got left in there by accident.

peeplika:
@brendandburns Thanks for explaining .  Would you like me to remove it, or will you do it yourself ?

brendandburns:
@peeplika feel free to modify this PR to make that change.

peeplika:
@brendandburns I'm new to this and still learning about this , I couldn't quite figure out what went wrong with the PR,  am 
  I supposed to perform some sort of checks on the code before making the PR.  can you guide me on what should I do now?

brendandburns:
@peeplika sorry for the late response. You need to run `mvn test` locally and get all of the tests passing, then they should pass CI/CD also.

brendandburns:
Closing this for now, feel free to use `/reopen` if you update it later.

Issue #2786, opened 2023-09-08
Example code is out of date.


Ananya2001-an:
**Describe the bug**
I tried running the example for `listPodForAllNamespaces` in my local maven project but it's giving me compilation error. The null values being passed as arguments to the method seems to be the problem. I am not sure if the example provided is wrong or something else seems to the issue here? The same thing happened with other methods as well...

```java
package com.example;

import io.kubernetes.client.openapi.ApiClient;
import io.kubernetes.client.openapi.ApiException;
import io.kubernetes.client.openapi.Configuration;
import io.kubernetes.client.openapi.apis.CoreV1Api;
import io.kubernetes.client.openapi.models.V1Pod;
import io.kubernetes.client.openapi.models.V1PodList;
import io.kubernetes.client.util.Config;
import java.io.IOException;

/**
 * A simple example of how to use the Java API
 *
 * <p>Easiest way to run this: mvn exec:java
 * -Dexec.mainClass="io.kubernetes.client.examples.Example"
 *
 * <p>From inside $REPO_DIR/examples
 */
public class KubernetesClientDemo {
  public static void main(String[] args) throws IOException, ApiException {
    ApiClient client = Config.defaultClient();
    Configuration.setDefaultApiClient(client);

    CoreV1Api api = new CoreV1Api();
    V1PodList list =
        api.listPodForAllNamespaces(null, null, null, null, null, null, null, null, null, null, null);
    for (V1Pod item : list.getItems()) {
      System.out.println(item.getMetadata().getName());
    }
  }
}
```

![Screenshot 2023-09-08 175233](https://github.com/kubernetes-client/java/assets/55504616/a4dea10c-5375-45e1-b234-7df1dc40df56)

**Client Version**
e.g. `18.0.1`

**Kubernetes Version**
minikube

**Java Version**
e.g. Java 11

**Expected behavior**
Should have listed all the pods.

**Server (please complete the following information):**
 - OS: [e.g. WSL]

brendandburns:
Please re-read the error message. You are passing the wrong number of parameters. Passing nulls (at least for nullable types) is fine.

Closing this issue as it is not a problem with this client.

brendandburns:
I realized that we haven't brought the examples forward for recent releases (e.g. 18.0.1) so the older examples are not correct. Reopening this issue to cover updating the examples.

rohan-kapse:
is it still active?

PR #2788, opened 2023-09-09
chore: update examples-release-18

Ananya2001-an:
Related to #2786 
Corrected parameter count for few examples for release 18.0.1
Also added 2 more examples:
- NamespaceList
- ResourceQuotaList

**Additional context:**
The client still remains open after executing a function; we can mitigate it by implementing the AutoCloseable interface and using a TWR block.

Also for some methods we are passing the `pretty` parameter as a String which I guess should be a Boolean IMO. Wdyt? We would have to change it everywhere tho...

### DIFF
--- a/examples/examples-release-18/src/main/java/io/kubernetes/client/examples/Example.java
+++ b/examples/examples-release-18/src/main/java/io/kubernetes/client/examples/Example.java
@@ -36,9 +36,11 @@ public class Example {
 
     CoreV1Api api = new CoreV1Api();
     V1PodList list =
-        api.listPodForAllNamespaces(null, null, null, null, null, null, null, null, null, null, null);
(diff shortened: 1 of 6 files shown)


Ananya2001-an:
the `listPodsForAllNamespaces` method in the latest 18.0.1 version is accepting only 10 parameters but in the CoreV1Api implementation of the method we have defined 11 params including `sendInitialEvents`. In the Javadoc as well we have mentioned only 10 params and it's producing results properly with 10 args...that's why the tests were failing in the last commit since the CoreV1Api implementation has 11 params defined...

yue9944882:
fixed example by https://github.com/kubernetes-client/java/pull/2800

/close

please feel free to reopen if i missed anything

PR #2791, opened 2023-09-14
Call exceptionHandler on failure to acquire lease

wpapsco-apex:
Clients may want to process failures to acquire leases so we should call the exception handler so they are notified of the exception and can process it.

### DIFF
--- a/extended/src/main/java/io/kubernetes/client/extended/leaderelection/LeaderElector.java
+++ b/extended/src/main/java/io/kubernetes/client/extended/leaderelection/LeaderElector.java
@@ -291,6 +291,7 @@ public class LeaderElector implements AutoCloseable {
       oldLeaderElectionRecord = lock.get();
     } catch (ApiException e) {
       if (e.getCode() != HttpURLConnection.HTTP_NOT_FOUND) {
+        exceptionHandler.accept(e);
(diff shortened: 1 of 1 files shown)


brendandburns:
The failure to acquire the lease is indicated by the fact that it returns `false` so I'm not sure why the exception handler needs to be invoked? Can you describe your scenario in more detail?

Thanks!

wpapsco-apex:
The idea is I'd like to know why it has failed to acquire the lease. For example a 403 would tell me the application needs permissions. I also thought a 404 might tell me there's an issue connecting to k8s but now I see that's the same error used when no lease has been created yet.

Basically I'm looking for a way to detect issues with the deployment and cause a larger failure so it's more noticeable that a leader is failing to be elected. Currently it's hard to differentiate between simply waiting to acquire a lock in a healthy environment vs failing to acquire a lock due to some error like permissions or connection issues.

I think I might instead use a `can-I` [equivalent](https://github.com/fabric8io/kubernetes-client/blob/main/kubernetes-examples/src/main/java/io/fabric8/kubernetes/examples/kubectl/equivalents/CanIEquivalent.java) and if that fails, cause a larger failure. That covers both scenarios I'm mainly concerned about. If you have other suggestions that would be great but otherwise I'll just close the PR as I don't think this covers my use case anymore.

PR #2796, opened 2023-09-19
Update pom.xml to skip release (I hope)

brendandburns:
This should skip staging these jars to nexus

https://help.sonatype.com/repomanager3/integrations/nexus-repository-maven-plugin#NexusRepositoryMavenPlugin-Configuration

And hopefully fix the release break here:
https://github.com/kubernetes-client/java/actions/runs/6226166183/job/16898118043


examples/examples-release-16/pom.xml
```diff
@@ -93,6 +93,14 @@
 
 	<build>
 		<plugins>
+		    <plugin>
```
yue9944882, line 96: nit: indent

yue9944882 (commented):
we can verify this by looking into the snapshot release log. if this works, these modules should also be skipped in snapshots

examples/examples-release-16/pom.xml
brendandburns: done.

brendandburns:
@yue9944882 addressed the formatting (we should perhaps add a pre-commit check for xml formatting)

ptal.

yue9944882 (commented):
/lgtm
/approve

PR #2700, opened 2023-06-11
Handler for notifying that Reconcile for Object that existed before DefaultController is started is completed

deftfitf:
https://github.com/kubernetes-client/java/issues/2699

Issue #2699, opened 2023-06-11
[Suggestion] Handler for notifying that Reconcile for Object that existed before DefaultController is started is completed


deftfitf:
Hi. I'm looking into implementing a custom controller using kubernetes-client/java.

When implementing using DefaultController/Reconciler, one problem occurred that we had to consider.
A case where you want to enable a redness probe when a custom controller starts and successfully reconcile all valid Kubernetes Objects at the time of startup.

By passing Informer's hasSync to readyFuncs, we can determine that Informer's cache is synchronized, but it seems difficult to determine specifically whether it was processed by the reconciler.

So I thought I could achieve this use case by slightly modifying the existing implementation, so I'd like to submit a patch.

PR #2803, opened 2023-09-27
Handler for notifying that Reconcile for Object that existed before DefaultController is started is completed

deftfitf:
https://github.com/kubernetes-client/java/issues/2699
reopen: https://github.com/kubernetes-client/java/pull/2700


deftfitf:
Sorry for being late to address that build failure about `Java 8.0.x Maven Test on ubuntu-latest`.

There seems to be weird behavior that Mockito has, so I changed some test code to be aligned with other test code in the same file.

I've also checked if it works well on Java 8 (AdoptOpenJDK 8) as well as Java 17.

Could you check this again ?

PR #2811, opened 2023-10-03
Remove download logs from CI/CI

brendandburns:
Complete removing download logs to reduce logspam during CI/CD error investigation.


yue9944882 (commented):
/lgtm
/approve

PR #2815, opened 2023-10-04
Bump staging version to 20.0.0-SNAPSHOT

yue9944882:
Bumping root version to 20.0.0-SNAPSHOT as we cut a `release-19` branch

after this we can regenerate using 1.28 k8s openapi spec


brendandburns:
Sounds good. fwiw, there are reports (https://github.com/kubernetes-client/gen/issues/253) that generation is broken. At the very least we need to remove `JSON.java` from the .openapi-ignore file. I will send that PR.

/lgtm
/approve

PR #2818, opened 2023-10-05
Remove JSON.java from openapi ignore file.

brendandburns:
cc @yue9944882

### DIFF
--- a/kubernetes/.openapi-generator-ignore
+++ b/kubernetes/.openapi-generator-ignore
@@ -6,5 +6,3 @@ git_push.sh
 README.md
 pom.xml
 
-# Remove when changes in kubernetes-client/java#366,#240 make into upstream openapi-generator
-src/main/java/io/kubernetes/client/openapi/JSON.java


yue9944882:
telling from the comment, looks like we're stopping JSON generation due to the following two PRs:
- #240 might be a blocker as we have a test case asserting the html escape issue https://github.com/kubernetes-client/java/blob/master/kubernetes/src/test/java/io/kubernetes/client/openapi/JSONTest.java#L27
- #366 doesn't involve any functional changes, it was meant for better OSGi support

yue9944882:
since we just bumped root version to 20.0.0-SNAPSHOT, we can try a code generation based on this PR to see if this is backward compatible. @brendandburns is there any other driver we want to refresh this JSON file generation

yue9944882:
@brendandburns trying to pinpoint that incompatibility introduced. but it looks the last commit in the generator upstream was one year ago:

> https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/Java/libraries/okhttp-gson/JSON.mustache

so we're probably already using `6.x` generator for a while, can you help me understand which part broke the compatibilty? 🤔

brendandburns:
We've been previously locked on the `4.x` release of the project, the PR to change that only merged in August:

https://github.com/kubernetes-client/gen/pull/247/files

That PR was done in an effort to migrate to `jakarta.annotation` instead of `javax.annotation`

yue9944882:
@brendandburns gotcha, as a follow-up of this. i will cherry-pick the two PRs onto `6.x` generated source code

yue9944882 (commented):
/lgtm

PR #2826, opened 2023-10-10
Update pom.xml

brendandburns:
@yue9944882 the last attempt to release didn't upload anything.

There is apparently a bug here:
https://issues.sonatype.org/browse/NEXUS-19853

That if skipped modules are last in the list they don't upload.

Changing the order around to see if it fixes things...


yue9944882 (commented):
/lgtm
/approve

PR #2828, opened 2023-10-12
Release-19: Reset release 19 pom version to 19.0.0-SNAPSHOT

yue9944882:
So that we can publish 19.0.0 again from release-19 branch @brendandburns


brendandburns:
/lgtm
/approve

We may have to clean up some tags also?

PR #2845, opened 2023-10-20
Revert "Release-19: Reset release 19 pom version to 19.0.0-SNAPSHOT"

yue9944882:
Reverts kubernetes-client/java#2828

@brendandburns looks like 19.0.0 is successfully out, am i missing anything? if that worked, we should revert this PR so the version matches


brendandburns:
/lgtm
/approve

I think maybe I made a typo in the github action. We should consider modifying it to prevent such errors.

PR #2873, opened 2023-11-06
feat: add prometheus operator CR probe

morganxf:


brendandburns:
Also we need the CLA to be signed.

Issue #2886, opened 2023-11-13
Build against a newer well supported Java version


javanegmond:
**Describe the bug**
The kubernetes java client is built against Java 8. This leads to a plethora of issues when using newer JDKs in projects.

**Client Version**
Any

**Kubernetes Version**
Doesn't matter

**Java Version**
Java 8

**To Reproduce**
Any Build

**Expected behavior**
kubernetes-client is built against a new, well supported, version of Java

**KubeConfig**
Not applicable

**Server (please complete the following information):**
Not Applicable

**Additional context**
As just one example, Bytebuddy (used for Mockito in testing, among others) can no longer enhance bytecode for the kubernetes client in combination with OpenJDK 21 without adding a flag for experimental support.

In #2452 some sort of workaround is laid out to be able to use the kubernetes client with Spring 6. More importantly, it is mentioned that:

> because we need to maintain Java 8 compatability
And I strongly wonder why there is this need. I couldn't find it personally. Wouldn't it be more apt to move on to a new, well supported, version of Java and let projects that are stuck on Java 8 deal with it by forking this repo instead?

brendandburns:
We're committed to supporting Java8 for the time being at least. One alternative might be to create a `legacy` build which supports Java8, while moving the general library forward to a newer version of the JDK.

We would be happy to entertain proposals or PRs to achieve that.

However, given the number of projects out there that still target Java8, moving forward exclusively would harm a large number of existing users and asking them to fork is not a user-centric approach.

javanegmond:
> However, given the number of projects out there that still target Java8, moving forward exclusively would harm a large number of existing users and asking them to fork is not a user-centric approach.

Not moving forward is not a community centric approach instead. A choice should be made: support a small set of users sticking to old, unsupported versions of the language, or the bigger community? If anything, we should encourage users still on Java 8 to move on.

I'll see if I can find some time to see what needs to be done to move this project towards the current era. As for legacy support (a legacy build or an untouched branch maybe); that will not be on me as you'll understand.

javanegmond:
FWIW, running unit tests this projects master branch as is using JDK 21 runs into an absolute smashing list of errors like this one:

```
[ERROR]   ResyncRunnableTest.testSupplierReturnsNull » Mockito 
Mockito cannot mock this class: class io.kubernetes.client.informer.cache.DeltaFIFO.

If you're not sure why you're getting this error, please open an issue on GitHub.


Java               : 21
[...]

PR #2888, opened 2023-11-14
Build with JDK 21

javanegmond:
As outlined in #2886, here's a change that will build and pass all tests except e2e tests. I've not had time to run e2e tests, but they compile.

Note that this very eagerly changes all locations where a java version is referenced to 21. What level of support is desired for this library is up to the maintainers. I'll argue that at the very least it's gotta be newer than Java 8.

### DIFF
--- a/docs/java-controller-tutorial-rewrite-rs-controller.md
+++ b/docs/java-controller-tutorial-rewrite-rs-controller.md
@@ -32,8 +32,8 @@ plugins are present in the [pom.xml](https://github.com/yue9944882/replicaset-co
     <groupId>org.apache.maven.plugins</groupId>
     <artifactId>maven-compiler-plugin</artifactId>
     <configuration>
-        <source>8</source>
-        <target>8</target>
+        <source>21</source>
+        <target>21</target>
     </configuration>
 </plugin>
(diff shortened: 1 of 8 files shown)


brendandburns:
I think it would be ok to take this as a separate branch, but we're not going to move to Java 21 on the mainline for now.

Even if we move forward, we're going to move forward to Java 11 or Java 17 not Java 21.

If you want to merge this into the mainline, the right option to move forward is to update the pom.xml to build two different versions of the library, not to replace the current version.

@yue9944882 wdyt?

javanegmond:
> Even if we move forward, we're going to move forward to Java 11 or Java 17 not Java 21.

Sounds absolutely fair to me. Java 11 still has good support from multiple providers until some time in 2024 IIRC.

> If you want to merge this into the mainline, the right option to move forward is to update the pom.xml to build two different versions of the library, not to replace the current version.

All up to you guys. I just wanted to do a quick run of what effort is necessary to be able to build with Java 21, which turns out is not a lot, but it's not a freebie either.

brendandburns:
Closing this as we have moved to a different plan to move forward to JDK11

PR #2901, opened 2023-11-20
add overloaded methods

### LINKED ISSUE #2890: use overloading for required params
I feel that we should use method overloading to make it easier for users to call methods with just the required params...since passing all nulls is kind of a hustle... IMHO most of the times users are just passing these params as values to the methods and thus keeping in mind all the optional fields seems kind of less efficient... 

It sucks that we can't set default values for the params but at least this is better than what we are doing now... creating this option for the users would be nice :)  

can use **coreApi.createNamespacedPod(namespace, pod);** instead of this **coreApi.createNamespacedPod(namespace, pod, null, null, null, null);**

```java
createNamespacedPod(String namespace, V1Pod body){
     createNamespacedPod(namespace, body, null, null, null, null);
}

createNamespacedPod(namespace, body, {optional params}) { ... }
```

Ananya2001-an:
fixes #2890 

- created `overloading` package under `utils`
- added Pods class with overloaded methods; will do same for other common methods.... (namespaces, deployments...)


brendandburns:
This looks good. It needs formatting for style.

Please run `mvn spotless:apply` and that should take care of it.

Ananya2001-an:
will also add unit tests for the same in a separate PR

brendandburns:
This looks good. Given the simplicity of the code, I'm not sure that unit tests will add much, so don't feel obligated.

One note before we merge, I think that the package name should change. Perhaps something like `simplified` or `ergonomic`?

Thanks!

Ananya2001-an:
Yeah I was going to ask u about the name...I think simplified is good!

brendandburns:
/lgtm
/approve

Thanks for the PR!

PR #2906, opened 2023-11-23
Add Kubectl delete tests

awesominat:
this PR aims to add tests to Kubectl.delete and its ignoreNotFound method.

~I'm currently having issues getting it to work. I can't get Kubect.delete to recognize V1Job as a namespaced class if I set `KubectlDelete\<V1Job>`'s ApiClient. If I don't set the ApiClient, then executing delete fails to connect to local host.~
~I'm considering executing `ModelMapper.addModelMap` to add V1Job as namespaced class for Kubectl.delete to execute, but I'd like feedback on whether the changes I've made make sense or should be refactored.~


awesominat:
/cc @brendandburns 

Hi, whenever you're free, can you give me your feedback on my changes?

awesominat:
I have hard coded more http returns, but now it passes even when it should error for the second call. I am able to get it to error if I create a normal file that isn't a test and use Kubectl there. I feel like I am missing a crucial part of mocking these http calls

extended/src/test/java/io/kubernetes/client/extended/kubectl/KubectlDeleteTest.java
```diff
@@ -0,0 +1,185 @@
...
+
+    private static final String DISCOVERY_API =
+            new File(KubectlDeleteTest.class.getClassLoader().getResource("discovery-api.json").getPath())
+                    .toString();
+
+    private static final String DISCOVERY_APIV1 =
+            new File(
+                    KubectlDeleteTest.class
+                            .getClassLoader()
+                            .getResource("discovery-api-v1.json")
```
brendandburns, line 51: can you use `getResourceAsStream()` and load this into a byte array instead of the `File` stuff, I find it confusing to read.

brendandburns:
Thanks for adding tests!!

One small comment on simplifying the way resources are loaded, then LGTM.

brendandburns:
/lgtm
/approve

Thanks for the quick changes!

awesominat:
`readAllBytes()` does not exist in Java 8 (introduced in Java 9), I will look for an alternative.

awesominat:
verified that the code does compile on Java 8 and 17, should work now

PR #2928, opened 2023-12-04
Rename client-java-api to *-legacy

yue9944882:


brendandburns:
Sorry, I'm thinking about this some more. Do you think it would be easier to maintain this as a branch in git, rather than duplicating the code on `master`?

yue9944882:
> Sorry, I'm thinking about this some more. Do you think it would be easier to maintain this as a branch in git, rather than duplicating the code on master?

that's definitely viable but i'm slight worried if it's as convenient to maintain as a legacy module. additionally, putting it into a separate branch will sort of double the effort for code generation for every kubernetes release because we will need to open (or automate) 2 separate PRs upon every release. do you see any downside of having a legacy module?

yue9944882:
@brendandburns gotcha, let's move those to a separate branch then. i will file another PR after closing this!

brendandburns:
Sounds good. Thank you!

PR #2945, opened 2023-12-13
Branch(java8): Add `-legacy` version qualifier to java8 compatible modules

yue9944882:
@brendandburns alternatively, do you think we should make these legacy java8-compatible codes a new module (specifically a new module name) or the same module name with a suffixed version e.g. `19.0.0.legacy`?

### DIFF
--- a/client-java-contrib/admissionreview/pom.xml
+++ b/client-java-contrib/admissionreview/pom.xml
@@ -7,10 +7,10 @@
     <parent>
         <groupId>io.kubernetes</groupId>
         <artifactId>client-java-parent</artifactId>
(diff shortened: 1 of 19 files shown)


brendandburns:
Suffix feels fine to me for now at least, if people complain we can always change it in future releases.

yue9944882:
/hold cancel

@brendandburns update the PR to add a `-legacy` suffix to the existing versioning e,g:

- `20.0.0-legacy`
- `20.0.0-legacy-SNAPSHOT`

i suppose the new suffix conforms to the maven release version convention here: https://docs.oracle.com/middleware/1212/core/MAVEN/maven_version.htm

Issue #2241, opened 2022-05-05
Error creating deployment with deployment.yaml file - Unknown apiVersionKind apps/v1/Deployment is it registered?


cdkgenie:
I am creating a deployment using deployment.yaml file below with version 15.0.1 (Current Latest)

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: demoapp-sample
spec:
  selector:
    matchLabels:
      app: demoapp-sample
  replicas: 1
  template:
    metadata:
      labels:
        app: demoapp-sample
    spec:
      containers:
        - name: demoapp
          image: test/demoapp
          # Image that will be used to containers in the cluster
          imagePullPolicy: Always
          ports:
            - containerPort: 9005
              # The port that the container is running on in the cluster
```

My java code is, in which **resourceFile** object has the deployment.yaml content:

```
try {
            ApiClient client = ClientBuilder.cluster().build();
            Configuration.setDefaultApiClient(client);
            AppsV1Api api = new AppsV1Api();
            V1Deployment yamlDeployment = (V1Deployment) Yaml.load(new InputStreamReader(resourceFile.getInputStream()));
            V1Deployment createResult = api.createNamespacedDeployment("default", yamlDeployment, null, null, null, null);
            System.out.println(createResult);
            return "Deployment created successfully.";
        } catch (Exception e) {
            e.printStackTrace();
            return e.getMessage();
        }
```

My cluster-role.yaml file for the container running the above code is: 

```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  # "namespace" omitted since ClusterRoles are not namespaced
  name: demo-cluster-role
rules:
  - apiGroups: ["","apps"]
    resources: ["pods","deployments","services"]
    verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
```
So I have permissions to create the deployments. But when I run this code, Im getting exception as 

` Unknown apiVersionKind apps/v1/Deployment is it registered?`

How can I resolve this issue?

cdkgenie:
Added the following line to resolve the issue. Note: That method is deprecated.

Yaml.addModelMap("apps/v1","Deployment", V1Deployment.class);

cdkgenie:
I use spring boot, don't know what is causing that issue. but without that, it was not working.

yue9944882:
i suppose it's related to https://github.com/kubernetes-client/java/issues/1659

yue9944882:
[...]

Issue #2655, opened 2023-05-09
Unknown apiVersionKind secrets-store.csi.x-k8s.io/v1/SecretProviderClass is it registered?   - K8S java client 


payalgodhanioracle:
**Describe the bug**
Unknown apiVersionKind secrets-store.csi.x-k8s.io/v1/SecretProviderClass is it registered?   

Error Stack

Failed to load resources:
java.io.IOException: Unknown apiVersionKind secrets-store.csi.x-k8s.io/v1/SecretProviderClass is it registered?
	at io.kubernetes.client.util.Yaml.modelMapper(Yaml.java:554)
	at io.kubernetes.client.util.Yaml.loadAll(Yaml.java:179)
	at io.kubernetes.client.util.Yaml.loadAll(Yaml.java:145)

**Client Version**
e.g. `17.0.2 and 18.0.0`

**Kubernetes Version**
e.g. `1.23.4`

**Java Version**
e.g. Java 8

**To Reproduce**

**Expected behavior**
Should be able to create the resource - secrets-store.csi.x-k8s.io
This resource can be created with kubectl and helm manually but not with Java K8S client. We see that the fix is already in the master branch. This is the pull request which was merged but there was no java client release post this.

https://github.com/kubernetes-client/java/pull/2620

Could we create a new release from the master branch or have an ETA for the next release ?

brendandburns:
I don't think that this is the issue that you referenced.

This library only has the built-in resources by default. That resource is a custom resource (CRD) that is dynamically loaded into the cluster. You need to either register your own implementation of that resource using the `ModelMapper` class or you need to use the `generic` client as shown here:

https://github.com/kubernetes-client/java/blob/master/examples/examples-release-15/src/main/java/io/kubernetes/client/examples/GenericClientExample.java

There are also instructions here:
https://github.com/kubernetes-client/java/wiki/5.-Generate-Java-CRD-Model

If you want to generate the client models from an OpenAPI specification.

payalgodhanioracle:
The SecretProviderClass is registered with the cluster, and we still see the above error and we see that it works fine with kubectl

Issue #2377, opened 2022-09-16
Exception in YAML.load(File file): Unknown apiVersionKind batch/v1/CronJob is it registered?


Babak-abd:
Version: 16.0.0

I am creating a CronJob as below:

```
ApiClient client = Config.defaultClient();
Configuration.setDefaultApiClient(client);
InputStream cronJobYAML = Kub.class.getResourceAsStream("/cronJob.yaml");
String yaml = new BufferedReader(new InputStreamReader(cronJobYAML, StandardCharsets.UTF_8)).lines()
				.collect(Collectors.joining("\n"));
//TODO: Add ModelMapper.addModelMap() here
V1CronJob yamlCronJob = (V1CronJob) Yaml.load(yaml);  //<---------Exception Here
BatchV1Api batchAPI = new BatchV1Api();
V1CronJob createNamespacedCronJob = batchAPI.createNamespacedCronJob("default", yamlCronJob, "true", null, null, null);
```

but when I call YAML.load I get a null pointer exception followed by:
```
[err] java.io.IOException: Unknown apiVersionKind batch/v1/CronJob is it registered?
[err] 	at io.kubernetes.client.util.Yaml.modelMapper(Yaml.java:554)
[err] 	at io.kubernetes.client.util.Yaml.load(Yaml.java:93)
[err] 	at io.kubernetes.client.util.Yaml.load(Yaml.java:70)
```
The same code works for me when I create a maven project and run as jar.

I tried to debug more and I found that the issue is in https://github.com/kubernetes-client/java/blob/v16.0.0/util/src/main/java/io/kubernetes/client/util/ModelMapper.java#L461

`packageURL = "wsjar:file:/C:/Users/Babak/.m2/repository/io/kubernetes/client-java-api/16.0.0/client-java-api-16.0.0.jar!/io/kubernetes/client/openapi/models"`
So when `uri.getPath()` is called, it throws NullPointerException.

`ClassLoader Type :  com.ibm.ws.classloading.internal.AppClassLoader (I am using IBM websphere)`

For now I am adding `ModelMapper.addModelMap("batch", "v1", "CronJob", "CronJobs", V1CronJob.class, null);` but I am not sure if something else would go wrong later.
Also, I am not sure if the value I put for the 4th param (CronJobs) is correct.

brendandburns:
The client tries to statically load a bunch of these things via introspection, but different classloaders break this behavior sometimes.

[...]

PR #2952, opened 2023-12-17
#2241 fix bug: when running with a fat jar, need to read entries with openConnection

wkclz:
When running in IDE or a thin jar, it is ok.
When running with a fat jar in server, new JarFile can not open and read the entries.

### DIFF
--- a/util/src/main/java/io/kubernetes/client/util/ModelMapper.java
+++ b/util/src/main/java/io/kubernetes/client/util/ModelMapper.java
@@ -25,6 +25,7 @@ import java.io.IOException;
 import java.net.JarURLConnection;
 import java.net.URI;
 import java.net.URL;
+import java.net.URLDecoder;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Enumeration;
--- a/util/src/main/java/io/kubernetes/client/util/ModelMapper.java
+++ b/util/src/main/java/io/kubernetes/client/util/ModelMapper.java
@@ -484,7 +484,11 @@ public class ModelMapper {
   }
 
   private static void processJarPackage(URL packageURL, String packageName, String pkg, ArrayList<String> names) throws IOException {
-    logger.info("Loading classes from jar {}", packageURL.getFile());
+    String jarFileName = URLDecoder.decode(packageURL.getFile(), "UTF-8");
+    if (!jarFileName.startsWith("jar:") && !jarFileName.startsWith("nested:")) {
+      logger.error("Loading classes from jar with error packageURL: {}", jarFileName);
+    }
+    logger.info("Loading classes from jar {}", jarFileName);
     try (JarFile jf = ((JarURLConnection) packageURL.openConnection()).getJarFile()) {
       Enumeration<JarEntry> jarEntries = jf.entries();
       while (jarEntries.hasMoreElements()) {
--- a/util/src/main/java/io/kubernetes/client/util/ModelMapper.java
+++ b/util/src/main/java/io/kubernetes/client/util/ModelMapper.java
@@ -486,17 +486,26 @@ public class ModelMapper {
 
   private static void processJarPackage(URL packageURL, String packageName, String pkg, ArrayList<String> names) throws IOException {
     String jarFileName = URLDecoder.decode(packageURL.getFile(), "UTF-8");
-    if (!jarFileName.startsWith("jar:") && !jarFileName.startsWith("nested:")) {
+    JarFile jf = null;
+    // jar: client in repository; nested: client in a fat jar
+    if (jarFileName.startsWith("jar:") || jarFileName.startsWith("nested:")) {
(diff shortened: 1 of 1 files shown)


brendandburns:
Hrm, do you have a pointer to documentation about why this works this way? I definitely don't want this library to be opening URLs over the network, so is there a way to restrict this to file based entries in the fat jar?

brendandburns:
Also, you need to sign the CLA.

wkclz:
Running Yaml.load("yaml String"), in "io.kubernetes.client.util.Yaml:553"
When I run it in IDE, it can get Class by Kind, But in a fat jar, the class is null, and then, I got "Unknown apiVersionKind " + apiVersion + "/" + kind + " is it registered?"
I found several issues about this problem: issues: #1659, #2655, #2241, #2377 ...

Make a break point to trace the code, ModelMapper.initModelMap() is in the static, it will scan and cache classes to classesByGVK or preBuiltClassesByGVK from  "io.kubernetes.client.openapi.models".
The problem is in here: io.kubernetes.client.util.ModelMapper.processJarPackage:
When the project is a fat jar, the packageURL is like this: "nested:/home/apps.jar/!BOOT-INF/lib/client-java-api-19.0.0.jar!/io/kubernetes/client/openapi/models", so it would get an IOException.

If try to get the jar like this: ((JarURLConnection) packageURL.openConnection()).getJarFile(), it will be OK. I have try it by overwrite this class.
This URL is only the jar url in current project, not over the network, so it is safe.
If the URL is not a JarURL, it would an ClassCastException.

In this issue #2241, I output the stack of the problem.

CLA is ok now.

util/src/main/java/io/kubernetes/client/util/ModelMapper.java
```diff
@@ -484,10 +484,8 @@ private static List<String> getClassNamesFromPackage(ClassLoader classLoader, St
   }
 
   private static void processJarPackage(URL packageURL, String packageName, String pkg, ArrayList<String> names) throws IOException {
-    String jarFileName = URLDecoder.decode(packageURL.getFile(), "UTF-8");
-    jarFileName = jarFileName.substring(5, jarFileName.indexOf("!"));
-    logger.info("Loading classes from jar {}", jarFileName);
-    try (JarFile jf = new JarFile(jarFileName)) {
+    logger.info("Loading classes from jar {}", packageURL.getFile());
```
brendandburns, line 487: Can you add an explicit check for `packageURL.startsWith("jar:")` here?

brendandburns:
Ok, this approach works for me, I asked for a single explicit check for the `jar:` prefix otherwise looks good.

Is there a way that we can add a unit test for this?

wkclz:
Running unit test with `Yaml.load("yaml string");` can only test for a packageURL start with "jar:".
You may create a project with springboot, and package it with the plugin `spring-boot-maven-plugin`, and then, you will get a packageURL start with "nested:", e.g.: `nested:/home/apps.jar/!BOOT-INF/lib/client-java-api-19.0.0.jar!/io/kubernetes/client/openapi/models`

brendandburns:
/lgtm
/approve

I'm fine to merge this w/o a test given the complexity of creating one.

brendandburns:
Looks like this isn't working right in our e2e tests, it can't find the class for nodes any more.

wkclz:
fixed, and tested e2e tests

yue9944882:
Fixes: https://github.com/kubernetes-client/java/issues/1659

util/src/main/java/io/kubernetes/client/util/ModelMapper.java
```diff
@@ -485,14 +486,26 @@ private static List<String> getClassNamesFromPackage(ClassLoader classLoader, St
...
+    if (jarFileName.startsWith("jar:") || jarFileName.startsWith("nested:")) {
+      jf = ((JarURLConnection) packageURL.openConnection()).getJarFile();
+    }
+    // file: client is a file in target (unit test)
+    if (jarFileName.startsWith("file:") ) {
+      jarFileName = jarFileName.substring(5, jarFileName.indexOf("!"));
+      jf = new JarFile(jarFileName);
+    }
+    if (jf == null) {
+      logger.error("Loading classes from jar with error packageURL: {}", jarFileName);
```
yue9944882, line 500:
```suggestion
      logger.error("Loading classes from jar with error packageURL: {}", packageURL);
```
nit: looks like we might want to dump the packageURL here?

wkclz: Thx

wkclz: the same with packageURL

brendandburns:
@yue9944882 I'll wait for your lgtm on this one.

yue9944882 (commented):
/lgtm

Issue #2973, opened 2024-01-03
Request to upgrade okhttp


Babak-abd:
Could you please upgrade okhttp to 4.12.0?
It's currently on version 4.11.0 but that version has a security issue (CVE-2023-3656)

brendandburns:
We'd be happy to take a PR to update that dependency if you want to send one.

brendandburns:
Closed by #3014

PR #2988, opened 2024-01-10
Request to upgrade okhttp #2973

qwertyroop:
It's currently on version 4.11.0 and has been changed to 4.12.0

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -55,7 +55,7 @@
     <bouncycastle.version>1.77</bouncycastle.version>
     <gson.version>2.10.1</gson.version>
     <jsr305.version>3.0.2</jsr305.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -273,7 +273,7 @@
       <dependency>
         <groupId>org.mockito</groupId>
         <artifactId>mockito-core</artifactId>
-        <version>4.11.0</version>
+        <version>4.12.0</version>
```
brendandburns, line 276: mockito core 4.12.0 doesn't exist, I think that this might be an accidental search and replace that happened to hit a different library with the same version.

brendandburns:
@qwertyroop any chance you could update this PR?

brendandburns:
replaced by #3014

PR #2991, opened 2024-01-11
OpenApi generate from k8s 1.28 with `USE_SINGLE_PARAMETER` and client-side openapi validation enabled

yue9944882:
With this PR, the generated openapi client will land a non-backward-compatible change which refactors the api interface simpler and clearer, e.g.:

Previously the API interface was like:
```
V1PodList list =
        api.listPodForAllNamespaces(null, null, null, null, null, null, null, null, null, null);
```

After enabling `USE_SINGLE_PARAMETER`, it will be:

```
    V1PodList list = api.listPodForAllNamespaces()
            .fieldSelector(..) // optional builder flow
            .resourceVersion(..)
            .execute();
```


brendandburns:
This will be a pretty disruptive change. Is there a way to have both options be feasible for a time? Otherwise people adopting this change will require some major code changes.

Or maybe we can do this change in the `main` branch but not in the `-legacy` version? Is that the plan?

yue9944882:
@brendandburns right, for those who are not ready to onboard the new API interface, they can still use the legacy version module by adding `-legacy` suffix to their pom dependency:

https://github.com/kubernetes-client/java/blob/master-java8/pom.xml#L5

in the foreseeable future, we publish both from master branch and master-java8 branch together

brendandburns:
Sounds good.

/lgtm
/approve

PR #3021, opened 2024-01-31
Tracking manual edits in git diffs to unlock openapi ignore

yue9944882:
since we bumped the openapi generator to v6 from v4, there're a few additional manual code changes we need to made on top of the generated code. in the past, the only manual edits are in the `JSON` class however now it's extended to `V1ListMeta` and `V1Status` to be compatible with kubernetes graceful deletion (the returned type for k8s delete API is a status object however it will return the deleted object upon graceful/progressive deletion). this pull automates those manual changes by git patches instead of ignoring these models generation by the `.openapi-generator-ignore` configuration file. 

with this PR, the code generation workflow behavior will change from (1) ignoring manually editted source files to (2) always re-generate but apply git patches after generation.

an example of the code-generation PR from the new proposed workflow here can seen at: https://github.com/kubernetes-client/java/pull/3020

### DIFF
--- a/.github/workflows/generate.yml
+++ b/.github/workflows/generate.yml
@@ -62,6 +62,16 @@ jobs:
           USE_SINGLE_PARAMETER=true bash java.sh ../../kubernetes/ settings
           popd
           rm -rf gen
+          git config user.email "k8s.ci.robot@gmail.com"
+          git config user.name "Kubernetes Prow Robot"
+          git checkout -b "$BRANCH"
+          git add .
+          git commit -s -m 'Automated openapi generation from ${{ github.event.inputs.kubernetesBranch }}'
+      - name: Apply Manual Diffs
+        run: |
+          ls scripts/patches/*.diff | xargs git apply
+          git add .
+          git commit -s -m 'Applied patches under scripts/patches/*.diff'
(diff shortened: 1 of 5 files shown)


yue9944882:
@brendandburns to make sure the code generation workflow works correctly after the recent major refactoring, what do you think if we re-run code-generation for both main and legacy branch again before we publish new releases?

brendandburns:
/lgtm
/approve

Did you see: https://github.com/kubernetes-client/java/issues/431

Perhaps there is a patch to fix that we can add?

yue9944882:
>https://github.com/kubernetes-client/gen/pull/263

@brendandburns  we dont need to apply any patch to get that fixed, just updating the codegen configuration will do. it is accidentally rolled back due to lacking unit tests. i will open another PR to add testings

yue9944882:
> [ERROR]   V1SecretTest.testV1SecretEquality:22 » InaccessibleObject Unable to make field private static final long java.util.HashMap.serialVersionUID accessible: module java.base does not "opens java.util" to unnamed module @7549926c

actually, the fix above won't work anymore because the reflection-based equal function is not compatible with java17 with above error when i tested locally. we still need a hard patch for this

PR #3081, opened 2024-02-15
Add null guards for Exec.ExecutionBuilder

pingpingy1:
The class `Exec.ExecutionBuilder` has three `private final` properties: `namespace`, `name`, and `command`. While it makes little sense to provide `null` as their values, if this happens, this would throw a `NullPointerException` at a much later time, complicating the debugging process.
This commit implements a fail-fast `null` guards for these variables.

Example:
```java
import io.kubernetes.client.Exec;

public class Test {
    public static void main(String args[]) throws Exception {
        Exec exec = new Exec(null);
        Exec.ExecutionBuilder builder = exec.newExecutionBuilder("", "", null);
        builder.execute();
    }
}
```

```console
Exception in thread "main" java.lang.NullPointerException: Cannot invoke "io.kubernetes.client.openapi.ApiClient.escapeString(String)" because "this.localVarApiClient" is null
        at io.kubernetes.client.openapi.apis.CoreV1Api.readNamespacedPodCall(CoreV1Api.java:26682)
        at io.kubernetes.client.openapi.apis.CoreV1Api.readNamespacedPodValidateBeforeCall(CoreV1Api.java:26726)
        at io.kubernetes.client.openapi.apis.CoreV1Api.readNamespacedPodWithHttpInfo(CoreV1Api.java:26767)
        at io.kubernetes.client.openapi.apis.CoreV1Api.readNamespacedPod(CoreV1Api.java:26747)
        at io.kubernetes.client.Exec$ExecutionBuilder.execute(Exec.java:466)
        at Test.main(Test.java:7)
```

### DIFF
--- a/util/src/main/java/io/kubernetes/client/Exec.java
+++ b/util/src/main/java/io/kubernetes/client/Exec.java
@@ -39,6 +39,7 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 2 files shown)


brendandburns:
/lgtm
/approve

Thanks for the PR, and especially thanks for adding unit tests!

Issue #3082, opened 2024-02-15
Support for lower case suffixes (g, k) in suffix and quantity formatter


mmpataki:
**Describe the bug**
We tried to pass the memory required as 16g and it's not supported. We get below exception 
```java
io.kubernetes.client.custom.QuantityFormatException: Unable to parse numeric part of quantity: 16g
        at io.kubernetes.client.custom.QuantityFormatter.parseNumericValue(QuantityFormatter.java:42)
        at io.kubernetes.client.custom.QuantityFormatter.parse(QuantityFormatter.java:28)
        at io.kubernetes.client.custom.Quantity.fromString(Quantity.java:66)
        at io.kubernetes.client.custom.Quantity.<init>(Quantity.java:52)
```

**Client Version**
e.g. `20.0.0`

**Kubernetes Version**
NA

**Java Version**
NA

**To Reproduce**
Use the below API
```java
Quantity q = new Quantity("16g");
```

**Expected behavior**
The behavior similar to `java` is expected. It supports 16g, 16G, 16k, 16K (M and m cannot be supported because 'm' means milli here)

Here is how `java` implements the `-Xmx` parsing 
https://github.com/openjdk/jdk11/blob/37115c8ea4aff13a8148ee2b8832b20888a5d880/src/hotspot/share/runtime/arguments.cpp#L814

**KubeConfig**
NA

**Server (please complete the following information):**
NA

**Additional context**
NA

brendandburns:
Since Kubernetes is a cross-language API, it doesn't necessarily follow the Java standards.

The memory units for Kubernetes are specified here:
https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-memory

Closing this issue since it is working as intended.

PR #3083, opened 2024-02-15
Support for lower case suffixes (g, k) in suffix and quantity formatter

mmpataki:
Support for lower case suffixes (g, k) in suffix and quantity formatter #3082


brendandburns:
Thank you for sending the PR, but as I mentioned in the corresponding issue, the units for Kubernetes are specified and do not match the Java units:

https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-memory

Closing this PR.

Issue #2593, opened 2023-03-13
java.lang.IllegalStateException: Not a string


a364176773:
**Describe the bug**
My pod doesn't set livenessProbe, why does k8sclient have to read it instead of determining if it's null or not?
```
com.fasterxml.jackson.databind.JsonMappingException: Not a string (through reference chain: io.kubernetes.client.util.Watch$Response["object"]->io.kubernetes.client.openapi.models.V1Pod["spec"]->io.kubernetes.client.openapi.models.V1PodSpec["containers"]->java.util.ArrayList[0]->io.kubernetes.client.openapi.models.V1Container["livenessProbe"]->io.kubernetes.client.openapi.models.V1Probe["httpGet"]->io.kubernetes.client.openapi.models.V1HTTPGetAction["port"]->io.kubernetes.client.custom.IntOrString["strValue"])
at com.fasterxml.jackson.databind.JsonMappingException.wrapWithPath(JsonMappingException.java:397)
at com.fasterxml.jackson.databind.JsonMappingException.wrapWithPath(JsonMappingException.java:356)
at com.fasterxml.jackson.databind.ser.std.StdSerializer.wrapAndThrow(StdSerializer.java:316)
at com.fasterxml.jackson.databind.ser.std.BeanSerializerBase.serializeFields(BeanSerializerBase.java:763)
at com.fasterxml.jackson.databind.ser.BeanSerializer.serialize(BeanSerializer.java:178)
at com.fasterxml.jackson.databind.ser.BeanPropertyWriter.serializeAsField(BeanPropertyWriter.java:728)
at com.fasterxml.jackson.databind.ser.std.BeanSerializerBase.serializeFields(BeanSerializerBase.java:755)
at com.fasterxml.jackson.databind.ser.BeanSerializer.serialize(BeanSerializer.java:178)
at com.fasterxml.jackson.databind.ser.BeanPropertyWriter.serializeAsField(BeanPropertyWriter.java:728)
at com.fasterxml.jackson.databind.ser.std.BeanSerializerBase.serializeFields(BeanSerializerBase.java:755)
at com.fasterxml.jackson.databind.ser.BeanSerializer.serialize(BeanSerializer.java:178)
at com.fasterxml.jackson.databind.ser.BeanPropertyWriter.serializeAsField(BeanPropertyWriter.java:728)
at com.fasterxml.jackson.databind.ser.std.BeanSerializerBase.serializeFields(BeanSerializerBase.java:755)
[...]

Issue #2566, opened 2023-02-16
IllegalStateException: Not an integer


youngledo:
**Describe the bug**
I have a **V1Deployment** and try to parse it using jackson-dataformats-text(https://github.com/FasterXML/jackson-dataformats-text/tree/master/yaml), but an exception occurs.

Thanks.

**Client Version**
16.0.0

**Java Version**
Java 11

**To Reproduce**

1. A V1Deployment：
<img width="602" alt="企业微信截图_978f61e2-03e3-48cd-9d29-445a16b52c4c" src="https://user-images.githubusercontent.com/68590897/219313801-51a1c292-8f01-4ea0-8ad2-7c6114689bc0.png">

2. Use this method：com.fasterxml.jackson.databind.ObjectMapper#writeValueAsString
<img width="963" alt="企业微信截图_58686c34-ce79-4513-9593-6d79acd3ed66" src="https://user-images.githubusercontent.com/68590897/219313822-547d6e7e-30d4-46a7-ad76-8ae26c1fd5ef.png">

brendandburns:
IntOrString is a custom type that is either an int or a string. To parse it correctly you need to set up a custom encoder/decoder like we do here:

https://github.com/kubernetes-client/java/blob/master/util/src/main/java/io/kubernetes/client/util/Yaml.java#L249

I don't know the exact details of how to do that for the FasterJackson library. That's beyond the scope of this client, so I'm closing this issue as I believe the question is answered. Please use `/reopen` if you need further assistance.

youngledo:
> Can it return null instead of throw an exception "Not a string" or "Not an integer" ? It's ok when using Gson, but not all of the project want to dependency Gson when jackson is the default serialize tool.
> 
> ```
>   public String getStrValue() {
>     return isInt ? null : strValue;
>   }
> 
>   public Integer getIntValue() {
>     return isInt ? intValue : null;
>   }
> ```
It seems that Jackson does not support this kind of customization, need to provide a PR to Jackson

wkclz:
我在 IntOrString 上指定序列化逻辑了，在 #3113 上。Jackson 不会帮干这种事的，标准的做法就是 getter setter.

PR #3113, opened 2024-02-24
Add JacksonSerializer to serialize IntOrString

wkclz:
It's OK when using Gson, but not all of the project want to dependency Gson when jackson is the default serialize tool. If return null instead of an exception, it can work well with jackson or other serialize tool.

### DIFF
--- a/kubernetes/src/main/java/io/kubernetes/client/custom/IntOrStringJacksonSerializer.java
+++ b/kubernetes/src/main/java/io/kubernetes/client/custom/IntOrStringJacksonSerializer.java
@@ -1,0 +1,39 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package io.kubernetes.client.custom;
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.ser.std.StdSerializer;
+
+import java.io.IOException;
+
+public class IntOrStringJacksonSerializer extends StdSerializer<IntOrString> {
+
+  public IntOrStringJacksonSerializer() {
+    this(null);
+  }
+
+  public IntOrStringJacksonSerializer(Class<IntOrString> t) {
+    super(IntOrString.class);
+  }
+
+  @Override
+  public void serialize(IntOrString value, JsonGenerator jg, SerializerProvider provider) throws IOException {
+    if (value.isInteger()) {
+      jg.writeNumber(value.getIntValue());
+    } else {
+      jg.writeString(value.getStrValue());
+    }
+  }
+}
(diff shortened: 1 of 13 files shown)


brendandburns:
Can you explain more where the problem is? Neither of these methods should be involved in deserialization as far as I know.

Perhaps add a unit test that demonstrates the issue?

wkclz:
问题在于，序列化工具会使用getter来序列化任意字段，也许也不想序列化它。当序列化到 IntOrString 时，就抛出异常了。
The problem is, serialization tool would serialize any field with getter even if you don't want to serialize it.. When we serialize an object witch contant IntOrString, we got the exception.

这里写了一个 Gson 的序列化适配，但对于Jackson (Spring框架的默认序列化工具)，却没法适配，所以可以找到 34个由此导致的 issue.
Here is an adapter with Gson, but not for Jackson, Jackson is the default serialization tool in spring framework. So we got the issues like: #2593, #2566, (34 issues). 

我没搞明白为什么这里有一个不明确类型的类，也许它本应该是一个字符串，在需要它时，再转换为数字。
I don't understand why there is such an ambiguous type, perhaps it should always be a string, and then judge whether it should be a number when needed.

wkclz:
demo: #2566

brendandburns:
IntOrString is part of the Kubernetes spec, so it's needed. The API server chooses to interpret it is either a string or an int and so we really can't do anything other than the way it is currently done. The Kubernetes API server sees `"1"` as different than `1`.

I think that you need to write a custom serialization adapter for Jackson:

https://www.baeldung.com/jackson-custom-serialization

wkclz:
It's okay now.

wkclz:
但我仍然认为不应该设置一个不准确的返回类型给序列化工具，而导致增加额外的工作去适配，尤其是抛出异常的方式。
> But I still believe that we should not give an inaccurate type to serialization tools, in particular an Exception, it would make us more workload.

后续优化建议这两个方向，看是否合理：
> Here are two suggestions if reasonable:

- 在尝试获取错误的类型时，返回空而不是抛出异常，会对序列化工具更友好，也能准确判断出 String 或 Integer
> When try to get an error type of the value, return null instead of exception. it will be friendly to serialization tools, and can accurately determine String or Integer.

- 使用两个字段表示不同的值，在将数据推送给 Kubernetes API时，再进行判别
> Using two different fields to save two type values (one of them is null), and when push to Kubernetes API server, choose the right one.

或者，在这次修改后功能已经稳定了，不再做任何改动。
> Finally, we would do nothing after this PR if it's already stable enough

kubernetes/src/main/java/io/kubernetes/client/custom/IntOrStringJacksonSerializer.java
```diff
@@ -0,0 +1,39 @@
+/*
+Copyright 2020 The Kubernetes Authors.
```
brendandburns, line 2: nit: 2024

brendandburns:
/lgtm
/approve

Thanks for the PR, especially thanks for writing unit tests!

brendandburns:
Sorry one last thing, can you remove the wiremock changes? I'd rather that was sent as a separate PR.

wkclz:
The old version of wiremock dependency old Jackson and conflicts with it.
But it can update wiremock (with Jackson) and then fix IntOrString.

Do we need to split into two PR?

yue9944882:
+1 separating into two PRs for clarity

brendandburns:
Thanks, please send the wiremock update PR and we will merge it first, then you can rebase this PR.

codefromthecrypt:
@wkclz thanks for sorting this out!

wkclz:
update: update wiremock & jackson #3126

fix: Add JacksonSerializer to serialize IntOrString #3127

Then, this PR can be reject and close.

PR #3116, opened 2024-02-26
Redact API key from debug logging

codefromthecrypt:
This redacts the API key, which ends up in Authorization header in debug logs. This helps both reduce the size and also protect secrets.

As a side effect, this adds a mockwebserver test, matching version with the okhttp client used in openapi generated code. This part is important as we improve the debug experience for watch events later.

### DIFF
--- a/kubernetes/src/main/java/io/kubernetes/client/openapi/ApiClient.java
+++ b/kubernetes/src/main/java/io/kubernetes/client/openapi/ApiClient.java
@@ -38,7 +38,6 @@ import java.security.SecureRandom;
 import java.security.cert.Certificate;
 import java.security.cert.CertificateException;
 import java.security.cert.CertificateFactory;
-import java.security.cert.X509Certificate;
 import java.text.DateFormat;
 import java.time.LocalDate;
 import java.time.OffsetDateTime;
@@ -51,7 +50,6 @@ import java.util.regex.Pattern;
 
 import io.kubernetes.client.openapi.auth.Authentication;
 import io.kubernetes.client.openapi.auth.HttpBasicAuth;
-import io.kubernetes.client.openapi.auth.HttpBearerAuth;
 import io.kubernetes.client.openapi.auth.ApiKeyAuth;
 
 /**
@@ -88,6 +86,8 @@ public class ApiClient {
     private OkHttpClient httpClient;
     private JSON json;
 
+    // Visible for testing
+    HttpLoggingInterceptor.Logger debugLogger = HttpLoggingInterceptor.Logger.DEFAULT;
     private HttpLoggingInterceptor loggingInterceptor;
 
     /**
@@ -518,8 +518,9 @@ public class ApiClient {
     public ApiClient setDebugging(boolean debugging) {
         if (debugging != this.debugging) {
             if (debugging) {
-                loggingInterceptor = new HttpLoggingInterceptor();
+                loggingInterceptor = new HttpLoggingInterceptor(debugLogger);
                 loggingInterceptor.setLevel(Level.BODY);
+                loggingInterceptor.redactHeader("authorization");
                 httpClient = httpClient.newBuilder().addInterceptor(loggingInterceptor).build();
             } else {
                 final OkHttpClient.Builder builder = httpClient.newBuilder();
--- a/kubernetes/src/main/java/io/kubernetes/client/openapi/ApiClient.java
+++ b/kubernetes/src/main/java/io/kubernetes/client/openapi/ApiClient.java
@@ -12,16 +12,9 @@ limitations under the License.
 */
 package io.kubernetes.client.openapi;
 
-import okhttp3.*;
-import okhttp3.internal.http.HttpMethod;
-import okhttp3.internal.tls.OkHostnameVerifier;
-import okhttp3.logging.HttpLoggingInterceptor;
-import okhttp3.logging.HttpLoggingInterceptor.Level;
-import okio.Buffer;
-import okio.BufferedSink;
-import okio.Okio;
-
-import javax.net.ssl.*;
+import io.kubernetes.client.openapi.auth.ApiKeyAuth;
+import io.kubernetes.client.openapi.auth.Authentication;
+import io.kubernetes.client.openapi.auth.HttpBasicAuth;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
@@ -47,15 +40,21 @@ import java.util.Map.Entry;
 import java.util.concurrent.TimeUnit;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-
-import io.kubernetes.client.openapi.auth.Authentication;
-import io.kubernetes.client.openapi.auth.HttpBasicAuth;
-import io.kubernetes.client.openapi.auth.ApiKeyAuth;
+import javax.net.ssl.*;
+import okhttp3.*;
+import okhttp3.internal.http.HttpMethod;
+import okhttp3.internal.tls.OkHostnameVerifier;
+import okhttp3.logging.HttpLoggingInterceptor;
+import okhttp3.logging.HttpLoggingInterceptor.Level;
+import okio.Buffer;
+import okio.BufferedSink;
+import okio.Okio;
 
 /**
  * <p>ApiClient class.</p>
  */
 public class ApiClient {
+    static final String AUTHORIZATION_NAME = "authorization";
 
     private String basePath = "http://localhost";
     protected List<ServerConfiguration> servers = new ArrayList<ServerConfiguration>(Arrays.asList(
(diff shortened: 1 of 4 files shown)


kubernetes/src/test/java/io/kubernetes/client/openapi/ApiClientTest.java
```diff
@@ -0,0 +1,101 @@
...
+            .addHeader("Content-Type", "application/json")
+            .setBody(TEST_RESPONSE_BODY));
+
+    // Ensure the request completes, by verifying it received the expected response
+    try (Response response = testCall.execute()) {
+      assertThat(response.body().string(), equalTo(TEST_RESPONSE_BODY));
+    }
+
+    // Verify the server saw the API key
+    assertThat(server.takeRequest().getHeader("authorization"), equalTo(TEST_API_KEY));
```
codefromthecrypt, line 95: ps I would personally love to switch everything off hamcrest to assertj, something that is mostly automatable with openrewrite. However, for this change I left it alone. Let me know and I'll happily fix everything to latest assertj instead.

See https://www.moderne.io/blog/migrating-from-hamcrest-to-assertj-with-openrewrite-and-moderne

codefromthecrypt:
TL;DR; I'll take the okhttp mock server out for now, but it would be my strong preference for the follow-up change that would be less annoying to test if I can use this (the debug with streaming thing)

> We have typically used Wiremock for mocking HTTP calls instead of mockwebserver. I'd prefer to stick with a single way of mocking HTTP unless there is a strong reason to use it for this use case.

Sure, indeed this warrants explanation. So, the client is written in okhttp, and the most natural, most popular and longest living way to test okhttp clients is with their own mock framework. Besides functional reasons, you have only one version to manage, and less build CVEs to worry about.

I recently worked on some code that uses wire mock and found some tests commenting difficulty with redirect tests from one thing to another. I have personally never had a problem with okhttp and mainly put the infrastructure here as it has straight-forward code to setup tests for web sockets and streaming responses.

I would offer "freedom and responsibility" on this one and offer to redo all the other tests to okhttp's mock server and then you can see how clean or not it is. Meanwhile, I'll revert this to wire mock as agree personally I don't like "winky changes" where some parts do one thing and another something else.

codefromthecrypt:
Changed to use wire mock (updating its version as it was quite old).

Note that this dropped test performance time from .2ms to over 5 seconds. This was running it several times in the same IDE.

Screenshot 2024-02-27 at 07 35 46
Screenshot 2024-02-27 at 07 36 53

brendandburns:
Thanks for switching to WireMock.

One nit, can you revert the version update and send as a separate PR? I'd prefer to keep PRs clean and focused on a single change.

If you are willing to write the PR to move to mock okhttp server I'm happy to review it, but I also don't guarantee it will merge, so I'd hate to waste your time.

I haven't really had any problems with WireMock and it feels unnecessary to change, but if you're motivated to do the work, I'm not going to stop you, my main requirement is that we only have one.

@yue9944882 wdyt?

codefromthecrypt:
@brendandburns agree some are fine with wiremock, just in reflection it was the odd choice to begin with. It also has a lot of dependency interference, so trips up things like snyk etc due to the vast amount of dependencies it brings in. Regardless, I'll keep this topic to a separate PR and you both can decide.

```
[INFO] \- com.github.tomakehurst:wiremock:jar:2.27.2:test
[INFO]    +- org.eclipse.jetty:jetty-server:jar:9.2.28.v20190418:test
[INFO]    |  +- javax.servlet:javax.servlet-api:jar:3.1.0:test
[INFO]    |  +- org.eclipse.jetty:jetty-http:jar:9.2.28.v20190418:test
[INFO]    |  \- org.eclipse.jetty:jetty-io:jar:9.2.28.v20190418:test
[INFO]    +- org.eclipse.jetty:jetty-servlet:jar:9.2.28.v20190418:test
[INFO]    |  \- org.eclipse.jetty:jetty-security:jar:9.2.28.v20190418:test
[INFO]    +- org.eclipse.jetty:jetty-servlets:jar:9.2.28.v20190418:test
[INFO]    |  +- org.eclipse.jetty:jetty-continuation:jar:9.2.28.v20190418:test
[INFO]    |  \- org.eclipse.jetty:jetty-util:jar:9.2.28.v20190418:test
[INFO]    +- org.eclipse.jetty:jetty-webapp:jar:9.2.28.v20190418:test
[INFO]    |  \- org.eclipse.jetty:jetty-xml:jar:9.2.28.v20190418:test
[INFO]    +- org.eclipse.jetty:jetty-proxy:jar:9.2.28.v20190418:test
[INFO]    |  \- org.eclipse.jetty:jetty-client:jar:9.2.28.v20190418:test
[INFO]    +- com.google.guava:guava:jar:20.0:test
[INFO]    +- com.fasterxml.jackson.core:jackson-core:jar:2.11.0:test
[INFO]    +- com.fasterxml.jackson.core:jackson-annotations:jar:2.16.1:test
[INFO]    +- com.fasterxml.jackson.core:jackson-databind:jar:2.11.0:test
[INFO]    +- org.apache.httpcomponents:httpclient:jar:4.5.12:test
[INFO]    |  +- org.apache.httpcomponents:httpcore:jar:4.4.13:test
[INFO]    |  +- commons-logging:commons-logging:jar:1.2:test
[INFO]    |  \- commons-codec:commons-codec:jar:1.16.1:test
[INFO]    +- org.xmlunit:xmlunit-core:jar:2.7.0:test
[INFO]    |  \- javax.xml.bind:jaxb-api:jar:2.3.0:test
[INFO]    +- org.xmlunit:xmlunit-legacy:jar:2.7.0:test
[INFO]    +- org.xmlunit:xmlunit-placeholders:jar:2.7.0:test
[INFO]    +- com.jayway.jsonpath:json-path:jar:2.4.0:test
[INFO]    |  \- net.minidev:json-smart:jar:2.3:test
[INFO]    |     \- net.minidev:accessors-smart:jar:1.2:test
[INFO]    +- org.ow2.asm:asm:jar:7.0:test
[INFO]    +- net.sf.jopt-simple:jopt-simple:jar:5.0.3:test
[INFO]    +- com.github.jknack:handlebars:jar:4.0.7:test
[INFO]    |  \- org.antlr:antlr4-runtime:jar:4.7.1:test
[INFO]    +- com.github.jknack:handlebars-helpers:jar:4.0.7:test
[INFO]    +- com.flipkart.zjsonpatch:zjsonpatch:jar:0.4.16:test
[INFO]    |  \- org.apache.commons:commons-collections4:jar:4.4:test
[INFO]    \- commons-fileupload:commons-fileupload:jar:1.4:test
[INFO]       \- commons-io:commons-io:jar:2.15.1:test
```

codefromthecrypt:
note: the part about wiremock -> okhttp is really low priority for me, so I won't likely do this soon. I'm more interested in the other in-flight changes.

kubernetes/src/main/java/io/kubernetes/client/openapi/ApiClient.java
```diff
@@ -518,8 +518,9 @@ public boolean isDebugging() {
     public ApiClient setDebugging(boolean debugging) {
         if (debugging != this.debugging) {
             if (debugging) {
-                loggingInterceptor = new HttpLoggingInterceptor();
+                loggingInterceptor = new HttpLoggingInterceptor(debugLogger);
                 loggingInterceptor.setLevel(Level.BODY);
+                loggingInterceptor.redactHeader("authorization");
```
yue9944882, line 523: is it case sensitive? can we add a test cases to see if `Authorization` is also redacted?

codefromthecrypt: sure. I think we are the ones adding this header and we add it as lowercase, but no problem regardless

brendandburns:
@codefromthecrypt btw, someone else may also send the update to the wiremock dependency.

https://github.com/kubernetes-client/java/pull/3113

kubernetes/src/main/java/io/kubernetes/client/openapi/ApiClient.java
codefromthecrypt: done

wkclz:
Updated wiremock in PR #3126

brendandburns:
Sorry to make more trouble, I realized that this was in a generated file (ApiClient.java) we will need to add this as a patch here:

https://github.com/kubernetes-client/java/tree/master/scripts/patches

So that it persists after a new code generation.

codefromthecrypt:
raised upstream here https://github.com/OpenAPITools/openapi-generator/pull/18010

I'm out of time for the day, so I'll look into how to create and verify the diff tomorrow.

brendandburns:
@codefromthecrypt fwiw, the openapi project has been really good about accepting fixes, so if you'd rather submit a PR to that project, we can then regenerate the code. It is a little slower, but it does have benefits for more than just this project.

Either way is fine with me.

codefromthecrypt:
good idea. I'll close this one out and we can track https://github.com/OpenAPITools/openapi-generator/pull/18010 instead and just apply it once done.

codefromthecrypt:
@brendandburns I would really like to remove the tech debt caused by wiremock. If you ack https://github.com/kubernetes-client/java/issues/3146 I'll give free labor on it.

PR #3118, opened 2024-02-27
update maven wrapper to latest, and adjust CI to use it

codefromthecrypt:
This updates the maven wrapper to the latest version, and adjusts test and release CI config to use it.

This is something we should land prior to other tech debt cleanup, including updates of old plugin versions in our poms. If we continue the master-java8 branch (something I advise against), we should apply this change there also.

Things intentionally not done, mostly to avoid conversations that are independent of this:
* I didn't update the dev container which installs a different maven and possibly also gradle even though gradle isn't used.
* I didn't update CI for jobs run in other repos (notably gen). I am happy to update the gen repo as well, then once that's up to date switch to here
* I didn't update the examples READMEs as many of them are broken and should be fixed in another PR


.mvn/wrapper/MavenWrapperDownloader.java
```diff
@@ -1,117 +0,0 @@
...
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-import java.net.*;
-import java.io.*;
-import java.nio.channels.*;
-import java.util.Properties;
-
-public class MavenWrapperDownloader {
```
codefromthecrypt, line 21: as you'll notice, the jar and properties were checked in, so this downloader script wasn't used. takari's wrapper is no longer used, as modern maven versions have a version of the same. It is commonplace to check in the wrapper jar as we were doing prior as well.

brendandburns:
/lgtm
/approve

We should cherry-pick this into the `java8` branch. We need to maintain a java8 compatible client for the time being, but keeping the entire repo java8 compatible was holding us back.

brendandburns:
Also, I really don't like the fact that we have these scripts checked in. Is there a better way to manage this so that we always download from an authoritative source?

codefromthecrypt:
Fixed the paths I think. Let me know if you want a second PR or you will cherry-pick offline, either works.

> We should cherry-pick this into the java8 branch. We need to maintain a java8 compatible client for the time being, but keeping the entire repo java8 compatible was holding us back.

Aside to this PR, but.. right now the bytecode level on master is set to java 8, so I guess this is about being able to use JDK 8 to compile the project, or some other legacy API reason, vs consuming bytecode as java 8

```
$ file kubernetes/target/classes//io/kubernetes/client/custom/V1Patch.class
kubernetes/target/classes//io/kubernetes/client/custom/V1Patch.class: compiled Java class data, version 52.0 (Java 1.8)
```

Anyway, keep us posted on killing that branch, like maybe an issue saying what circumstances can stop it. When we do refactoring, cherry-picking will start to become difficult, or we'll have to refactor both branches. Food for thought.

codefromthecrypt:
@yue9944882 

> i think we should bump it to java11 even though i don't think the code on master branch can compile as it's currently using Jarkarta dependency.

Leave this to me. I'll do it in another PR, so that this one can be cherry-picked

PR #3125, opened 2024-02-28
Updates to minimum Java 11

codefromthecrypt:
This updates to minimum Java 11 [per interest](https://github.com/kubernetes-client/java/pull/3118#issuecomment-1968064738) by @yue9944882

1. use openrewrite plugin (first time here!)
```
$./mvnw -U org.openrewrite.maven:rewrite-maven-plugin:run -Drewrite.recipeArtifactCoordinates=org.openrewrite.recipe:rewrite-migrate-java:RELEASE -Drewrite.activeRecipes=org.openrewrite.java.migrate.Java8toJava11
```
2. fix up formatting fuzz

I stopped there, though sometimes IntelliJ also has some migrations it is aware of, too. Thanks @timtebeek for keeping this plugin in top form!


timtebeek:
You're welcome; glad to see these recipes run here. Note that this is a one-off recipe run, but you can also build up a collection of recipes that you want to run as a check on any future pull requests, and gently nudge users to adopt their code suggestions. See this PR as an example of how that can be achieved:
- https://github.com/langchain4j/langchain4j/pull/673

proto/pom.xml
```diff
@@ -34,9 +34,8 @@
 			<plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
-                <configuration>
-                    <source>8</source>
-                    <target>8</target>
+								<configuration>
+							  		<release>${java.version}</release>
```
brendandburns, line 38: nit: fix indentation.

codefromthecrypt: yep good catch.

codefromthecrypt:
looks like a little tech debt. I will raise in another PR

```
Error:  /home/runner/work/java/java/kubernetes/src/main/java/io/kubernetes/client/openapi/JSON.java:19: error: package com.google.gson.internal.bind.util is not visible
Error:  import com.google.gson.internal.bind.util.ISO8601Utils;
Error:                                      ^
Error:    (package com.google.gson.internal.bind.util is declared in module com.google.gson, which does not export it)
```

codefromthecrypt:
above comment is tricky to unwind quickly. I'll find another way in this PR, so it isn't locked up too long.

codefromthecrypt:
ok I think I found a workaround until we can solve the root problem once and for all, that the openapi generator is making a dependency on internal types.

PR #3133, opened 2024-02-29
use code generation with jakarta instead of javax

### LINKED ISSUE #3117: crd-model-gen generates models with javax annotations
**Describe the bug**
I am using `ghcr.io/kubernetes-client/java/crd-model-gen:v1.0.6` to generate some models based on CRDs.
The generated code uses `@javax.annotation.Nullable` and `@javax.annotation.Generated` annotations.
Is it possible to generate code with Jakarta annotations?

I found this issue https://github.com/kubernetes-client/gen/issues/246 which supposedly allows use of `useJakartaEe` and I saw that this option is set to true here: https://github.com/kubernetes-client/gen/blob/master/openapi/java.xml#L105.
But last version of crd-model-gen was published 2 years ago and still has the older version.
I'm not sure, but maybe releasing new version would solve this problem?

**Client Version**
`1.19`

**Kubernetes Version**
`1.24`

**Java Version**
Java 8

**To Reproduce**
- generate models using `CRD Java Model Generate` workflow

**Expected behavior**
Generated code has jakarta annotations instead of javax.

**KubeConfig**
If applicable, add a KubeConfig file with secrets redacted.

**Server (please complete the following information):**
 - OS: Linux
 - Environment [e.g. container]
 - Cloud [e.g. Azure]

**Additional context**

kozjan:
This updates crd-model-gen image to use current version of `gen` used - latest commit from master, which contains changes that allow generation of models with jakarta annotations instead of javax.

I removed `Dockerfile.gen` and `Dockerfile.gen-base` since they look like they are based on this Dockerfile: https://github.com/kubernetes-client/gen/blob/master/openapi/openapi-generator/Dockerfile
And if I understand correctly, they are not exactly used - the `publish-gen-image.sh` publishes the same image to 2 different names, unless maybe I don't understand how Docker works and this .gen .gen-base suffix does something.

Other unexpected behaviour:
- generated code contained `AbstractOpenApiSchema.java` file, which I explicitly remove in `generate.sh` - added a comment to the issue which I think is linked to this problem
- previously I wasn't using -n flag to pass the prefix of the target CRD's api group and it still only generated necessary classes - now it generates many more unless this prefix is specified

Closes #3117


kozjan:
hey, I ran update.sh scripts and commited the results to show what differences come up after updating the image - it seems like a lot of classes are added, but I checked it briefly and it looks to me like they are all in the crds used for generation, I guess more of them were generated due to changes in OpenAPI generator

kozjan:
Skipped spotless in prometheus-operator, but still have one more problem - one of the crds has [enum with lowercase and camelcase values](https://github.com/prometheus-operator/prometheus-operator/blob/main/example/prometheus-operator-crd/monitoring.coreos.com_podmonitors.yaml#L246), which generates repeated enums:
```
public enum ActionEnum {
    REPLACE("replace"),
    REPLACE("Replace"),
    ...
    }
```
which fails the checks since it can't build - I am not sure what can be done about this.

kozjan:
changed version of the image to v2.0.0 since there were many changes in openapi and also javax -> jakarta swap

kozjan:
I rolled back update.sh of prometheus-operator, so that checks can pass, since I think the problem is in the specific CRDs it uses as described here: https://github.com/kubernetes-client/java/pull/3133#issuecomment-1975878132

PR #3136, opened 2024-03-01
Migrate all test assertions to AssertJ

codefromthecrypt:
This migrates all test assertions to AssertJ. For example, we no longer use assertions from JUnit (4) or hamcrest. As a result, this removes the dependency on hamcrest.

Notes:
 * There are a couple places where it seems we fail tests on an exception not raised on the test thread. I retained JUnit's `fail` function for that, with a TODO for once we move to jupiter.
 * There were a couple places we used hamcrest matchers in awaitility, so I used a normal predicate instead

On OpenRewrite:

The openrewrite recipes for `MigrateHamcrestToAssertJ` and `JUnitToAssertj` didn't affect many files.. I'm not sure why. I did most of this manually via IntelliJ regex replacements. Later, I cleaned up by running `UnnecessaryThrows` `RemoveUnusedImports` (then reverting the changes to main) and finally `Assertj` to polish existing and new usage. cc @timtebeek

e.g. This applies current AssertJ best practices.
```bash
./mvnw -U org.openrewrite.maven:rewrite-maven-plugin:run -Drewrite.recipeArtifactCoordinates=org.openrewrite.recipe:rewrite-testing-frameworks:RELEASE -Drewrite.activeRecipes=org.openrewrite.java.testing.assertj.Assertj
```

### DIFF
--- a/e2e/pom.xml
+++ b/e2e/pom.xml
@@ -39,6 +39,11 @@
             <artifactId>junit</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <scope>test</scope>
+        </dependency>
 
         <dependency>
             <groupId>org.slf4j</groupId>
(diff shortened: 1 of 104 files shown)


codefromthecrypt:
ps reason I raised this is basically I kept accumulating reasons. final straw was trying to get rid of that internal iso date dep on gson, by using the same tests, and noticing we didn't have assertj in all projects. Note: we were using assertj before, just not consistently.

codefromthecrypt:
also, once this is in, migration to jupiter should be a lot more straightforward. We should migrate to jupiter because junit is no longer receiving maintenance and tends to create CVE noise.

codefromthecrypt:
rebased

PR #3139, opened 2024-03-01
Reduce allocations in ProtoClient codec

codefromthecrypt:
I was trying to understand the proto client and noticed some things that could be improved, if you are pretty familiar with how okio works. I ran the ProtoExample afterwards and it has the same behaviors.

Note: The upstream ApiClient isn't managing resources very well, particularly ensuring responses close. So, even though I adjusted in proto, there are other paths which can leak responses, especially on exception.

### DIFF
--- a/util/src/main/java/io/kubernetes/client/ProtoClient.java
+++ b/util/src/main/java/io/kubernetes/client/ProtoClient.java
@@ -12,6 +12,7 @@ limitations under the License.
 */
 package io.kubernetes.client;
 
+import com.google.protobuf.CodedOutputStream;
 import com.google.protobuf.Message;
 import com.google.protobuf.Message.Builder;
 import io.kubernetes.client.openapi.ApiClient;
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
@brendanburns I don't want to swamp you with review work, but if you are keen to have a "benchmarks" module which uses JMH, one day I'll add it. That way, you can see before/after and we don't need to guess if allocations are better or worse than before. It works similar (enough) to go's bench, even if not as convenient as isn't built-in.

brendandburns:
/lgtm
/approve

The proto client hasn't been worked on in a _long_ time and the protocol buffer objects haven't been regenerated for an even longer time. So if you want to take on improvements in this area, contributions are very welcome, but there's definitely going to be rough edges in the current experience.

Here's where the protos are generated:
https://github.com/kubernetes-client/gen/tree/master/proto

PR #3142, opened 2024-03-01
Regenerates proto source

codefromthecrypt:
regenerated per notes by [here](https://github.com/kubernetes-client/java/pull/3139#issuecomment-1973320983)

then, I ran the proto example
```bash
$ cd /examples/examples-release-latest
$  mvn -q exec:java -Dexec.mainClass="io.kubernetes.client.examples.ProtoExample"
metadata {
  name: "demo-demo1-7dc45dc8c9-4nkx6"
  generateName: "demo-demo1-7dc45dc8c9-"
  namespace: "default"
  selfLink: ""
  uid: "e5bb5ab2-936a-494e-9df3-dc6f8678a8af"
  resourceVersion: "86377"
  generation: 0
  creationTimestamp {
    seconds: 1708674165
--snip--
```

Thanks for the tips @brendanburns!


brendandburns:
I think we need to add an exclude for the proto files directory to the format-cop check similar to this:

https://github.com/kubernetes-client/java/blob/master/fluent/pom.xml#L43

codefromthecrypt:
thanks, I also fixed the formatting in the proto pom as it was unlike others

PR #3144, opened 2024-03-03
Skip deployment of E2E test module to maven repository

codefromthecrypt:
also removed redundant versions in the skip config for examples


examples/examples-release-17/pom.xml
```diff
@@ -107,7 +107,6 @@
 			<plugin>
 				<groupId>org.apache.maven.plugins</groupId>
 				<artifactId>maven-deploy-plugin</artifactId>
-				<version>3.1.1</version>
```
codefromthecrypt, line 110: when the version is set in dependencyManagement section of the parent, you don't need to repeat it in any child module.

Issue #3135, opened 2024-02-29
Migrate from spock to normal junit (eventually jupiter)


codefromthecrypt:
Right now, E2E tests are using a different programming language, groovy, and a less used framework, spock. This complicates both the act of adding new tests as well dealing with dependencies.

At the time, it may have made sense to try something like Spock. However, if you look closely at the E2E tests, they are simple and neither need to use BDD style, nor need to use a different programming language.

Basic junit is easier for new maintainers and the code, and I believe this should be reverted to normal java. I'm happy to prove this in a PR if maintainers agree.

brendandburns:
I'm fine with this normalization.

codefromthecrypt:
> I'm fine with this normalization.

great. I'll raise a PR as it will make other infra migrations (like junit -> Jupiter) much easier and comprehensive as they'll also apply to E2E.

PR #3145, opened 2024-03-03
Migrates e2e tests from spock to plain JUnit

codefromthecrypt:
Per rationale in #3135, this migrates from groovy/spock to Java/JUnit/AssertJ. This reduces the complexity of the project and its dependencies, and makes an easier migration to JUnit Jupiter. Finally, it eliminates the need for people to learn Groovy only to maintain tests.


codefromthecrypt:
rebased

PR #3176, opened 2024-03-08
Fix CVEs brought in by accidental spring boot version override

codefromthecrypt:
We accidentally masked the parent definition of spring-boot version and in doing so pin older dependencies than we meant to. This was detected by `trivy repo .`.


brendandburns:
I'm surprised that dependabot didn't notice this...

/lgtm
/approve

PR #3178, opened 2024-03-09
Migrate to JUnit Jupiter

### LINKED ISSUE #3135: Migrate from spock to normal junit (eventually jupiter)

codefromthecrypt:
This migrates to JUnit Jupiter with a bulk of help by OpenRewrite, but due to some use of systemstubs and wiremock, a significant amount of manual work also. It is surely the case this was easier with OpenRewrite despite the glitches.

Here are the main recipes that helped, though I couldn't run sequentially and in some cases I needed to roll back some changes.

```bash
$ mvn -U org.openrewrite.maven:rewrite-maven-plugin:run -Drewrite.recipeArtifactCoordinates=org.openrewrite.recipe:rewrite-testing-frameworks:RELEASE -Drewrite.activeRecipes=org.openrewrite.java.testing.junit5.JUnit4to5Migration
$  mvn -U org.openrewrite.maven:rewrite-maven-plugin:run -Drewrite.recipeArtifactCoordinates=org.openrewrite.recipe:rewrite-testing-frameworks:RELEASE -Drewrite.activeRecipes=org.openrewrite.java.testing.junit5.JUnit5BestPractices
# below I have to revert the changes to src/main as they are generated files
$ mvn -U org.openrewrite.maven:rewrite-maven-plugin:run -Drewrite.activeRecipes=org.openrewrite.java.RemoveUnusedImports
```

Here are notes on manual changes I did, most were functional, but some to remove distraction. Anecdotally, everything I did this morning and spent between 3 and 4 hours soup to nuts. This would have taken a lot longer without openrewrite.
* pom changes had to be redone
* most wiremock tests had to be edited manually due to use of static imports
* systemstubs had to be redone
* spring tests had to polished to not use redundant annotations
* temp directories had to be redone to simplify the result. this was in combination of moving from File to Path
* parameterized tests added some boilerplate looked better if was removed (think single test)
* some assertions became jupiter and had to be rolled back to assertj
* javadoc link to tests renamed from testX to x drift
* some test fixtures were copy pasted from other tests but not referenced. I removed these vs port them for no reason
* some tests were missing `@Test` in the first place so weren't run ever or in a long time. I added the annotation rather than delete those tests.

Fixes #3135

### DIFF
--- a/examples/examples-release-18/src/test/java/io/kubernetes/client/examples/ExampleTest.java
+++ b/examples/examples-release-18/src/test/java/io/kubernetes/client/examples/ExampleTest.java
@@ -14,33 +14,36 @@ package io.kubernetes.client.examples;
 
 import static com.github.tomakehurst.wiremock.client.WireMock.aResponse;
 import static com.github.tomakehurst.wiremock.client.WireMock.get;
-import static com.github.tomakehurst.wiremock.client.WireMock.stubFor;
 import static com.github.tomakehurst.wiremock.client.WireMock.urlEqualTo;
-import static org.junit.Assert.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertEquals;
 
-import com.github.tomakehurst.wiremock.junit.WireMockRule;
+import com.github.tomakehurst.wiremock.core.WireMockConfiguration;
+import com.github.tomakehurst.wiremock.junit5.WireMockExtension;
 import io.kubernetes.client.openapi.ApiClient;
 import io.kubernetes.client.openapi.ApiException;
 import io.kubernetes.client.openapi.Configuration;
 import io.kubernetes.client.openapi.apis.CoreV1Api;
 import io.kubernetes.client.openapi.models.V1Namespace;
 import io.kubernetes.client.openapi.models.V1ObjectMeta;
-import org.junit.Rule;
-import org.junit.Test;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
 
-public class ExampleTest {
+class ExampleTest {
   private static final int PORT = 8089;
-  @Rule public WireMockRule wireMockRule = new WireMockRule(PORT);
+
+  @RegisterExtension
+  static WireMockExtension apiServer =
+      WireMockExtension.newInstance().options(WireMockConfiguration.options().port(PORT)).build();
 
   @Test
-  public void exactUrlOnly() throws ApiException {
+  void exactUrlOnly() throws ApiException {
     ApiClient client = new ApiClient();
     client.setBasePath("http://localhost:" + PORT);
     Configuration.setDefaultApiClient(client);
 
     V1Namespace ns1 = new V1Namespace().metadata(new V1ObjectMeta().name("name"));
 
-    stubFor(
+    apiServer.stubFor(
         get(urlEqualTo("/api/v1/namespaces/name"))
             .willReturn(
                 aResponse()
--- a/util/src/test/java/io/kubernetes/client/ExecTest.java
+++ b/util/src/test/java/io/kubernetes/client/ExecTest.java
@@ -25,11 +25,8 @@ import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 
-import com.github.tomakehurst.wiremock.core.Admin;
 import com.github.tomakehurst.wiremock.extension.Parameters;
-import com.github.tomakehurst.wiremock.extension.PostServeAction;
-import com.github.tomakehurst.wiremock.junit.WireMockRule;
-import com.github.tomakehurst.wiremock.stubbing.ServeEvent;
+import com.github.tomakehurst.wiremock.junit5.WireMockExtension;
 import io.kubernetes.client.Exec.ExecProcess;
 import io.kubernetes.client.openapi.ApiClient;
 import io.kubernetes.client.openapi.ApiException;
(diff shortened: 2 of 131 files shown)


codefromthecrypt:
cc @timtebeek

examples/examples-release-18/src/test/java/io/kubernetes/client/examples/ExampleTest.java
```diff
@@ -14,33 +14,36 @@
...
-  public void exactUrlOnly() throws ApiException {
+  void exactUrlOnly() throws ApiException {
     ApiClient client = new ApiClient();
     client.setBasePath("http://localhost:" + PORT);
     Configuration.setDefaultApiClient(client);
 
     V1Namespace ns1 = new V1Namespace().metadata(new V1ObjectMeta().name("name"));
 
-    stubFor(
+    apiServer.stubFor(
```
codefromthecrypt, line 46: this is an example which bit me.. basically when using the junit rule, you can mix/match using static method or the method on the rule. in the extension you must use the method and not use the static.

util/src/test/java/io/kubernetes/client/ExecTest.java
```diff
@@ -46,25 +43,12 @@
...
-import org.junit.Rule;
-import org.junit.Test;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
 
 /** Tests for the Exec helper class */
-public class ExecTest {
-
-  public static class CountRequestAction extends PostServeAction {
```
codefromthecrypt, line 56: this was never referenced, so I deleted it

examples/examples-release-18/src/test/java/io/kubernetes/client/examples/ExampleTest.java
codefromthecrypt: when you don't do this, things will fail about reaching port 8080 and you have to think through why. I think this maybe should become an openrewrite recipe for wiremock (removing the static method footgun and always even in junit 4 using the instance to do stubbing and verification cc @timtebeek)

codefromthecrypt:
thanks for clicking run.

looks like problem on windows:
```
Error:  Errors: 
Error:    FilePersisterTest.persistence(Path) � IO Failed to delete temp directory C:\Users\RUNNER~1\AppData\Local\Temp\junit5731886460482893149. The following paths could not be deleted (see suppressed exceptions for details): <root>, testconfig1129927534881067784.tmp
Error:    KubeConfigTest.gcpAuthProvider(Path) � IO Failed to delete temp directory C:\Users\RUNNER~1\AppData\Local\Temp\junit1608042563520011234. The following paths could not be deleted (see suppressed exceptions for details): <root>, config15826931468867220118.tmp
Error:    KubeConfigTest.gcpAuthProviderStringDate(Path) � IO Failed to delete temp directory C:\Users\RUNNER~1\AppData\Local\Temp\junit15928002169510443350. The following paths could not be deleted (see suppressed exceptions for details): <root>, config14368825936767128865.tmp
```

and also some issue in LeaderElectorTest.

I'll try to bite these off in the next couple days as a change like this has massive drift work risk

codefromthecrypt:
cool looks like green!

examples/examples-release-18/src/test/java/io/kubernetes/client/examples/ExampleTest.java
timtebeek: Thanks for the suggestion! That's bit me before in the past as well; would be neat to have a recipe that changes any static method use to use through `rule` / `extension` ahead of any other WireMock migration steps. A good addition to this earlier issue.
- https://github.com/openrewrite/rewrite-testing-frameworks/issues/278

brendandburns:
Thanks for the patience!

timtebeek:
Great to see this merged! I've done a quick check through the platform and it seems there's two files that mere missed, perhaps as part of resolving a merge conflict. If desired a PR can be opened from this URL: https://app.moderne.io/results/KozquRLdd
```patch
diff --git a/util/src/test/java/io/kubernetes/client/CopyTest.java b/util/src/test/java/io/kubernetes/client/CopyTest.java
index 82fce66..3611992 100644
--- a/util/src/test/java/io/kubernetes/client/CopyTest.java
+++ b/util/src/test/java/io/kubernetes/client/CopyTest.java
@@ -183,7 +183,7 @@ io.moderne.worker.execution.RecipeMapper
   }
 
   @Test
-  void testCopyDirectoryFromPod(@TempDir Path tempDir) throws Exception {
+  void copyDirectoryFromPod(@TempDir Path tempDir) throws Exception {
     Copy copy = new Copy(client);
 
     apiServer.stubFor(
diff --git a/util/src/test/java/io/kubernetes/client/ExecCallbacksTest.java b/util/src/test/java/io/kubernetes/client/ExecCallbacksTest.java
index d150eb0..3e99fc9 100644
--- a/util/src/test/java/io/kubernetes/client/ExecCallbacksTest.java
+++ b/util/src/test/java/io/kubernetes/client/ExecCallbacksTest.java
@@ -13,7 +13,6 @@ io.moderne.worker.execution.RecipeMapper
 package io.kubernetes.client;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.junit.jupiter.api.Assertions.fail;
 
 import io.kubernetes.client.custom.IOTrio;
 import io.kubernetes.client.openapi.ApiException;
@@ -31,6 +30,7 @@
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.BiConsumer;
+import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
 class ExecCallbacksTest {
@@ -201,11 +201,9 @@
   }
 
   private void quietly(NoisyRunnable task) {
-    try {
+    Assertions.assertDoesNotThrow(() -> {
       task.run();
-    } catch (Exception e) {
-      fail(e);
-    }
+    });
   }
 
   private String read(InputStream is, String expectedText) {
@@ -219,13 +217,11 @@
   }
 
   private String read(InputStream is, int count) {
-    try {
+    Assertions.assertDoesNotThrow(() -> {
       byte[] buff = new byte[count];
       is.read(buff);
       return new String(buff);
-    } catch (IOException e) {
-      fail(e);
-    }
+    });
     return null;
   }
 ```

codefromthecrypt:
hi @timtebeek. so looked intentionally doing something different, so I left it at `fail`. Also, if it weren't I would use assertj not Jupiter's assertThrows for consistency with everything else. I should have put a `TODO: why` on those, to save you the effort.

test renames are legit though!

Issue #3240, opened 2024-03-22
KubeConfig.runExec doesn't work with kubeconfig exec command


nicolas-goudry:
**Describe the bug**

This issue occurs with any kubeconfig file which uses an `exec` command to get user credentials.

When instantiating the `ApiClient` class through `ClientBuilder.kubeconfig` and a kubeconfig file built with `KubeConfig.loadKubeConfig`, the following error occurs:

```
java.lang.NullPointerException: Cannot invoke "java.io.File.toPath()" because "this.file" is null
    at io.kubernetes.client.util.KubeConfig.runExec (KubeConfig.java:333)
    at io.kubernetes.client.util.KubeConfig.getCredentialsViaExecCredential (KubeConfig.java:291)
    at io.kubernetes.client.util.KubeConfig.getCredentials (KubeConfig.java:239)
    at io.kubernetes.client.util.credentials.KubeconfigAuthentication.<init> (KubeconfigAuthentication.java:59)
    at io.kubernetes.client.util.ClientBuilder.kubeconfig (ClientBuilder.java:300)
    at io.kubernetes.client.examples.KubeConfigFileClientExample.main (KubeConfigFileClientExample.java:43)
    at org.codehaus.mojo.exec.ExecJavaMojo.lambda$execute$0 (ExecJavaMojo.java:283)
    at java.lang.Thread.run (Thread.java:833)
```

It seems that the `KubeConfig.runExec` method expects the kubeconfig to be provided as a file rather than a `KubeConfig` object.

The `KubeConfigFileClientExample.java` example can be fixed with the following changes:

```diff
diff --git a/examples/examples-release-20/src/main/java/io/kubernetes/client/examples/KubeConfigFileClientExample.java b/examples/examples-release-20/src/main/java/io/kubernetes/client/examples/KubeConfigFileClientExample.java
index 7a0ba4dcf..bb62b7365 100644
--- a/examples/examples-release-20/src/main/java/io/kubernetes/client/examples/KubeConfigFileClientExample.java
+++ b/examples/examples-release-20/src/main/java/io/kubernetes/client/examples/KubeConfigFileClientExample.java
@@ -22,6 +22,7 @@ import io.kubernetes.client.util.ClientBuilder;
 import io.kubernetes.client.util.KubeConfig;
 import java.io.FileReader;
 import java.io.IOException;
+import java.nio.file.Paths;
 
 /**
  * A simple example of how to use the Java API from an application outside a kubernetes cluster
[...]

PR #3244, opened 2024-03-23
Fix an NPE bug in command handling if there is no directory for the kubeconfig file.

brendandburns:


nicolas-goudry:
After testing, I can confirm this fixes issue #3240

brendandburns:
cc @yue9944882 for a review here.

yue9944882 (commented):
/lgtm
/approve

Issue #3140, opened 2024-03-01
Old examples use the current parent pom


codefromthecrypt:
Currently, the source tree holds old versions of examples. However, they use the parent-pom, which can lead to problems as we update dependencies.

I would suggest we either stop copying old versions of examples into the source tree (instead add README links to prior versions), or pin the examples to their corresponding parent poms.

For example, we could do this, but it might be better to just delete the old examples and only keep the current.
```diff
--- a/examples/examples-release-17/pom.xml
+++ b/examples/examples-release-17/pom.xml
@@ -6,8 +6,7 @@
        <parent>
                <groupId>io.kubernetes</groupId>
                <artifactId>client-java-examples-parent</artifactId>
-               <version>20.0.0-SNAPSHOT</version>
-               <relativePath>..</relativePath>
+               <version>17.0.0</version>
        </parent>
```

brendandburns:
The issue with deleting the older examples has to do with discoverability. If it is too hard to find the older examples, people file issues when the current examples don't work with older versions of the libraries.

If there is a proposal for how to achieve that (perhaps linking to git tags?) I'm open to proposals.

codefromthecrypt:
I was thinking we could add a link with the release tag in the release notes to where the example is. Since the tag won't change that will always be there. another way is we could make a release asset (tar.gz or zip) with the examples, if we think people will struggle with git checkout?

Basically, I think there are some considerations to find the right path.. ones I thought of are..

1. are we doing this because we think users can't `git checkout <release-version>`? If so, we have limited options..
  * keep as-is, but unhook the parent pom so the old versions can keep coherent deps
  * attachment on a release with a zip or tar-gz they can click and get
3. regardless, how do we communicate samples?
  * link in release notes to that version's sample
  * links in the version matrix
  * attachment on a release

what do you think?

codefromthecrypt:
(My proposal would be to keep only one sample and make sure release notes link to it by its version tag. That's the simplest, and we can edit the past releases to add this, too)

codefromthecrypt:
if we go with my suggestion, the github action has support for a default body, so I think we can make it happen automatically (text with links to where the samples are by the git tag of that release) https://github.com/ncipollo/release-action

PR #3259, opened 2024-03-28
Update protocol buffers and protocol buffer dependency.

brendandburns:
cc @codefromthecrypt


codefromthecrypt:
I'm on a spike on something I have to finish today. If this can wait until the weekend, I can sort it out no problem!

brendandburns:
@codefromthecrypt no worries, I actually have time today, I expect I'll get it straightend out. Looks like a dependency issue.

codefromthecrypt:
@brendandburns yep and this will fix (or is at least a workaround to) #3140. I think what's left is to copy plugin configuration that is currently required in our parent pom.xml into the examples whose parents didn't include it (e.g. spotless)

codefromthecrypt (approved):
LGTM

PR #3265, opened 2024-03-30
Fix NetworkingV1Api.java Patch

roshandash411:
The patchingress is broken in 1.20 version. So raised a fix for it. Even the open api spec mentions about using
            "application/json-patch+json", "application/merge-patch+json", "application/strategic-merge-patch+json", "application/apply-patch+yaml"


yue9944882:
close and re-open to trigger CI job

brendandburns:
Closing this PR, because the right solution here is to use `PatchUtils.java`

Issue #2438, opened 2022-11-03
Java Client doesn't refresh client auth exec-based credentials


micahhausler:
**Describe the bug**

Kubernetes supports [executing an external process](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#configuration) to get API server credentials, and the Java K8s client does too. 

For processes that run longer than a credential's lifetime, client-go [consumes the `.Status.ExpirationTimestamp`](https://github.com/kubernetes/kubernetes/blob/6824cd3addbb214e88e00f451babb8a44eb738d1/staging/src/k8s.io/client-go/plugin/pkg/client/auth/exec/exec.go#L472), caches the credential and will refresh when the credential expires. 

Today, the [Java client ignores the `.Status.ExpirationTimestamp`](https://github.com/kubernetes-client/java/blob/v16.0.1/util/src/main/java/io/kubernetes/client/util/KubeConfig.java#L310) and only gets the exec-based credential once at process start, and will not refresh.  

Amazon EKS's tokens (using [kubernetes-sigs/aws-iam-authenticator](https://github.com/kubernetes-sigs/aws-iam-authenticator)) use this exec-based format, and return an expiration timestamp [15 minuets in the future](https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/6da4c44ca4aa1c83dbc0820c5413312606dc09f3/pkg/token/token.go#L361) (same for the [AWS CLI implementation](https://github.com/aws/aws-cli/blob/2.8.8/awscli/customizations/eks/get_token.py#L140)). This causes  long-running Java clients (such as a Spark job) to fail authentication after 15 minuets. 

**Client Version**

All versions

**Kubernetes Version**

All versions

**Java Version**
N/A

**To Reproduce**
N/A

**Expected behavior**

The Java client should cache the returned exec-based credential, and refresh automatically

**KubeConfig**
N/A

**Server (please complete the following information):**
 - OS: [e.g. Linux]
 - Environment [e.g. container]
 - Cloud [e.g. Azure]

**Additional context**
N/A

brendandburns:
We'd be happy to take a PR to implement this, you will need to implement the `Authenticator` interface and register it for `exec` providers.

See:
https://github.com/kubernetes-client/java/blob/master/util/src/main/java/io/kubernetes/client/util/authenticators/AzureActiveDirectoryAuthenticator.java

For an example.

brendandburns:
[...]

PR #3406, opened 2024-05-16
fix: AuthenticationRefresher wrapper solution

raykrueger:
This fixes/relates to #2438 and #290

This commit introduces the AuthenticationRefresher as an implementation of Authentication. It's purpose is to warp another Authentication instance and refresh it very n seconds by calling `provide` on the wrapped Authentication object.

This code is intedned to be used a solution to issue #2438.

Any Authentication object that provides credentials, that must be refreshed, to the ApiClient do not have a way to do so.

There are two ways to use the AuthenticationRefresher with ClientBuilder. The first option is to use the
ClientBuilder.setAuthentication(...) method. An Authentication instance can be wrapped in an AuthenticationRefresher and passed into the setAuthentication method. For example,

```
ClientBuilder.standard().setAuthentication(
  //wrap an auth and refresh it every 15 minutes (900s)
  new Authentication(someDelegateAuthenticationInstance, 900)
);
```

This is integrated up into the ClientBuilder.build() method by adding a new authenticationRefreshSeconds field and getter/setter pair. If a refresh interval is set, the Authentication object used by ClientBuilder will be wrapped with an AuthenticationRefresher.

### DIFF
--- a/util/src/main/java/io/kubernetes/client/util/credentials/AuthenticationRefresher.java
+++ b/util/src/main/java/io/kubernetes/client/util/credentials/AuthenticationRefresher.java
@@ -15,6 +15,7 @@ package io.kubernetes.client.util.credentials;
 import java.lang.ref.WeakReference;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -41,12 +42,13 @@ public class AuthenticationRefresher implements Authentication {
     private final Authentication delegateAuthentication;
     private final Long expirationSeconds;
 
-    private final ScheduledExecutorService executor;
+    private final ScheduledExecutorService executor = Executors.newSingleThreadScheduledExecutor();
+
+    private ScheduledFuture<?> currentSchedule;
 
     public AuthenticationRefresher(Authentication delegateAuthentication, long expirationSeconds) {
         this.delegateAuthentication = delegateAuthentication;
         this.expirationSeconds = expirationSeconds;
-        this.executor = Executors.newSingleThreadScheduledExecutor();
         log.debug("AuthenticationRefresher initialized with expirationSeconds: " + expirationSeconds);
     }
 
@@ -59,13 +61,19 @@ public class AuthenticationRefresher implements Authentication {
         log.debug("Invoking authentication");
         this.delegateAuthentication.provide(client);
         // Schedule the next run.
-        scheduleRefresh(client);
+        synchronized (this) {
+            if (currentSchedule != null) {
+                stop();
+            }
+            this.currentSchedule = scheduleRefresh(client);
+        }
+
     }
 
-    private void scheduleRefresh(final ApiClient clientArg) {
+    private ScheduledFuture<?> scheduleRefresh(final ApiClient clientArg) {
         WeakReference<ApiClient> clientReference = new WeakReference<>(clientArg);
 
-        executor.scheduleAtFixedRate(new Runnable() {
+        return executor.scheduleAtFixedRate(new Runnable() {
             @Override
             public void run() {
                 log.debug("Refreshing authentication");
@@ -87,10 +95,14 @@ public class AuthenticationRefresher implements Authentication {
     }
 
     public boolean isRunning() {
-        return !executor.isShutdown();
+        return !this.currentSchedule.isCancelled();
     }
 
     public void stop() {
-        executor.shutdownNow();
+        this.currentSchedule.cancel(true);
+    }
+
+    public Authentication getDelegateAuthentication() {
+        return delegateAuthentication;
     }
 }
--- a/util/src/main/java/io/kubernetes/client/util/credentials/AuthenticationRefresher.java
+++ b/util/src/main/java/io/kubernetes/client/util/credentials/AuthenticationRefresher.java
@@ -1,0 +1,66 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package io.kubernetes.client.util.credentials;
+
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.kubernetes.client.openapi.ApiClient;
+
+/**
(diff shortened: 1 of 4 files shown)


util/src/main/java/io/kubernetes/client/util/credentials/AuthenticationRefresher.java
```diff
@@ -0,0 +1,66 @@
...
+        this.expirationSeconds = expirationSeconds;
+        log.debug("AuthenticationRefresher initialized with expirationSeconds: " + expirationSeconds);
+    }
+
+    /**
+     * Calls delegateAuthentication every expirationSeconds
+     */
+    @Override
+    public void provide(ApiClient client) {
+        ScheduledExecutorService executor = Executors.newSingleThreadScheduledExecutor();
```
brendandburns, line 53: This should be moved out into the constructor (or a static singleton)

util/src/test/java/io/kubernetes/client/util/credentials/AuthenticationRefresherTest.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @Test
+    void credentialProviderTimerInvoked() throws Exception {
+        final Authentication provider = mock(Authentication.class);
+        final Authentication wrapper = new AuthenticationRefresher(provider, 2);
+
+        final ApiClient client = ClientBuilder.standard().setAuthentication(wrapper).build();
+
+        Thread.sleep(3000);
```
brendandburns, line 45: Tests with timing are inherently flaky. Please use synchronization primitives.

dims:
@raykrueger is there a way/need to be able to cancel the task running in the executor?

raykrueger:
> @raykrueger is there a way/need to be able to cancel the task running in the executor?

Yep! I'm combining Brendan's concern with some WeakReference foo to shut the timerdown when the APIClient gets GC'd. Commit incoming!

raykrueger:
I started the code review thing on accident and don't know what to do about that heh.

raykrueger:
(Sorry for spamming, wanted a new line in the previous message). Please have a look at the latest. I've added some safety around time timer and moved it to a field. The timer is now cancelled and nulled out if the APICilent is garbage collected. Additionally I removed the flaky Thread.sleep timing from the tests in favor of Countdown Latches.

util/src/test/java/io/kubernetes/client/util/credentials/AuthenticationRefresherTest.java
```diff
@@ -0,0 +1,98 @@
...
+            @Override
+            public void provide(ApiClient client) {
+                latch.countDown();
+            }
+        };
+        final Authentication wrapper = new AuthenticationRefresher(provider, 2);
+
+        ClientBuilder.standard().setAuthentication(wrapper).build();
+
+        boolean result = latch.await(4, TimeUnit.SECONDS);
```
brendandburns, line 62: Don't have a timeout here. The overall test already has a timeout, so it's better to have the test timeout, vs having the await take 4 seconds and cause the test to flake.

I realize it seems unlikely that this could take 4 seconds, but on overloaded test infrastructure machines, weird scheduling things happen.

util/src/test/java/io/kubernetes/client/util/credentials/AuthenticationRefresherTest.java
```diff
@@ -0,0 +1,98 @@
...
+
+        //Crazy hack to force garbage collection and wait for it
+        Object obj = new Object();
+        WeakReference<Object> ref = new WeakReference<>(obj);
+        obj = null;
+        while (ref.get() != null) {
+            System.gc();
+        }
+
+        boolean result = latch.await(4, TimeUnit.SECONDS);
```
brendandburns, line 93: Same comment wrt. timeouts.

util/src/main/java/io/kubernetes/client/util/credentials/AuthenticationRefresher.java
```diff
@@ -0,0 +1,96 @@
...
+        log.debug("AuthenticationRefresher initialized with expirationSeconds: " + expirationSeconds);
+    }
+
+    /**
+     * Calls delegateAuthentication every expirationSeconds
+     */
+    @Override
+    public void provide(ApiClient client) {
+
+        synchronized (this) {
```
brendandburns, line 61: Why do this lazily on the first call to provide vs. doing it in the contructor? Do you expect that the ApiClient will get garbage collected repeatedly but this authentication will stick around?

brendandburns:
Few more comments. Thanks.

raykrueger:
Ok, I've simplified things a bit. Let me know if this makes better sense. Thanks!

util/src/main/java/io/kubernetes/client/util/credentials/AuthenticationRefresher.java
```diff
@@ -0,0 +1,96 @@
...
+    /**
+     * Calls delegateAuthentication every expirationSeconds
+     */
+    @Override
+    public void provide(ApiClient client) {
+        // Run it now, synchronously.
+        log.debug("Invoking authentication");
+        this.delegateAuthentication.provide(client);
+        // Schedule the next run.
+        scheduleRefresh(client);
```
brendandburns, line 62: Isn't this going to schedule a refresh every time the credential is provided? I think you want to only schedule it once.

util/src/test/java/io/kubernetes/client/util/credentials/AuthenticationRefresherTest.java
```diff
@@ -0,0 +1,87 @@
...
+        Authentication mockProvider = mock(Authentication.class);
+        AuthenticationRefresher wrapper = new AuthenticationRefresher(mockProvider, 2);
+
+        ClientBuilder.standard().setAuthentication(wrapper).build();
+        wrapper.stop();
+
+        assertFalse(wrapper.isRunning(), "After our unused ApiClient returned from .build() is garbage collected, the refresher should stop.");
+        verify(mockProvider, times(1)).provide(any(ApiClient.class));
+    }
+
```
brendandburns, line 86: If my comment above is correct and there is a bug, add a test to validate the bug and then ensure it stays fixed.

raykrueger:
I pushed some code a few days ago, but didn’t reach out here. Back to you sir!

util/src/main/java/io/kubernetes/client/util/credentials/AuthenticationRefresher.java
brendandburns: Still wondering about this. Why not just create the executor once in the constructor and be done with it? This current logic seems convulted and I'm not sure I see the value of the lazy evaluation.

brendandburns:
Sorry, I feel like I'm dragging this code review on and on, but I think this code could be simpler. Added a comment.

bryantbiggs:
any update on this - or is [this](https://github.com/kubernetes-client/java/pull/3406#discussion_r1616436450) comment still waiting to be resolved?

brendandburns:
@bryantbiggs still waiting afaik.

brendandburns:
btw, I dug into this deeper, this won't work correctly because when we construct the `ApiClient` the token is set once and it is never revisited.

Making token refresh work will require changes in the upstream code generator, or a patch to ApiClient.

Closing this for now since I think we will need to start with a fresh solution.

brendandburns:
See:
https://github.com/OpenAPITools/openapi-generator/pull/6036

PR #3471, opened 2024-06-09
Bump root pom to 21.0.0-SNAPSHOT

yue9944882:
To prepare for cutting 1.30 release, bumping root pom version from 20.0.0 to 21.0.0.

Next step: will cut a new release branch `release-21` and then run the release job.


brendandburns:
fwiw, I don't think this is required, I think the release github action is capable of moving between major versions w/o this.

Issue #3510, opened 2024-06-25
Memory leak in Cache indices when item is deleted


wuxudong:
**Describe the bug**
```
private void deleteFromIndices(ApiType oldObj, String key) {
    for (Map.Entry<String, Function<ApiType, List<String>>> indexEntry : this.indexers.entrySet()) {
      Function<ApiType, List<String>> indexFunc = indexEntry.getValue();
      List<String> indexValues = indexFunc.apply(oldObj);
      if (CollectionUtils.isEmpty(indexValues)) {
        continue;
      }

      Map<String, Set<String>> index = this.indices.get(indexEntry.getKey());
      if (index == null) {
        continue;
      }
      for (String indexValue : indexValues) {
        Set<String> indexSet = index.get(indexValue);
        if (indexSet != null) {
          indexSet.remove(key);
        }
      }
    }
  }

```

indexSet should also be deleted if it is empty; otherwise, a lot of untouchable HashSets will cause memory leaks. A proper fix would be:

```
private void deleteFromIndices(ApiType oldObj, String key) {
    for (Map.Entry<String, Function<ApiType, List<String>>> indexEntry : this.indexers.entrySet()) {
      Function<ApiType, List<String>> indexFunc = indexEntry.getValue();
      List<String> indexValues = indexFunc.apply(oldObj);
      if (CollectionUtils.isEmpty(indexValues)) {
        continue;
      }

      Map<String, Set<String>> index = this.indices.get(indexEntry.getKey());
      if (index == null) {
        continue;
      }
      for (String indexValue : indexValues) {
        Set<String> indexSet = index.get(indexValue);
        if (indexSet != null) {
          indexSet.remove(key);
          if (indexSet.isEmpty()) {
              index.remove(indexValue);
          }
        }
      }
    }
  }
```

**Client Version**
`15.0.1`

**Kubernetes Version**
`1.19.3`

**Java Version**
Java 8

**To Reproduce**

Memory grows higher after items are created and removed.

**Expected behavior**

no memory leak after item is removed.

**KubeConfig**
If applicable, add a KubeConfig file with secrets redacted.

**Server (please complete the following information):**
 - OS: [e.g. Linux]
 - Environment [e.g. container]
 - Cloud [e.g. Azure]

**Additional context**

wuxudong:
Compare to client-go, in thread_safe_store.go

```
func (i *storeIndex) deleteKeyFromIndex(key, indexValue string, index Index) {
	set := index[indexValue]
[...]

PR #3513, opened 2024-06-25
Fixed memory leak in cache indices when item is deleted

gdevakumar:
**Fix to the issue**:  #3510 

Memory leak in Cache indices when item is deleted #3510


njucjc:
@gdevakumar @brendandburns Is this fix in latest release yet?

njucjc:
> @gdevakumar @brendandburns @yue9944882 Is this fix in latest release yet?

I use `23.0.0-legacy`, but still encounter this issue.

PR #3534, opened 2024-07-03
Update pom.xml - com.bucket4j migration

danielburrell:
The groupId `com.github.vladimir-bukhtoyarov` has migrated to `com.bucket4j` see https://mvnrepository.com/artifact/com.github.vladimir-bukhtoyarov/bucket4j-core/8.0.1

Note:
I have not upgraded the version (there are newer versions). This is to allow this to be a like-for-like replacement.

Once this migration is made, dependabot will be able to see the newer versions and propose future upgrades.

### DIFF
--- a/extended/pom.xml
+++ b/extended/pom.xml
@@ -39,7 +39,7 @@
             <artifactId>commons-lang3</artifactId>
         </dependency>
         <dependency>
-            <groupId>com.github.vladimir-bukhtoyarov</groupId>
+            <groupId>com.bucket4j</groupId>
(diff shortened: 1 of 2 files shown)


danielburrell:
@brendandburns / @yue9944882 do you have an ETA for when this can be reviewed?

thank you in advance

danielburrell:
Sorry, I missed, com.github.vladimir-bukhtoyarov in the extended pom (didn't realise that existed). Will correct and revert shortly.

danielburrell:
@brendandburns my apologies for the misfire, can we retry?

Issue #3489, opened 2024-06-17
Latest release 20.0.1 doesn't contain "client-java-contrib" module that contains V1 cert-manager models.


10dds:
**Describe the bug**
Latest client-java release 20.0.1 doesn't contain client-java-contrib module that contains V1 cert-manager models.
and recent cert-manager releases (>= v1.6) doesn't support v1alpha2, v1alpha3, and v1beta1.

**Client Version**
20.0.1

**Kubernetes Version**
N/A

**Java Version**
Java 8

**To Reproduce**
Upgrade cert-manager to >= v1.6 version.

**Expected behavior**
Kubernetes client-java library should support V1 cert-manager models.
https://cert-manager.io/docs/releases/release-notes/release-notes-1.6/.

**KubeConfig**
If applicable, add a KubeConfig file with secrets redacted.

**Server (please complete the following information):**
 - OS: [e.g. Linux]
 - Environment [e.g. container]
 - Cloud [e.g. Azure]

**Additional context**

brendandburns:
I think the code needs to be regenerated following the instructions here:
https://github.com/kubernetes-client/java/tree/master/client-java-contrib/cert-manager#code-generation

We'd be happy to take a PR with the regenerated code. Thanks!

vaidikcode:
i tried to follow the steps and run the update sh command but access is denied what can i do about that?

vaidikcode:
@brendandburns i tried to follow the steps and run the update sh command but access is denied what can i do about that?

adithyanms-debug:
Hey i am new to open source contribution is it a good issue to work on

PR #3548, opened 2024-07-09
Regenerate code of client-java-contrib cert-manager

THUzxj:
As discussed in #3489, I followed the instructions to regenerate the code of client-java-contrib cert-manager. It seems that nothing changed except timestamps and version numbers.
I am not authorized the docker image on github so I built an image locally.


brendandburns:
This PR looks fine, but I don't think that it is going to fix the issue you raised in #3489 it's not adding any new models.

brendandburns:
Closing this PR as it only updated timestamps, no substantial code changes.

PR #3563, opened 2024-07-12
Migrate deprecated Bucket4J class usage to Bucket class to unblock upgrade

danielburrell:
Migrate `@deprecated` `Bucket4J` class usage to `Bucket` class to unblock dependabot upgrade of bucket4j from `8.0.1` to `8.10.1`. 

I include the upgrade of the Bucket4J library as proposed by dependabot to demonstrate that it all works together.
Tested locally.


brendandburns (approved):
/lgtm
/approve

Issue #3463, opened 2024-06-04
Kubectl should support cascading deletes for Jobs


gautam-rl:
**Describe the bug**
Kubectl doesn't make it possible to delete a Job and have the delete cascade to the underlying pod. By default it appears to orphan the pod as described here: https://kubernetes.io/docs/concepts/workloads/controllers/job/#ttl-mechanism-for-finished-jobs.

**Client Version**
20.0.1

**Kubernetes Version**
1.29

**Java Version**
Java 21

**To Reproduce**
Create a job and then delete it using Kubectl:
```
Kubectl.delete(V1Job.class)
    .apiClient(client)
    .name(jobId)
    .namespace(namespace)
    .ignoreNotFound(true)
    .execute());
```

**Expected behavior**
It should be possible to pass in `.deleteOptions()` and cascade or do foreground propagation.

**Additional context**
As a workaround we can use the batchApi:
```
V1Status v1Status =
          batchApi.deleteNamespacedJob(jobId, namespace).propagationPolicy("Foreground").execute();
```

brendandburns:
Feel free to send a PR to support this.

PR #3572, opened 2024-07-16
Added support for passing delete options when using kubectlDelete.#3463

callingmahendra:
closes: #3463 

Added an option to pass deleteOptions in kubectlDelete.

### DIFF
--- a/extended/src/test/java/io/kubernetes/client/extended/kubectl/KubectlDeleteTest.java
+++ b/extended/src/test/java/io/kubernetes/client/extended/kubectl/KubectlDeleteTest.java
@@ -213,7 +213,10 @@ class KubectlDeleteTest {
         deleteOptions.setPropagationPolicy("Foreground");
         kubectlDelete.deleteOptions(deleteOptions);
         kubectlDelete.execute();
-
+        apiServer.verify(
+            1,
+            deleteRequestedFor(urlPathEqualTo("/apis/batch%2Fv1/batch%2Fv1/namespaces/foo/jobs/bar"))
+                .withRequestBody(equalToJson("{\"propagationPolicy\" : \"Foreground\"}")));
         assertThatThrownBy(() -> {
             KubectlDelete<V1Job> kubectlDelete2 = Kubectl.delete(V1Job.class);
             kubectlDelete2.apiClient(apiClient);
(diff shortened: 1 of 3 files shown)


extended/src/test/java/io/kubernetes/client/extended/kubectl/KubectlDeleteTest.java
```diff
@@ -207,6 +209,9 @@ void kubectlDelete() throws KubectlException, ApiException {
         KubectlDelete<V1Job> kubectlDelete = Kubectl.delete(V1Job.class);
         kubectlDelete.apiClient(apiClient);
         kubectlDelete.namespace("foo").name("bar");
+        DeleteOptions deleteOptions = new DeleteOptions();
+        deleteOptions.setPropagationPolicy("Foreground");
+        kubectlDelete.deleteOptions(deleteOptions);
```
brendandburns, line 214: Can you add validation that this worked correctly to the URL validation below

brendandburns:
Thanks for the PR. I think you need to run `mvn spotless:apply` to perform formatting for style, and expand the unit test slightly.

extended/src/test/java/io/kubernetes/client/extended/kubectl/KubectlDeleteTest.java
callingmahendra: Added check below for URL validation and you were correct it was not passing deleteOptions, hence GenericKubernetesApi.java to pass it.

util/src/main/java/io/kubernetes/client/util/generic/GenericKubernetesApi.java
```diff
@@ -1159,6 +1159,7 @@ private CallBuilder makeNamespacedDeleteCallBuilder(
             namespace,
             this.resourcePlural,
             name)
+                .body(deleteOptions)
                 // TODO: fill/convert the option
```
callingmahendra, line 1163: Should we remove this TODO as this is no more valid?

callingmahendra:
> Thanks for the PR. I think you need to run `mvn spotless:apply` to perform formatting for style, and expand the unit test slightly.

Done both the changes, please review

util/src/main/java/io/kubernetes/client/util/generic/GenericKubernetesApi.java
brendandburns: Yes, please delete this now.

brendandburns:
Thanks for adding the test, and fixing the code.

I think it still needs formatting, and that TODO removed, but then I think it's good to merge.

callingmahendra:
> Thanks for adding the test, and fixing the code.
> 
> I think it still needs formatting, and that TODO removed, but then I think it's good to merge.

Removed the TODO, and also rerun mvn spotless:apply to make sure formatting apply.

Issue #3590, opened 2024-07-24
Add support for FIPS Bouncy Castle library 


robhafner:
**Describe the bug**

We are attempting to use version 19.0.1 of the Kubernetes Java Client from a Spring Boot 2.7 application to invoke the Kubernetes API server to validate a service account token. Our Spring Boot application is configured with the FIPS enabled version 1.0.2.4 of the Bouncy Castle library (and does not include the non FIPS version of the Bouncy Castle library).  The call to the API with a valid token results in the following status. 

class V1TokenReviewStatus {
    audiences: null
    authenticated: null
    error: [invalid bearer token, service account token has been invalidated]
    user: class V1UserInfo {
        extra: null
        groups: null
        uid: null
        username: null
    }
} 

Stepping through the debugger the token appears to be consider invalid as a result of the non FIPS enabled bouncy castle class not being available which is defined in src/main/java/io/kubernetes/client/SSLUtils.java 

`
static {
    Security.addProvider(new org.bouncycastle.jce.provider.BouncyCastleProvider());
}
`

Updating the SSLUtils.java with the following changes allows the token to be verified successfully. 

`
  static –{
    Provider provider;
    try {
      Class clazz  = getProvider();
        provider = (Provider) clazz.getDeclaredConstructor(null).newInstance();
    } catch (Exception e) {
        throw new RuntimeException(e);
    }

    Security.addProvider(provider);
  }

  public static Class getProvider() throws ClassNotFoundException {
    Class clazz;
    try {
      clazz = Class.forName("org.bouncycastle.jcajce.provider.BouncyCastleFipsProvider");
    } catch(ClassNotFoundException cnf) {
      clazz = Class.forName("org.bouncycastle.jce.provider.BouncyCastleProvider");
    }
    return clazz;
  }
`

 I'd be happy to put a PR together with this change to resolve this issue. However, I ran across another issue which seemed very similar which was closed without a fix. 
 
 https://github.com/kubernetes-client/java/issues/2086
 
Can you confirm if the Kubernetes Java Client project is willing to support the FIPS version of Bouncy Castle?  If not, we will likely be forced to fork the library to meet our needs. 

**Client Version**
1.29.2

**Kubernetes Version**
1.28.2

**Java Version**
Java 17

**To Reproduce**

**Expected behavior**
The call to verify a token works successfully when only the FIPS version of bouncy castle is on the classpath. 

**KubeConfig**
If applicable, add a KubeConfig file with secrets redacted.

[...]

PR #3595, opened 2024-07-25
feat: add support for FIPS bouncy castle #3590

robhafner:
Prefer FIPS version of bouncy castle if available. Otherwise, fallback to non FIPS version.

### DIFF
--- a/util/src/main/java/io/kubernetes/client/util/SSLUtils.java
+++ b/util/src/main/java/io/kubernetes/client/util/SSLUtils.java
@@ -24,6 +24,7 @@ import java.security.KeyStore;
 import java.security.KeyStoreException;
 import java.security.NoSuchAlgorithmException;
 import java.security.PrivateKey;
+import java.security.Provider;
 import java.security.Security;
 import java.security.UnrecoverableKeyException;
 import java.security.cert.Certificate;
@@ -32,6 +33,7 @@ import java.security.cert.CertificateFactory;
 import java.security.cert.X509Certificate;
 import java.security.spec.InvalidKeySpecException;
 import java.util.Collection;
+import java.util.ServiceLoader;
 import javax.net.ssl.KeyManager;
 import javax.net.ssl.KeyManagerFactory;
 import org.bouncycastle.asn1.pkcs.PrivateKeyInfo;
@@ -40,10 +42,18 @@ import org.bouncycastle.openssl.PEMParser;
 import org.bouncycastle.openssl.jcajce.JcaMiscPEMGenerator;
 import org.bouncycastle.openssl.jcajce.JcaPEMKeyConverter;
 import org.bouncycastle.util.io.pem.PemWriter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class SSLUtils {
+  private static final Logger log = LoggerFactory.getLogger(SSLUtils.class);
+
   static {
-    Security.addProvider(new org.bouncycastle.jce.provider.BouncyCastleProvider());
+    ServiceLoader<Provider> services = ServiceLoader.load(java.security.Provider.class);
+    for (Provider service : services) {
+      log.debug("Found security provider: " + service.getName());
+      Security.addProvider(service);
+    }
   }
 
   public static boolean isNotNullOrEmpty(String val) {


brendandburns:
Thanks for the PR. Instead of this approach, I think I would prefer to keep the default and then use some sort of `-D` flag which would specify the security provider to load.

robhafner:
So we have the bc-fips-1.0.2.4.jar on our classpath and it has a META-INF/services/java.security.Provider file with the "org.bouncycastle.jcajce.provider.BouncyCastleFipsProvider" class defined in it. I would expect that file would trigger loading the required fips provider.  

It's a bit unclear to me if there is a security risk with attempting to load the non fips provider class as well (in the event that the non fips provider class ended up on our classpath). We've chosen to only support FIPS compliant algorithms in our software. 

This change is based on what we noticed in another project with respect to how they check the provider class.

https://github.com/puppetlabs/jvm-ssl-utils/blob/main/src/java/com/puppetlabs/ssl_utils/SSLUtils.java#L120

brendandburns:
ok, how about something like 

```java
static {
  String className = System.getProperty("io.kubernetes.SecurityProvider");
  if (className == null) {
     className == <non-fips-provider>;
  }
  Security.Provider = Class.forName(className);
}
```

robhafner:
Thinking about this a bit more,  would the right thing to do in this case be to let the BouncyCastle jars register the provider via the Java Service Provider Interface (SPI)?

For example:
- The bc-fips-1.0.2.4.jar  defines a META-INF/services/java.security.Provider file with a value of org.bouncycastle.jcajce.provider.BouncyCastleFipsProvider.
- The bcprov-jdk18on-1.76.jar defines a  META-INF/services/java.security.Provider file with the following values:

org.bouncycastle.jce.provider.BouncyCastleProvider
org.bouncycastle.pqc.jcajce.provider.BouncyCastlePQCProvider


Note, the bcprov-jdk18on-1.76.jar is pulled in when adding the following dependency in gradle.

    implementation("io.kubernetes:client-java:19.0.1")

We currently exclude the non-fips bouncy castle jars in our build.gradle file like this:

    implementation("io.kubernetes:client-java:19.0.1") {
        exclude group: 'org.bouncycastle', module: 'bcpkix-jdk18on'
        exclude group: 'org.bouncycastle', module: 'bcprov-jdk18on'
        exclude group: 'org.bouncycastle', module: 'bcutil-jdk18on
    }

And add the following dependencies to include the fips version.

    // bouncycastle FIPS jars
    implementation "org.bouncycastle:bc-fips"
    implementation "org.bouncycastle:bcpkix-fips"


Given this, I think a better fix might be to remove the static block of code initializing the provider in the Kubernetes Java client and let the bouncy castles jars perform the registration via Java SPI. This would allow a user of the Kubernetes Java Client to pick the bouncy castle provider they wish to use based on the jars that they include in their project.





From: Brendan Burns ***@***.***>
Date: Thursday, July 25, 2024 at 5:17 PM
To: kubernetes-client/java ***@***.***>
Cc: Robert Hafner ***@***.***>, Mention ***@***.***>
Subject: Re: [kubernetes-client/java] feat: add support for FIPS bouncy castle #3590 (PR #3595)

ok, how about something like

static {

  String className = System.getProperty("io.kubernetes.SecurityProvider");

  if (className == null) {

     className == <non-fips-provider>;

  }

  Security.Provider = Class.forName(className);

}

robhafner:
Forgot to mention, the kubernetes java client may need to invoke the ServiceLoader to load the provider classes when removing the static code block.

ServiceLoader<Provider> services = ServiceLoader.load(java.security.Provider.class);
for (Provider service : services) {
            logger.info("Found: " + service.getName());
}

Based on the documentation I see here:

https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html


From: Robert Hafner ***@***.***>
Date: Wednesday, July 31, 2024 at 9:01 AM
To: kubernetes-client/java ***@***.***>, kubernetes-client/java ***@***.***>
Cc: Mention ***@***.***>
Subject: Re: [kubernetes-client/java] feat: add support for FIPS bouncy castle #3590 (PR #3595)
Thinking about this a bit more,  would the right thing to do in this case be to let the BouncyCastle jars register the provider via the Java Service Provider Interface (SPI)?

For example:
- The bc-fips-1.0.2.4.jar  defines a META-INF/services/java.security.Provider file with a value of org.bouncycastle.jcajce.provider.BouncyCastleFipsProvider.
- The bcprov-jdk18on-1.76.jar defines a  META-INF/services/java.security.Provider file with the following values:
org.bouncycastle.jce.provider.BouncyCastleProvider
org.bouncycastle.pqc.jcajce.provider.BouncyCastlePQCProvider

Note, the bcprov-jdk18on-1.76.jar is pulled in when adding the following dependency in gradle.

    implementation("io.kubernetes:client-java:19.0.1")

We currently exclude the non-fips bouncy castle jars in our build.gradle file like this:

    implementation("io.kubernetes:client-java:19.0.1") {
        exclude group: 'org.bouncycastle', module: 'bcpkix-jdk18on'
        exclude group: 'org.bouncycastle', module: 'bcprov-jdk18on'
        exclude group: 'org.bouncycastle', module: 'bcutil-jdk18on
    }

And add the following dependencies to include the fips version.

    // bouncycastle FIPS jars
    implementation "org.bouncycastle:bc-fips"
    implementation "org.bouncycastle:bcpkix-fips"


Given this, I think a better fix might be to remove the static block of code initializing the provider in the Kubernetes Java client and let the bouncy castles jars perform the registration via Java SPI. This would allow a user of the Kubernetes Java Client to pick the bouncy castle provider they wish to use based on the jars that they include in their project.





From: Brendan Burns ***@***.***>
Date: Thursday, July 25, 2024 at 5:17 PM
To: kubernetes-client/java ***@***.***>
Cc: Robert Hafner ***@***.***>, Mention ***@***.***>
Subject: Re: [kubernetes-client/java] feat: add support for FIPS bouncy castle #3590 (PR #3595)

ok, how about something like

static {

  String className = System.getProperty("io.kubernetes.SecurityProvider");

  if (className == null) {

     className == <non-fips-provider>;

  }

  Security.Provider = Class.forName(className);

}

brendandburns:
I think we need to make sure that this works without anyone doing anything special. I'm worried that if we don't explicitly load the provider that it won't be as easy for the users to use the client.

I want it to be feasible for FIPS users to change their providers, but I don't want to make it harder for the average user.

robhafner:
I agree with you with respect to it being easy and the user shouldn’t have to anything special.

I found this doc which suggests a way to dynamically load the providers using Java SPI.

https://docs.oracle.com/en/java/javase/11/security/howtoimplaprovider.html#GUID-FB9C6DB2-DE9A-4EFE-89B4-C2C168C5982D

Relevant section from the doc below:

Alternatively, you can register providers dynamically. To do so, a program (such as your test program, to be written in Step 9: Write and Compile Your Test Programs<https://docs.oracle.com/en/java/javase/11/security/howtoimplaprovider.html#GUID-C6054169-FE6E-4837-B2BD-382DFEB955C0>) call either the addProvider or insertProviderAt method in the Securityclass:

ServiceLoader<Provider> sl = ServiceLoader.load(java.security.Provider.class);
for (Provider p : sl) {
    System.out.println(p);
    if (p.getName().equals("MyProvider")) {
        Security.addProvider(p);
    }
}


I’ll try updating SSLUtils.java in the following manner. Assuming it does, I’ll update my PR.

public class SSLUtils {
  private static final Logger log = LoggerFactory.getLogger(SSLUtils.class);

  static {
    ServiceLoader<Provider> services = ServiceLoader.load(java.security.Provider.class);
    for (Provider service : services) {
      log.info("Found security provider: " + service.getName());
      Security.addProvider(service);
    }
  }


From: Brendan Burns ***@***.***>
Date: Wednesday, July 31, 2024 at 2:55 PM
To: kubernetes-client/java ***@***.***>
Cc: Robert Hafner ***@***.***>, Mention ***@***.***>
Subject: Re: [kubernetes-client/java] feat: add support for FIPS bouncy castle #3590 (PR #3595)

I think we need to make sure that this works without anyone doing anything special. I'm worried that if we don't explicitly load the provider that it won't be as easy for the users to use the client.

I want it to be feasible for FIPS users to change their providers, but I don't want to make it harder for the average user.

robhafner:
I’ve updated the PR with the logic described in my last message. This logic works fine in my test environment. Assuming it provides the simplicity and flexibility that you are looking for I think we can move forward with it.



From: Robert Hafner ***@***.***>
Date: Wednesday, July 31, 2024 at 4:00 PM
To: kubernetes-client/java ***@***.***>, kubernetes-client/java ***@***.***>
Cc: Mention ***@***.***>
Subject: Re: [kubernetes-client/java] feat: add support for FIPS bouncy castle #3590 (PR #3595)
I agree with you with respect to it being easy and the user shouldn’t have to anything special.

I found this doc which suggests a way to dynamically load the providers using Java SPI.
https://docs.oracle.com/en/java/javase/11/security/howtoimplaprovider.html#GUID-FB9C6DB2-DE9A-4EFE-89B4-C2C168C5982D

Relevant section from the doc below:
Alternatively, you can register providers dynamically. To do so, a program (such as your test program, to be written in Step 9: Write and Compile Your Test Programs<https://docs.oracle.com/en/java/javase/11/security/howtoimplaprovider.html#GUID-C6054169-FE6E-4837-B2BD-382DFEB955C0>) call either the addProvider or insertProviderAt method in the Securityclass:

ServiceLoader<Provider> sl = ServiceLoader.load(java.security.Provider.class);
for (Provider p : sl) {
    System.out.println(p);
    if (p.getName().equals("MyProvider")) {
        Security.addProvider(p);
    }
}


I’ll try updating SSLUtils.java in the following manner. Assuming it does, I’ll update my PR.

public class SSLUtils {
  private static final Logger log = LoggerFactory.getLogger(SSLUtils.class);

  static {
    ServiceLoader<Provider> services = ServiceLoader.load(java.security.Provider.class);
    for (Provider service : services) {
      log.info("Found security provider: " + service.getName());
      Security.addProvider(service);
    }
  }


From: Brendan Burns ***@***.***>
Date: Wednesday, July 31, 2024 at 2:55 PM
To: kubernetes-client/java ***@***.***>
Cc: Robert Hafner ***@***.***>, Mention ***@***.***>
Subject: Re: [kubernetes-client/java] feat: add support for FIPS bouncy castle #3590 (PR #3595)

I think we need to make sure that this works without anyone doing anything special. I'm worried that if we don't explicitly load the provider that it won't be as easy for the users to use the client.

I want it to be feasible for FIPS users to change their providers, but I don't want to make it harder for the average user.

brendandburns:
Thanks for digging in here. Assuming tests pass, this looks good to me.

jwojcie:
hi, two questions if I may
1. When this fix is going to be available in maven repo? I understand it is not there yet in 21.0.2: https://github.com/kubernetes-client/java/blob/v21.0.2/util/src/main/java/io/kubernetes/client/util/SSLUtils.java
2. are you sure this fix you made is going to work? I tested it locally on my FIPS hardened Java, which means that apart from other stuff I have this in java.security:
`security.provider.1=org.bouncycastle.jcajce.provider.BouncyCastleFipsProvider C:DEFRND[SHA256];ENABLE{ALL};
security.provider.2=org.bouncycastle.jsse.provider.BouncyCastleJsseProvider fips:BCFIPS
security.provider.3=SUN`
I use this small code snippet:

`public static void listProviders() {

        System.out.println("Kubernetes fix: +++++++++++++++++++++++++++++");
        ServiceLoader<Provider> services = ServiceLoader.load(java.security.Provider.class);
        for (Provider service : services) {
            System.out.println("Found security provider: " + service.getName());
            //Security.addProvider(service);
        }

        System.out.println("Actual providers: +++++++++++++++++++++++++++++++");
        for (Provider provider : Security.getProviders()) {
            System.out.println("Provider: " + provider.getName() + ", Version: " + provider.getVersion());
        }

        System.out.println("Just for reference: +++++++++++++++++++++++++++++++");
        Provider provider = new org.bouncycastle.jcajce.provider.BouncyCastleFipsProvider();
        System.out.println("Provider: " + provider.getName() + ", Version: " + provider.getVersion());

    }`

And the result is kind of surprising:
`Kubernetes fix: +++++++++++++++++++++++++++++
Found security provider: SunPCSC
Found security provider: SunEC
Found security provider: SunPKCS11
Found security provider: SunMSCAPI
Found security provider: JdkSASL
Found security provider: XMLDSig
Found security provider: SunJGSS
Found security provider: SunSASL
Found security provider: JdkLDAP
Actual providers: +++++++++++++++++++++++++++++++
Provider: BCFIPS, Version: 2.0
Provider: BCJSSE, Version: 2.0019
Provider: SUN, Version: 11.0
Just for reference: +++++++++++++++++++++++++++++++
Provider: BCFIPS, Version: 2.0

Process finished with exit code 0
`

Which makes me think that fix is not going to work, or do I missed something here?

brendandburns:
I think that you need to make sure that those other providers are not in your classpath.

jwojcie:
Can you tell me when this feature is going to be available in the release?

PR #3612, opened 2024-08-07
Update codeql action to remove autobuild

### LINKED ISSUE #3611: CI/CD for main branch is failing because CodeQL version is pinned
We pinned the codeql version to work around a problem with the way that it detects java versions, unfortunately, it looks like this pinned version has been deleted, CI/CD is failing at HEAD we'll need to update the CodeQL version and determine how to deal with the java version detection changes.

Ref https://github.com/github/codeql/issues/13541

brendandburns:
Fixes #3611


brendandburns:
Self merging to unbreak the build.

PR #3642, opened 2024-08-16
Add token refresh for exec credentials.

brendandburns:
This is a replacement for https://github.com/kubernetes-client/java/pull/3406 which had some design flaws due to the complicated way that auth is implemented w/ the code generator.

This follows the same pattern as was previously there for file tokens.

This is a little hacky and should be fixed in the upstream generator, but it works for now at least for tokens.

TODO: figure out if we can do something similar for certificates.


yue9944882 (commented):
/lgtm
/approve

yue9944882:
hmm looks like something is wrong with the cibot

ashirvadgupta:
@brendandburns I have a requirement where user credentials stored in kubeconfig file gets expired in 12 hours and I need to reinitialize ClientApi when that happens. Will this change solve this problem of refresh based on the value set in token refresh?

Issue #3641, opened 2024-08-16
How to set the terminal stty cols and rows through exec API ?


gohonsen:
**Describe the bug**
Hey, I want to enter the k8s pod container to run shell commands through web terminal, but I found that after the web terminal page size changed, I cann't synchronize set terminal size through exec API, and the page display is disordered.
 Is there any way to set the terminal stty cols and rows ? Thanks a lot

**Client Version**
`18.0.1`

**Kubernetes Version**
`v1.28.0`

**Java Version**
Java 8

brendandburns:
This isn't currently supported. You can read resize messages from the stream via `Exec.getResizeStream()` but you can't write to that stream currently.

We need something like the following in the Javascript client:

https://github.com/kubernetes-client/javascript/blob/70f651ca7e5c8d3647c071ab4f6aeb89f4ccae14/src/exec.ts#L66

We'd be happy to take a PR for this support.

brendandburns:
Alternately, you could use `ExecProcess.getInputStream(4).write("{ width: foo, height: bar}")` that should work.

https://github.com/kubernetes-client/java/blob/0e94811dbf8e3a0bfebb0c8a8cc9bf61a28e5e83/util/src/main/java/io/kubernetes/client/Exec.java#L610

gohonsen:
@brendandburns I got it, My code like this:

> execProcess = (Exec.ExecProcess) executionBuilder.execute();
            resizeOutStream = execProcess.getResizeStream();
            outputStream = execProcess.getOutputStream();
            inputStream = execProcess.getInputStream();
            String resize = "{\"Width\":" + cols + ",\"Height\":" + rows + "}";
            resizeOutStream.write(resize.getBytes());
>
and it works,  thanks.

brendandburns:
Glad it's working. Making this a little easier could be a good first issue, so I'll leave this open in case someone wants to take it on.

sriganeshres:
@brendandburns i want to work on this issue. But the problem i cant setup the code properly can you send me the steps to setup the code.

sriganeshres:
![image](https://github.com/user-attachments/assets/f8eedab7-5bd5-4066-8988-5ffcfdec595a)
i have done this till now.

thanks

sriganeshres:
then i did this
![image](https://github.com/user-attachments/assets/a1029ef6-d1f2-4932-b54d-e1bb6c5a7988)
[...]

PR #3657, opened 2024-08-27
Added a fuction/method setTerminalsize in Exec.ExecBuilber

sriganeshres:
I have added a function in Exec.ExecBuilder class which helps to set the Terminal size.

### DIFF
--- a/util/src/main/java/io/kubernetes/client/Exec.java
+++ b/util/src/main/java/io/kubernetes/client/Exec.java
@@ -92,7 +92,8 @@ public class Exec {
   }
 
   /**
-   * Get a {@link Consumer<Throwable>} that will be accepted if there is any unhandled exception
+   * Get a {@link Consumer<Throwable>} that will be accepted if there is any
+   * unhandled exception
    * while the websocket communication is happening.
    *
    * @return The {@link Consumer<Throwable>} that will be used.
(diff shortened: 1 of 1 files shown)


sriganeshres:
related to #3641

util/src/main/java/io/kubernetes/client/Exec.java
```diff
@@ -92,7 +92,8 @@ public void setApiClient(ApiClient apiClient) {
   }
 
   /**
-   * Get a {@link Consumer<Throwable>} that will be accepted if there is any unhandled exception
+   * Get a {@link Consumer<Throwable>} that will be accepted if there is any
```
brendandburns, line 95: Please revert all formatting changes here and below.

brendandburns:
Thanks for starting this. Two notes:
First, please don't include any unrelated formatting changes, it makes it harder to review the PR, you can use `git diff HEAD~1` to visualize any differences in your commit before you send the PR.

Second, this change isn't quite correct, you need to add a utility method to the `Exec` class, not to the builder. The builder is what sets up an `Exec` instance, but the terminal resize can only happen once the exec call starts. If the method is added to the `Exec` class, then someone who starts a call to `Exec` can resize the stream once the connection is established.

sriganeshres:
Sorry @brendandburns for very very late response. But can you guide me more than just that im facing a bit of issue here while trying to maintain that it should be able to be run after execute(). Thanks for the support till now.

Issue #3076, opened 2024-02-14
It's not possible to create a Pod, because of the default overhead value (empty map) in v20.0.0


kvmw:
**Describe the bug**

The default `overhead` value of Pod spec in version 20.0.0 is empty map which is causing the following error:
```java
    java.lang.RuntimeException: io.kubernetes.client.openapi.ApiException: Message: 
    HTTP response code: 403
    HTTP response body: {"kind":"Status","apiVersion":"((redacted))","metadata":{},"status":"Failure","message":"pods \"foo\" is forbidden: pod rejected: Pod Overhead set without corresponding RuntimeClass defined Overhead","reason":"Forbidden","details":{"name":"foo","kind":"pods"},"code":403}
```
As a workaround, one can set the `overhead` explicitly to null:
```java
V1Pod pod = new V1Pod()
                .metadata(new V1ObjectMeta().name("foo"))
                .spec(new V1PodSpec()
                        .overhead(null)  // HERE: setting the overhead value to null, to avoid above error. 
                        .addContainersItem(new V1ContainerBuilder()
                        // omitted 
```

**Client Version**
e.g. `20.0.0`

**Kubernetes Version**
any version

**Java Version**
any version

**To Reproduce**
Create a simple Pod, without specifying anything for `overhead`:
```java
V1Pod pod = new V1Pod()
                .metadata(new V1ObjectMeta().name("foo"))
                .spec(new V1PodSpec()
                        .addContainersItem(new V1ContainerBuilder()
                        .withImage("any-image")
                        .withName("foo-image").build()));

coreV1Api.createNamespacedPod("default", pod).execute();
```
**Expected behavior**
Pod creation should be successful but it fails due to `overhead: {}` being present in the request payload. 

I think, either the default value of `overhead` should be `null` or when serialising the request, the empty values should be omitted. 
**KubeConfig**
N/A

**Server (please complete the following information):**
Any

**Additional context**
N/A

brendandburns:
This is either a bug in the upstream swagger specification, or the way it is translated in the code generator. We'll have to investigate.

brendandburns:
The more I think about this, the more I think that it is a bug in Kubernetes. I feel like empty map and null map should be equivalent.

I'll file an issue on the main kubernetes repo and see what people think.

kvmw:
@brendandburns  agree. this might need to be addressed in k8s.

A simple template like the following, fails with the same error:
```yaml
apiVersion: v1
kind: Pod
[...]

PR #3680, opened 2024-09-09
Regenerate for 1.30, fix patches

brendandburns:
@yue9944882 I regenerated for 1.30 including the fixes for nullable arrays, also fixed up the patches so they applied cleanly.

Ref https://github.com/kubernetes-client/java/issues/3076

~Please take a look.~

~Tests are failing b/c the YAML is being generated with empty lists intead of nulls, I'm not sure this is easily fixable but will investigate.~

~YAML is fixed, but one JSON test is failing and needs a fix.~

This is ready for review.


brendandburns:
Ok, I updated this so that empty lists are omitted from YAML dumping and I _think_ that should be ok.

yue9944882:
```
Error:    KubectlDeleteTest.kubectlDelete:216 No requests exactly matched. Most similar request was:  expected:<
DELETE
/apis/batch%2Fv1/batch%2Fv1/namespaces/foo/jobs/bar

[equalToJson]
{
  "propagationPolicy" : "Foreground"
}> but was:<
DELETE
/apis/batch%2Fv1/batch%2Fv1/namespaces/foo/jobs/bar


{
  "dryRun" : [ ],
  "propagationPolicy" : "Foreground"
}>
```

it looks like k/k api added a new dry-run option for delete API, so we need to align with that. i will append another commit to fix it

brendandburns:
@yue9944882 this is fixed up and ready to review, I tried to keep the commits separate for easier reviewing, but I'll likely squash them in the end to fix the issues with authorship and the PR bot.

I had to remove a check from `KubectlDrain.java` to get tests to pass, but I'm not sure that it really made sense in the first place anyway. Also added an `exclusion strategy` to GSON for `metadata.managedFields`

yue9944882:
> EasyCLA — Missing CLA Authorization.

This PR failed CLA check due to bot commit, let me squash the branch into single commit to workaround

brendandburns:
manual lgtm, since I'm the original author of the PR.

PR #3737, opened 2024-10-12
Shutdown original queue before override

yuzp1996:
When overriding the workQueue for the controller, the original queue must be shut down to avoid thread leakage. If the original queue is not shut down here, it will remain active even after the controller is shut down. This is because the controller's queue has been replaced by the newly created queue.

Related Issue: https://github.com/kubernetes-client/java/issues/3738


yuzp1996:
@brendandburns  I have fixed the format error. Could you please take a look again and give your LGTM or approval? Thank you.

Issue #3336, opened 2024-04-21
how to submit a yaml like kubectl create -f file.yaml ?


xiaoyu1095:
Is there a method to submit YAML resources of any type to Kubernetes ?
What I mean is not having to differentiate between Kind types.

brendandburns:
This is not currently possible. You would need to modify the `Yaml` class here:

https://github.com/kubernetes-client/java/blob/master/util/src/main/java/io/kubernetes/client/util/Yaml.java

To load unstructured YAML and then pick out the API group/version/kind and then use the `ModelMapper` here:
https://github.com/kubernetes-client/java/blob/master/util/src/main/java/io/kubernetes/client/util/ModelMapper.java

To determine the right strongly typed object, and then use `Yaml.load` again to reload the YAML into that strongly typed object.

Then you would need to pass that object to a `GenericApi` handler that was configured with the group/version/kind that you found above.

So it's possible, but it will require you to write a bunch of code. If you do, do that, feel free to send a PR.

Ayush-Thakur-geek:
hey @brendandburns, i had a question.

So, do we need a feature that can differentiate between kinds without explicitly mentioning kinds in the yaml file.

reachvangala:
You can submit any Kubernetes YAML using:  kubectl apply -f your-resource.yaml
If your YAML file contains multiple resources separated by ---, this also works seamlessly:

apiVersion: v1
kind: Namespace
metadata:
  name: my-namespace
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: my-config
  namespace: my-namespace
data:
  key: value

Issue #3668, opened 2024-08-29
CoreV1Event JSON validation fails due to missing reportingController field


restenb:
**Describe the bug**
On docker-desktop with the given k8s version, event JSONs contain the field `reportingController`. This is not included in `CoreV1Event.java`, and the validation fails with the error message

**The field `reportingController` in the JSON string is not defined in the `CoreV1Event` properties.**

Raw JSON returned:
`{\"kind\":\"Event\",\"apiVersion\":\"events.k8s.io/v1\",\"metadata\":{\"name\":\"pod\",\"namespace\":\"pods\",\"uid\":\"312e32e1-0c8e-416f-ad47-a6958432bd24\",\"resourceVersion\":\"8108227\",\"creationTimestamp\":\"2024-08-29T14:37:04Z\",\"managedFields\":[{\"manager\":\"kube-scheduler\",\"operation\":\"Update\",\"apiVersion\":\"events.k8s.io/v1\",\"time\":\"2024-08-29T14:37:04Z\",\"fieldsType\":\"FieldsV1\",\"fieldsV1\":{\"f:action\":{},\"f:eventTime\":{},\"f:note\":{},\"f:reason\":{},\"f:regarding\":{},\"f:reportingController\":{},\"f:reportingInstance\":{},\"f:type\":{}}}]},\"eventTime\":\"2024-08-29T14:37:04.418330Z\",\"reportingController\":\"default-scheduler\",\"reportingInstance\":\"default-scheduler-docker-desktop\",\"action\":\"Binding\",\"reason\":\"Scheduled\",\"regarding\":{\"kind\":\"Pod\",\"namespace\":\"pods\",\"name\":\"pod\",\"uid\":\"52682376-1685-4d80-b507-9e13d827a728\",\"apiVersion\":\"v1\",\"resourceVersion\":\"8108224\"},\"note\":\"Successfully assigned pods/pod1 to docker-desktop\",\"type\":\"Normal\",\"deprecatedSource\":{},\"deprecatedFirstTimestamp\":null,\"deprecatedLastTimestamp\":null}"}`

**Client Version**
`21.0.1`

**Kubernetes Version**
`1.29.1`

**Java Version**
Java 21

**Expected behavior**
The field is present or ignored in validation. 

**Server (please complete the following information):**
 - OS: Windows 10 / Ubuntu 22 WSL
 - Environment: docker-desktop

Anubi5x0:
Download 
[...]

PR #3744, opened 2024-10-17
Load any Submit resources without mentioning kind types and api group/versions #3336

vaidikcode:
#3336 As described I tried to follow your approach and created a method to load and submit resources and then passed it on to the Api handler
I don't know where to mention the use case example-

public static void main(String[] args) throws Exception {
   
    String yamlContent = "apiVersion: v1\n"
            + "kind: Pod\n"
            + "metadata:\n"
            + "  name: my-pod\n";

   
    Object createdResource = Yaml.loadAndSubmitResource(yamlContent);

 
    System.out.println(createdResource);

### DIFF
--- a/examples/examples-release-19/pom.xml
+++ b/examples/examples-release-19/pom.xml
@@ -93,7 +93,7 @@
 		<dependency>
 			<groupId>com.amazonaws</groupId>
 			<artifactId>aws-java-sdk-sts</artifactId>
-			<version>1.12.773</version>
+			<version>1.12.774</version>
 		</dependency>
 
 	</dependencies>
--- a/examples/examples-release-20/pom.xml
+++ b/examples/examples-release-20/pom.xml
@@ -93,7 +93,7 @@
 		<dependency>
 			<groupId>com.amazonaws</groupId>
 			<artifactId>aws-java-sdk-sts</artifactId>
-			<version>1.12.773</version>
+			<version>1.12.774</version>
 		</dependency>
 
 	</dependencies>
--- a/pom.xml
+++ b/pom.xml
@@ -152,7 +152,7 @@
       <dependency>
         <groupId>com.amazonaws</groupId>
         <artifactId>aws-java-sdk-sts</artifactId>
-        <version>1.12.773</version>
+        <version>1.12.774</version>
       </dependency>
       <dependency>
         <groupId>com.google.protobuf</groupId>


vaidikcode:
@yue9944882  Yes, I think it would be much better. Can you elaborate on how I can do this? Can you also tell me if I need to include an example code to use this method, and if yes, please tell me where to include the example code.

util/src/main/java/io/kubernetes/client/util/Yaml.java
```diff
@@ -59,6 +58,100 @@ public class Yaml {
...
 
+  /**
+   * Load an API object from a YAML string representation without already knowing the object type and loads and submit any type of resource
+   * V1Pod
+   *
+   * @param content The YAML content
+   * @return An instantiation of the object.
+   * @throws IOException If an error occurs while reading the YAML.
+   */
+  public static Object loadAndSubmitResource(String content) throws Exception {
```
brendandburns, line 69: It's more flexible to load from a `Reader` rather than a `String` you can always create a `StringReader`

util/src/main/java/io/kubernetes/client/util/Yaml.java
```diff
@@ -59,6 +58,100 @@ public class Yaml {
...
+    //loading yaml content as unstructured object
+    Object unstructuredObject = Yaml.load(new StringReader(content));
+
+    if (!(unstructuredObject instanceof Map)) {
```
brendandburns, line 73: fwiw, this will only work for single-instance YAML files, if people put multiple objects in a single file, it will fail. That's fine for a starting point, but you may want to document it somewhere.

util/src/main/java/io/kubernetes/client/util/Yaml.java
```diff
@@ -59,6 +58,100 @@ public class Yaml {
...
+      throw new IllegalArgumentException("Invalid YAML");
```
brendandburns, line 74: I think `ParseException` might be more appropriate? (here and below)

util/src/main/java/io/kubernetes/client/util/Yaml.java
```diff
@@ -59,6 +58,100 @@ public class Yaml {
...
+    if (apiVersion.contains("/")) {
+      String[] parts = apiVersion.split("/");
+      group = parts[0];
+      version = parts[1];
+    } else {
+      group = "";
+      version = apiVersion;
+    }
+
+    //creating resource plurals from kind
```
brendandburns, line 101: Consider using the `Discovery` class here instead https://github.com/kubernetes-client/java/blob/f9c82f31b7b6833808bc8880dfc812f18b81851a/util/src/main/java/io/kubernetes/client/Discovery.java#L37

util/src/main/java/io/kubernetes/client/util/Yaml.java
```diff
@@ -59,6 +58,100 @@ public class Yaml {
...
+    String resourcePlural = kind.toLowerCase(Locale.ROOT) + "s";  // Simplistic pluralization logic
+
+    //getting the strongly typed object from model mapper
+    Class<?> modelClass = ModelMapper.getApiTypeClass(group, version, kind);
```
brendandburns, line 105: I think you can say `Class<? implements KubernetesObject>` here?

brendandburns:
Some style comments. Generally this code looks good to me.

Please add unit tests.

vaidikcode:
@brendandburns I've made the changes to the code and added unit tests, just as you suggested.

brendandburns:
@vaidikcode looks good to me. I agree with @yue9944882 that this is a better fit for `KubectlCreate.java` but otherwise ready to merge once you move it there.

brendandburns:
Alternately, feel free to keep the loader code in `Yaml.java` but move the create code to `KubectlCreate`

extended/src/main/java/io/kubernetes/client/extended/kubectl/KubectlCreate.java
```diff
@@ -65,4 +74,119 @@ public ApiType execute() throws KubectlException {
...
+
+    // Validate apiVersion and kind
+    if (apiVersion == null || kind == null) {
+      throw new ParseException("YAML does not contain apiVersion or kind", 0);
+    }
+
+    // Finding the resource in the discovery class
+    Set<Discovery.APIResource> resources = discovery.findAll();
+    Optional<Discovery.APIResource> apiResource = resources.stream()
+            .filter(r -> r.getKind().equalsIgnoreCase(kind))
```
brendandburns, line 117: You need to check group/version here also. There are a few Kinds which are duplicates in several different api groups :(

brendandburns: See https://github.com/kubernetes-client/java/issues/3668#issuecomment-2324638774

Core.V1Event vs Events.V1Event

util/src/main/java/io/kubernetes/client/util/Yaml.java
```diff
@@ -342,27 +342,27 @@ protected MappingNode representJavaBean(Set<Property> properties, Object javaBea
       node.setTag(Tag.MAP);
       // Sort the output of our map so that we put certain keys, such as apiVersion, first.
       Collections.sort(
-          node.getValue(),
```
brendandburns, line 345: I think these are all whitespace changes. please revert.

brendandburns:
Thanks for the update. One change is required for places where kind is the same across multiple API types. Also request to revert whitespace only changes on yaml.java

Thanks for the patience!

PR #3759, opened 2024-10-24
Load any Submit resources without mentioning kind types and api group/versions #3336 

vaidikcode:
I've made the changes to the code as per your suggestion. I couldn't revert back for some reason, so I created a new PR. The code now effectively handles duplicate kinds too.


brendandburns:
@vaidikcode this is failing CI/CD, once that's fixed should be ok to merge.

PR #3770, opened 2024-10-27
Modify ModelMapper to return DynamicKubernetesObject

scovl:
I solved the issue [https://github.com/kubernetes-client/java/issues/3336](https://github.com/kubernetes-client/java/issues/3336) by modifying the `getApiTypeClass` method in `ModelMapper.java` to return `DynamicKubernetesObject.class` when a specific class for a given `GroupVersionKind` is not found. This change enables the Kubernetes Java client to handle YAML resources of any type without differentiating between specific `kind` values. 

By defaulting to a dynamic Kubernetes object, the client can now load unstructured YAML, extract the necessary `apiVersion` and `kind`, map it to a generic handler, and interact with the Kubernetes API accordingly. This approach allows for the submission and management of any YAML resource, enhancing the client's flexibility and resolving the issue of handling resources without prior knowledge of their types. 

@brendandburns  @yue9944882 

/lgtm


brendandburns:
For some reason this didn't trigger CI/CD the first time, so close/reopen to re-trigger CI/CD.

/lgtm
/approve

PR #3786, opened 2024-11-09
Since PR generation is still not working, manual clone of the 22.0.0 release changes

brendandburns:
cc @yue9944882


yue9944882 (commented):
/lgtm
/approve

yue9944882:
@brendandburns what do you think if we workaround the PR generation restriction by signing the generated commit with our personal GPG key? i'm thinking about importing our own GPG key as a repository secret so that the github action workflow can use the key to sign the commits. that way the generated commit will comply to CLA check.

PR #3817, opened 2024-12-09
Refactor: Merge TokenFile auth with refresh auth

yue9944882:
This pull request addresses the TODO comment:

> // TODO: Merge this with TokenFileAuthentication.

It simplifies the TokenFileAuthentication by taking RefreshAuthentication as the parent class

### DIFF
--- a/util/src/main/java/io/kubernetes/client/util/credentials/TokenFileAuthentication.java
+++ b/util/src/main/java/io/kubernetes/client/util/credentials/TokenFileAuthentication.java
@@ -12,63 +12,32 @@ limitations under the License.
 */
 package io.kubernetes.client.util.credentials;
 
-import io.kubernetes.client.openapi.ApiClient;
 import java.io.IOException;
 import java.nio.charset.Charset;
 import java.nio.file.Files;
 import java.nio.file.Paths;
-import java.time.Instant;
-import okhttp3.Interceptor;
-import okhttp3.OkHttpClient;
-import okhttp3.Request;
-import okhttp3.Response;
+import java.time.Duration;
 
 // TODO: prefer OpenAPI backed Auentication once it is available. see details in
 // https://github.com/OpenAPITools/openapi-generator/pull/6036. currently, the
 // workaround is to hijack the http request.
-public class TokenFileAuthentication implements Authentication, Interceptor {
-  private String file;
-  private String token;
-  private Instant expiry;
-
+public class TokenFileAuthentication extends RefreshAuthentication {
   public TokenFileAuthentication(String file) {
-    this.expiry = Instant.MIN;
-    this.file = file;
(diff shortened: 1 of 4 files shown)


util/src/main/java/io/kubernetes/client/util/credentials/TokenFileAuthentication.java
```diff
@@ -26,49 +27,23 @@
...
-
-  @Override
-  public void provide(ApiClient client) {
-    OkHttpClient withInterceptor = client.getHttpClient().newBuilder().addInterceptor(this).build();
-    client.setHttpClient(withInterceptor);
+  private static String getToken(String file) {
+    try {
+      return new String(Files.readAllBytes(Paths.get(file)), Charset.defaultCharset()).trim();
+    } catch (IOException e) {
+      throw new RuntimeException("Cannot read file: " + file);
```
brendandburns, line 45: We should probably convert this to some sort of non-RuntimeException like "FailedTokenAquisitionException" and handle that inside the `RefreshAuthentication.java`

brendandburns:
One small comment and looks like format/testing is failing.

util/src/main/java/io/kubernetes/client/util/credentials/TokenFileAuthentication.java
yue9944882: That sounds like a good idea but I ended up adding the exception conversion as a runtime/unchecked exception in the parent class `RefreshAuthentication`. I also tried making it a checked exception in the token supplier function however it will break backward compatibility so I implemented the unchecked exception in the new commit. one example that compatibility would be broken is here where we're loading the token eagerly in `RefreshAuthentication` :

> https://github.com/kubernetes-client/java/blob/master/util/src/main/java/io/kubernetes/client/util/credentials/RefreshAuthentication.java#L43

so the new checked exception would have to be added to the constructor which will require all places referencing it adding explicit handling the checked exception. does that make sense?

util/src/test/java/io/kubernetes/client/util/credentials/TokenFileAuthenticationTest.java
```diff
@@ -63,19 +65,6 @@ void tokenProvided() throws ApiException {
         1,
         getRequestedFor(urlPathEqualTo("/api/v1/pods"))
             .withHeader("Authorization", equalTo("Bearer token1")));
-
-    this.auth.setFile(SERVICEACCOUNT_TOKEN2_PATH);
```
yue9944882, line 67: these unit tests are no longer needed as it was originally for testing token expiry/rotation behavior. now it's decoupled/delegated to the RefreshAuthentication class.

brendandburns:
/lgtm
/approve

Looks good to me. Let's consider refactoring these exceptions in the future.

PR #3869, opened 2025-01-16
Clarifying legacy release workflow instructions

yue9944882:
@brendandburns


brendandburns:
Ah, right. Thanks!

/lgtm
/approve

Issue #3891, opened 2025-02-01
Remove unused dependencies from pom.xml


brendandburns:
I think there are a number of unused dependencies that for historical reasons are still in our POM files.

We should use something like `mvn dependency:analyze` to strip these out.

yue9944882:
```
[WARNING] Used undeclared dependencies found:
[WARNING]    org.springframework:spring-beans:jar:6.2.2:compile
[WARNING]    io.kubernetes:client-java-extended:jar:23.0.0-SNAPSHOT:compile
[WARNING]    com.google.code.gson:gson:jar:2.12.1:compile
[WARNING]    org.springframework:spring-core:jar:6.2.2:compile
[WARNING]    io.swagger:swagger-annotations:jar:1.6.15:compile
[WARNING]    org.slf4j:slf4j-api:jar:2.0.16:compile
[WARNING]    org.springframework.boot:spring-boot-autoconfigure:jar:3.4.2:compile
[WARNING]    org.jetbrains:annotations:jar:13.0:compile
[WARNING] Unused declared dependencies found:
[WARNING]    io.kubernetes:client-java-spring-integration:jar:23.0.0-SNAPSHOT:compile
[WARNING]    org.springframework.boot:spring-boot:jar:3.4.2:compile
[WARNING]    org.junit.jupiter:junit-jupiter:jar:5.11.4:test
[WARNING]    org.springframework.boot:spring-boot-test:jar:3.4.2:test
[WARNING]    org.springframework:spring-test:jar:6.2.2:test

```

Yea we're safe to remove these packages I think

Rishabh-mikku:
Hi @brendandburns and @yue9944882 , I have understood this issue and I am willing to work on it. Can you please assign me this issue.

brendandburns:
@Rishabh-mikku feel free to send PRs.

WaleedAdnan08:
H @yue9944882! I'd like to work on this issue if It isn't closed yet.

Aariyan007:
Is this still up?

dasarathirout:
During [kubernetes-client/gen/penapi/openapi-generator/generate_client_in_container.sh#L105](https://github.com/kubernetes-client/gen/blob/master/openapi/openapi-generator/generate_client_in_container.sh#L105 )

[...]

PR #3907, opened 2025-02-07
chore(deps): remove unused jakarta.ws.rs-api dependency

JynLeazy:


JynLeazy:
as part of the open issue "Remove unused dependencies from pom.xml #3891"

kubernetes/pom.xml
```diff
@@ -182,10 +182,6 @@
       <groupId>com.google.code.findbugs</groupId>
       <artifactId>jsr305</artifactId>
     </dependency>
-    <dependency>
-      <groupId>jakarta.ws.rs</groupId>
-      <artifactId>jakarta.ws.rs-api</artifactId>
-    </dependency>
```
yue9944882, line 188: we might also want to remove this from the root pom i think

JynLeazy1: let me verify if the other modules don't need it

JynLeazy: @yue9944882  is done

Issue #3809, opened 2024-11-29
java client is not taking the updated ca.crt/token and error out


suryag10:
**Describe the bug**
updated ca.crt/token are not considered by the java k8s client and errors out
**java.util.concurrent.ExecutionException: io.kubernetes.client.openapi.ApiException: javax.net.ssl.SSLHandshakeException: PKIX path validation failed: java.security.cert.CertPathValidatorException: signature check failed**

**Client Version**
17.0.2

**Kubernetes Version**
1.21.12

**Java Version**
openjdk 17.0.12 2024-07-16 LTS

**To Reproduce**
update the k8s ca.crt and token with the new certificates.

**Expected behavior**
Java client should consider the latest ca.crt and token and K8S API access should succeed

**Server (please complete the following information):**
 - OS: Linux
 - Environment : Container
 - Cloud : Vanilla K8S

suryag10:
Any Help here?

suryag10:
Thanks @brendandburns , Will check if i can submit the PR for the same.

Hakky54:
> Yeah, the current auth code doesn't reload certificates if the files change. It would be a fairly significant effort to do so, but if you wanted to take it on, we'd be happy to review the design/PRs.

I have a library which makes it easy to reload the ssl while listening for file changes. See here https://github.com/Hakky54/sslcontext-kickstart?tab=readme-ov-file#support-for-reloading-ssl-at-runtime

I can also initiate a PR if you want

suryag10:
@Hakky54 Sure, can you please provide the patch? I will also check from my side.

Hakky54:
Hi @suryag10 and @brendandburns I added my PR for adding this feature to kubernetes-client, see here: https://github.com/kubernetes-client/java/issues/3809

brendandburns:
I would rather that we fixed this via the `isExpired` method on the `Authenticator` class (or we refactored authenticator to accommodate any changes needed for reloading).

Hakky54:
I dropped a new comment couple of days ago on the PR which I created before, not quite sure whether someone noticed it, but I copy it here.

I have the feeling that there is no progress yet with a different implementation for reloading ssl. Would you like to reconsider this PR https://github.com/kubernetes-client/java/pull/3927 for this feature request?

PR #3927, opened 2025-02-17
Added SSL reloading mechanism

Hakky54:
This PR resolves the following issue: https://github.com/kubernetes-client/java/issues/3809 

It uses my own ssl library which can be find here: https://github.com/Hakky54/sslcontext-kickstart
It has a built in reloading mechanism which is thread safe.

The client module will create an initial  ssl configuration as base sslfactory alongside with the client. If set cacerts, setkeymanager method is being called when the the library user has new ssl material then the reload mechanism will be executed. It will use the existing http client bowever it will reload the internal ssl properties.

### DIFF
--- a/e2e/pom.xml
+++ b/e2e/pom.xml
@@ -43,9 +43,8 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>org.slf4j</groupId>
-            <artifactId>slf4j-simple</artifactId>
-            <version>${slf4j.version}</version>
+            <groupId>io.github.hakky54</groupId>
+            <artifactId>logcaptor</artifactId>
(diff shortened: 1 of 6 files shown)


brendandburns:
Apologies, but I don't think we want to take a direct dependency on your library for SSL. Can you fit this within the standard Java plugin/extension mechanisms for SSL?

brendandburns:
Thanks for the understanding, I'm going to close this PR. I think there are better ways to solve the issue you cited than taking a full dependency on this library for SSL handling.

Hakky54:
@brendandburns I have the feeling that there is no progress yet with a different implementation for reloading ssl. Would you like to reconsider this PR?

PR #3929, opened 2025-02-17
Delete .devcontainer directory

governmentusa-debug:


brendandburns:
I don't think we want to delete the devcontainer. Is there a reason why you think we should?

PR #3930, opened 2025-02-19
fix: add missing continue token when build cluster list call

cyrilcsr:
Hi, there
This pr fixes missing continue token when listing resources without namespace specified. This will cause when passing continue token, we always get the result of first page. Seems in makeClusterListCallBuilder() the continue param from ListOptions is never used.


brendandburns:
Looks good. It would be great to add unit tests so that we don't regress in the future.

cyrilcsr:
> Looks good. It would be great to add unit tests so that we don't regress in the future.

I have added a few unit tests for building list call. I'm not sure which test file should I put these unit tests, so I created a new GenericKubernetesListApiTest file which refers to the GenericKubernetesGetApiTest file. Please take a look.

yue9944882 (commented):
/lgtm
/approve

Issue #4020, opened 2025-04-21
ApiClient cannot deserialize responses larger than 2GB


klin1344:
**Describe the bug**
The `ApiClient` cannot handle responses larger than 2GB. 

**Client Version**
`22.0.1-legacy` is what we're currently using, but this issue occurs in all versions

**Kubernetes Version**
`1.30.9`

**Java Version**
Java 21

**To Reproduce**
1. Set up a `SharedIndexInformer` for a resource in a cluster where the total size of all resources > 2GB.
2. The initial `list()` in the `ReflectorRunnable` will fail with:

```
class io.kubernetes.client.openapi.models.V1Pod#Reflector loop failed unexpectedly

java.lang.IllegalArgumentException: byteCount: 2293832235
  at okio.Buffer.readString(Buffer.kt:306)
  at okio.Buffer.readString(Buffer.kt:302)
  at okio.RealBufferedSource.readString(RealBufferedSource.kt:96)
  at okhttp3.ResponseBody.string(ResponseBody.kt:187)
  at io.kubernetes.client.openapi.ApiClient.deserialize(ApiClient.java:758)
  at io.kubernetes.client.openapi.ApiClient.handleResponse(ApiClient.java:978)
  at io.kubernetes.client.openapi.ApiClient.execute(ApiClient.java:905)
  at io.kubernetes.client.informer.SharedInformerFactory$1.list(SharedInformerFactory.java:271)
  at io.kubernetes.client.informer.cache.ReflectorRunnable.run(ReflectorRunnable.java:91)
  at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:572)
  at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:358)
  at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:305)
  at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
  at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
  at java.lang.Thread.run(Thread.java:1583)
```

**Expected behavior**
The `ApiClient` can deserialize responses greater than 2GB.

**Additional context**
[...]

PR #4021, opened 2025-04-21
ApiClient: support deserializing from InputStream instead of String to bypass 2GB Java String limit

klin1344:
Fixes issue https://github.com/kubernetes-client/java/issues/4020 .

Makes two changes:
1. In `ApiClient`, if we are deserializing JSON, get the response body as an `InputStream` instead of `String` for deserialization.
2. Adds a new `deserialize()` method in `JSON.java` which accepts `InputStream`.

These two changes enable the `ApiClient` to bypass the 2GB `String` length limit in Java and deserialize large responses greater than 2GB into Java objects.

### DIFF
--- a/kubernetes/src/main/java/io/kubernetes/client/openapi/ApiClient.java
+++ b/kubernetes/src/main/java/io/kubernetes/client/openapi/ApiClient.java
@@ -38,7 +38,6 @@ import java.security.SecureRandom;
 import java.security.cert.Certificate;
 import java.security.cert.CertificateException;
 import java.security.cert.CertificateFactory;
-import java.security.cert.X509Certificate;
 import java.text.DateFormat;
 import java.time.LocalDate;
 import java.time.OffsetDateTime;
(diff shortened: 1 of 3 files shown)


brendandburns:
Thanks for the PR. This code is generated. You need to make the changes here:

https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/Java/ApiClient.mustache

And then we will get the update next time that we regenerate the code.

klin1344:
> Fixes issue #4020 .
> 
> Makes two changes:
> 
> 1. In `ApiClient`, if we are deserializing JSON, get the response body as an `InputStream` instead of `String` for deserialization.
> 2. Adds a new `deserialize()` method in `JSON.java` which accepts `InputStream`.
> 
> These two changes enable the `ApiClient` to bypass the 2GB `String` length limit in Java and deserialize large responses greater than 2GB into Java objects.

Ah I see, thank you! Opened [this PR](https://github.com/OpenAPITools/openapi-generator/pull/21115) for the `openapi-generator` repo and will close this one.

PR #4086, opened 2025-06-06
Switch to PAT_TOKEN for git branch push

yue9944882:
> https://github.com/kubernetes-client/java/actions/runs/15476712105/job/43573728659

The github action lost permission to push branch after maven release even though the maven release was successful. Looks like the permission got tighten again recently because we didn't have this error when publishing 24.0.0.. This PR proposes to use PAT_TOKEN which has the repo write permission to automate the version release process.

> remote: Permission to kubernetes-client/java.git denied to github-actions[bot].
> fatal: unable to access 'https://github.com/kubernetes-client/java.git/': The requested URL returned error: 403

We need to choose a poison between having a long manual release process and the risk of increasing the permission for github action..

/cc @brendandburns

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -28,6 +28,8 @@ jobs:
           echo "${{ github.event.inputs.nextDevelopmentVersion }}" | perl -ne 'die unless m/^\d+\.\d+\.\d+-SNAPSHOT$/'
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          token: ${{ secrets.PAT_TOKEN }}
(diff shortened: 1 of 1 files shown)


brendandburns:
Why is this permission required for the checkout step?

yue9944882:
@brendandburns The checkout step is where we initialize the identity/permission for the follow-up git commands including `push`, `tag`, etc..

>https://github.com/actions/checkout/blob/main/README.md?plain=1#L50-L52

brendandburns:
/lgtm
/approve

Thanks for the link to the docs.

PR #4107, opened 2025-06-24
separate initial load to avoid retaining list items in memory

rafaelri:
Local variables `list` and `items` will retain the list of objects returned in the initial list in the heap while the `Thread` that is running `run` is alive. Leading to a minor but still a leak since those objects could be collected if the local variable was made out of scope.
This PR separates the initial load in a separate method that returns the `resourceVersion` and shortens the scope of the variables mentioned to allow garbage collection.

```java
 ApiListType list =
          listerWatcher.list(
              new CallGeneratorParams(Boolean.FALSE, getRelistResourceVersion(), null));

      V1ListMeta listMeta = list.getMetadata();
      String resourceVersion = listMeta.getResourceVersion();
      List<? extends KubernetesObject> items = list.getItems();
```

### DIFF
--- a/util/src/main/java/io/kubernetes/client/informer/cache/ReflectorRunnable.java
+++ b/util/src/main/java/io/kubernetes/client/informer/cache/ReflectorRunnable.java
@@ -12,6 +12,18 @@ limitations under the License.
 */
 package io.kubernetes.client.informer.cache;
 
+import java.io.IOException;
+import java.net.ConnectException;
+import java.net.HttpURLConnection;
+import java.time.Duration;
+import java.util.List;
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.BiConsumer;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import io.kubernetes.client.common.KubernetesListObject;
 import io.kubernetes.client.common.KubernetesObject;
 import io.kubernetes.client.informer.EventType;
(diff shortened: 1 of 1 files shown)


util/src/main/java/io/kubernetes/client/informer/cache/ReflectorRunnable.java
```diff
@@ -12,6 +12,18 @@
 */
 package io.kubernetes.client.informer.cache;
 
+import java.io.IOException;
```
brendandburns, line 15: Please revert this change and maintain the orginal order of imports.

util/src/main/java/io/kubernetes/client/informer/cache/ReflectorRunnable.java
```diff
@@ -278,6 +269,11 @@ private void watchHandler(Watchable<ApiType> watch) {
         log.debug("{}#Receiving resourceVersion {}", apiTypeClass, lastSyncResourceVersion);
       }
     }
+    try {
```
brendandburns, line 272: why is adding this code needed?

brendandburns:
Thank you for the PR!

I think it would be better if the PR was smaller/more focused. Please update this PR to just separate out the list/watch logic for your memory management purposes, and if you want to send a second refactor PR that's great too.

fwiw, I believe that explicitly setting things to `null` will also achieve the same nudge to the garbage collector and may be less invasive that this code refactor.

brendandburns:
Are there benchmarks or other validation that we can add to prove the effectiveness of this code?

Issue #4110, opened 2025-06-25
AWS V1 EOL


daniel-b2c2:
Looking at this dependency:

https://mvnrepository.com/artifact/io.kubernetes/client-java/24.0.0

The transitive dependency list shows a dependency on V1 of the AWS SDK, V2 has been out for a long time and amazon have set the EOL date for December this year - this was announced a year ago, so 18 months notice.

![Image](https://github.com/user-attachments/assets/6b1c75ee-7a1f-4f27-9c2e-90a15edac901)

![Image](https://github.com/user-attachments/assets/f05c677e-ff3f-4999-bce8-d99bd0e918cc)

Running with this library actually causes a warning message on STDOUT which you can suppress, but it's there for a reason...

```
com.amazonaws.util.VersionInfoUtils The AWS SDK for Java 1.x entered maintenance mode starting July 31, 2024 and will reach end of support on December 31, 2025. For more information, see https://aws.amazon.com/blogs/developer/the-aws-sdk-for-java-1-x-is-in-maintenance-mode-effective-july-31-2024/
You can print where on the file system the AWS SDK for Java 1.x core runtime is located by setting the AWS_JAVA_V1_PRINT_LOCATION environment variable or aws.java.v1.printLocation system property to 'true'.
This message can be disabled by setting the AWS_JAVA_V1_DISABLE_DEPRECATION_ANNOUNCEMENT environment variable or aws.java.v1.disableDeprecationAnnouncement system property to 'true'.
```

We're getting pretty close to the cut off, is the `kubernetes-client/java` team aware of this / Is there already a plan to remediate / accepting PRs?

brendandburns:
We're happy to accept PRs, that code is used in a pretty minimal way to implement Auth exclusively for AWS clusters:

https://github.com/kubernetes-client/java/blob/master/util/src/main/java/io/kubernetes/client/util/credentials/EKSAuthentication.java

We should probably mark that dependency as optional in the pom and if no one updates the dependency before they deprecate the library, we can just delete that EKS file.

daniel-b2c2:
Thanks for confirming, I'll put together a PR to use V2.

daniel-b2c2:
@brendandburns 

 I may have misunderstood, I've read your comment again and realise:
- my intention was to migrate it,
- your intention is to decommission it?

I'm fine either way, though I've put a PR that I think is correct for option 1 (migration), I haven't actually had a chance to test it on a real cluster.

PR #4112, opened 2025-06-27
#4110 Migrate EKS authentication from AWS 1.x -> AWS 2.x

daniel-b2c2:
Addresses #4110

AWS SDK 1.x is EOL in December 2025. 

This Pull requests migrates the EKS authentication code which generates a token via the STS service to use the AWS SDK 2.x library.

Tests pass but it's worth checking the actual output headers are right. I do not have an actual cluster to test this authentication with so testing is welcome.


daniel-b2c2:
Ah paperwork, my personal account already has a CLA. I will resubmit from there. Closing PR.

PR #4140, opened 2025-07-17
fix: correct fieldManager parameter in makeClusterCreateCallBuilder

### LINKED ISSUE #4138: GenericKubernetesApi.makeClusterCreateCallBuilder
**Describe the bug**
In the `GenericKubernetesApi.java` file, the private method `makeClusterCreateCallBuilder` incorrectly builds the API call for creating a cluster-scoped resource.
It mistakenly calls the `.dryRun()` method twice, with the second call incorrectly passing the `fieldManager` value from `CreateOptions`. This overwrites the actual `dryRun` value and fails to set the `fieldManager` parameter.

Here is the problematic code snippet:
// in [/util/src/main/java/io/kubernetes/client/util/generic/GenericKubernetesApi.java](https://github.com/kubernetes-client/java/blob/v24.0.0/util/src/main/java/io/kubernetes/client/util/generic/GenericKubernetesApi.java)
```java
  private CallBuilder makeClusterCreateCallBuilder(
      ApiType object, final CreateOptions createOptions) {
    return () ->
        customObjectsApi.createClusterCustomObject(
            this.apiGroup,
            this.apiVersion,
            this.resourcePlural,
            object)
                .dryRun(createOptions.getDryRun())
                .dryRun(createOptions.getFieldManager())
                .buildCall(null);
  }
```

The second .dryRun() call should be .fieldManager().

**Client Version**
e.g. `24.0.0`

**Kubernetes Version**
e.g. `1.30.13`

**Java Version**
e.g. Java 21

**To Reproduce**
1. Instantiate `GenericKubernetesApi` for a cluster-scoped custom resource.
2. Create an instance of `CreateOptions` with a specific `fieldManager` value (e.g., "`my-app-controller`").
3. Call the `api.create(customObject, createOptions)` method.
4. Observe the HTTP request sent to the Kubernetes API server.
5. You will see that the request includes a `dryRun=my-app-controller` parameter, and the `fieldManager` parameter is missing.

**Expected behavior**
The API call builder should correctly map the options. The `dryRun` value from `CreateOptions` should be used for the `dryRun` parameter, and the `fieldManager` value should be used for the `fieldManager` parameter.

**Server (please complete the following information):**
 - OS: MacOS 15.2
 - Environment : Docker Desktop & IDE

**Additional context**

zongzewu23:
Fix typo where .dryRun() was called twice instead of .fieldManager() in GenericKubernetesApi.makeClusterCreateCallBuilder method.

Fixes #4138

### DIFF
--- a/util/src/main/java/io/kubernetes/client/util/generic/GenericKubernetesApi.java
+++ b/util/src/main/java/io/kubernetes/client/util/generic/GenericKubernetesApi.java
@@ -707,7 +707,7 @@ public class GenericKubernetesApi<
             this.resourcePlural,
             object)
                 .dryRun(createOptions.getDryRun())
-                .dryRun(createOptions.getFieldManager())
+                .fieldManager(createOptions.getFieldManager())
                 .buildCall(null);
   }
 


brendandburns:
/lgtm
/approve

Thanks! (bonus points if we add a unit test for this :)

PR #4152, opened 2025-07-21
Add ApiClient in ClientBuilder

iseki0:
Since the ApiClient has another constructor: apiClient(OkHttpClient). It's useful to allowing user customize the argument. OkHttp also recommend user to reuse the client instance(to sharing the internal dispatchers and connection pool).


brendandburns:
/lgtm
/approve

Thanks for the PR

iseki0:
Consider make a release?

PR #4180, opened 2025-08-08
Incorrect User-Agent #3857 Read version from Configuration

dasarathirout:


brendandburns:
Thanks for the PR. Unfortunately, this code is generated, so the changes need to be made in the upstream code generator:

https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/Java/libraries/okhttp-gson/ApiClient.mustache

Then we'll pick it up in the next generation run.

PR #4182, opened 2025-08-11
 Remove Unused declared dependencies found in pom.xml #3891

dasarathirout:
Remove Unused declared dependencies found in pom.xml.
 mvn dependency:analyze output All "[WARNING] Unused declared dependencies found" are not correct, 
 few are required for compilation


yue9944882 (commented):
/lgtm
/approve

thanks for the cleanup

PR #4295, opened 2025-09-29
fix feat `TrustManager` that accepts all certificates

odaysec:
https://github.com/kubernetes-client/java/blob/c14bb93680786910ecb01bce1a428a3654ee4544/kubernetes/src/main/java/io/kubernetes/client/openapi/ApiClient.java#L1499-L1520

To address this vulnerability, we must remove the custom `TrustManager` and `HostnameVerifier` that accept all certificates.  
- If the intention is to allow self-signed certificates in some cases, explicit trusted certificates should be loaded from a secure location into a `KeyStore`, as demonstrated in the good example in the background.  
- The `TrustManager` should be initialized via a `TrustManagerFactory` with a `KeyStore` containing only the explicitly trusted certificates.
- Hostname verification should *not* be disabled; use the standard verifier unless on localhost for very specific reasons.
- Changes are required within the `applySslSettings()` method:  
  - Remove (or replace) the block starting on line 1498 that creates an insecure `TrustManager` and `HostnameVerifier`.
  - Instead, if `verifyingSsl` is `false`, throw an exception or require the user to provide a trusted CA cert, as in the `else` branch. This eliminates the insecure option entirely.

### References
[Security with HTTPS and SSL](https://developer.android.com/training/articles/security-ssl)

### DIFF
--- a/kubernetes/src/main/java/io/kubernetes/client/openapi/ApiClient.java
+++ b/kubernetes/src/main/java/io/kubernetes/client/openapi/ApiClient.java
@@ -1496,28 +1496,9 @@ public class ApiClient {
             TrustManager[] trustManagers;
             HostnameVerifier hostnameVerifier;
             if (!verifyingSsl) {
-                trustManagers = new TrustManager[]{
-                        new X509TrustManager() {
-                            @Override
-                            public void checkClientTrusted(java.security.cert.X509Certificate[] chain, String authType) throws CertificateException {
-                            }
-
-                            @Override
-                            public void checkServerTrusted(java.security.cert.X509Certificate[] chain, String authType) throws CertificateException {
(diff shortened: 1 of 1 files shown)


brendandburns:
@odaysec this code comes from the upstream code generator:

https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/Java/libraries/okhttp-gson/ApiClient.mustache#L1779

If you want to make a change, that's the place to make it, but I suspect that it will not be accepted, as there are circumstances when people want to use self-signed certificates in testing and they need this capability.

But if you do get changes accepted upstream, we'll pick them up next time we regenerate.

I'm going to close this PR for now since we won't merge this one.

PR #4343, opened 2025-10-23
Remove release 17~20 example

yue9944882:
These older release artifacts are no longer usable hence deleting them


brendandburns (approved):
/lgtm
/approve

PR #4351, opened 2025-10-28
Update parent version to 25.0.1-SNAPSHOT to fix fluent-gen

brendandburns:


brendandburns:
Typo removed, thanks for catching that!

PR #4417, opened 2025-11-20
Improve TLS handling in OpenIDConnectAuthenticator

abnsy:
I’m sending a small security improvement for the OIDC authenticator. When the kubeconfig doesn’t include idp-certificate authority-data the code leaves sslContext as null and falls back to the OS trust store. This can behave differently depending on the environment and isn’t always secure. In this update I added a default SSLContext that always uses the system CA.... If the user provides a custom CA, that one is still loaded and used instead. No behavior change other than making TLS more consistent and safer. Thanks!

### DIFF
--- a/util/src/main/java/io/kubernetes/client/util/authenticators/OpenIDConnectAuthenticator.java
+++ b/util/src/main/java/io/kubernetes/client/util/authenticators/OpenIDConnectAuthenticator.java
@@ -113,13 +113,27 @@ public class OpenIDConnectAuthenticator implements Authenticator {
     String clientSecret = (String) config.getOrDefault(OIDC_CLIENT_SECRET, "");
     String idpCert = (String) config.get(OIDC_IDP_CERT_DATA);
 
+    
     SSLContext sslContext = null;
 
+    // Initialize a secure default SSLContext using the system default TrustManager.
+    // This ensures that we always have certificate validation even when no idp-certificate-authority-data is provided.
+    try {
+      TrustManagerFactory defaultTmf =
+          TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
+      defaultTmf.init((KeyStore) null);
+      SSLContext defaultSslContext = SSLContext.getInstance("TLS");
+      defaultSslContext.init(defaultTmf.getTrustManagers(), null, new SecureRandom());
+      sslContext = defaultSslContext;
+    } catch (KeyStoreException | NoSuchAlgorithmException | KeyManagementException e) {
+      throw new RuntimeException("Unable to initialize default TLS context", e);
+    }
+
     if (idpCert != null) {
-      // fist, lets get the pem
+      // first, let's get the pem (custom CA provided by user)
       String pemCert = new String(Base64.getDecoder().decode(idpCert));
 
-      // next lets get a cert object
+      // next let's get a cert object
       // need an alias name to store the certificate in a keystore.  Also
       // java keystores need passwords. this value is as good as any as
       // there isn't anything actually secret being stored.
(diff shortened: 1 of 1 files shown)


brendandburns:
Can you clarify (especially w/ pointers to documentation) why this is preferable to using the OS default key store?

abnsy:
> Can you clarify (especially w/ pointers to documentation) why this is preferable to using the OS default key store?

Sure....happy to clarify.....The main reason is consistency. When sslcontext is null Java falls back to whatever CA bundle the OS or JVM provides. On containers, Windows, Linux distros,or JVM vendors the trusted roots can be completely different....By initializing SSLContext explicitly, the client always behaves the same way and we  know exactly what trust store is used... JSSE also recommends using your own SSLContext  if you want predictable TLS behavior https://docs.oracle.com/javase/8/docs/technotes/guides/security/jsse/JSSERefGuide.html#Customization Not trying to override the OS trust store  just avoiding the silent fallback that  varies across environments... thanks

brendandburns:
So it seems to me that by specifying a `null` keystore when you initialize the trust store you get whatever CA bundles are OS installed anyway, so it's not clear to me that this code is actually doing anything.

It also seems to me based on reading https://docs.oracle.com/javase/8/docs/technotes/guides/security/jsse/JSSERefGuide.html#CustomizingStores that you can specify all this stuff via java properties and the defaults will just work.

Perhaps it would be useful to add a unit test which illustrates the problem with the current code and then show how this update fixes it?

Apologies for being picky here, but whenever we mess with encrypt/auth code I want to make sure it's for a good reason.

abnsy:
> So it seems to me that by specifying a `null` keystore when you initialize the trust store you get whatever CA bundles are OS installed anyway, so it's not clear to me that this code is actually doing anything.
> 
> It also seems to me based on reading https://docs.oracle.com/javase/8/docs/technotes/guides/security/jsse/JSSERefGuide.html#CustomizingStores that you can specify all this stuff via java properties and the defaults will just work.
> 
> Perhaps it would be useful to add a unit test which illustrates the problem with the current code and then show how this update fixes it?
> 
> Apologies for being picky here, but whenever we mess with encrypt/auth code I want to make sure it's for a good reason.

Thanks for the detailed feedback  that helps.

About the null keystore part you’re right that when you do  tmf.init((KeyStore) null), you end up with whatever CA set the JVM/OS  provides The intent wasn’t to replace that, but to make the behavior explicit 
instead of relying on the implicit fallback that only happens when sslContext 
is left null. Right now when sslContext stays null the behavior depends on which 
HttpsURLConnection implementation you hit and what the environment does. 
For example:
- some distroless container images ship with a very limited CA set,
- some enterprise Windows environments inject internal root CAs,
- Alpine uses a different CA bundle than Debian-based images,
- and some embedded JVM runtimes include their own CA store.

So the fallback is whatever the platform feels like, while explicitly 
initializing SSLContext always gives the same behavior across environments. I'm absolutely fine switching this to TLSv1.2 to stay consistent with  the rest of the code and avoid older protocol versions.

On the unit test idea.....yeah, that makes sense. I can put together a small test 
showing the difference between:
1) sslContext == null (current behavior)
2) sslContext initialized with tmf.init(null)
3) sslContext initialized with custom CA

util/src/main/java/io/kubernetes/client/util/authenticators/OpenIDConnectAuthenticator.java
```diff
@@ -113,13 +113,27 @@ public Map<String, Object> refresh(Map<String, Object> config) {
     String clientSecret = (String) config.getOrDefault(OIDC_CLIENT_SECRET, "");
     String idpCert = (String) config.get(OIDC_IDP_CERT_DATA);
 
+    
     SSLContext sslContext = null;
 
+    // Initialize a secure default SSLContext using the system default TrustManager.
+    // This ensures that we always have certificate validation even when no idp-certificate-authority-data is provided.
+    try {
```
macarte: Shouldn't this new block be in the 'else' part of the following test, i.e.  if(idpCert != null) { existing block } else { this block }, otherwise this block runs even when there is an idpCert but the result is not used.   Unless there's some state outside of this method that changes based on this block and those changes are required for the following if(idpCert != null) block

macarte (changes requested):
A few recommendations

karianna:
@yue9944882 I think the GH actions need to be triggered by an ownser/admin of this repo
@abnsy Looks good overall, can you address Mat's feeedback?

karianna:
@abnsy Are you able to resolve the feedback so we can move this forward?

karianna:
Trying again: @abnsy Are you able to resolve the feedback so we can move this forward?

PR #4424, opened 2025-11-21
Automated Generate from openapi release-1.34

yue9944882:


brendandburns:
So I was hoping to pick up the recent changes I made to add tlsServerName support to the code generator, but I think I need to up the commit version that we use in the generator, this did pick up some fluent updates, so I think we may as well merge it, but we probably also need to re-squash it or something so tha the CLA bot is happy.

yue9944882:
just to confirm, does this PR include the tls change you are looking for? note our generation workflow is based on the latest content in this file:

> https://github.com/kubernetes-client/gen/blame/master/openapi/java.sh#L52

brendandburns:
No, we need to wait for an OpenAPI generator release which includes it and then rev the version forward.

We can still merge this though if we want looks like there are some set related fixes.

PR #4491, opened 2025-12-29
Clarify dependencyManagement usage for shared dependencies

Gautam-aman:
### Summary
While investigating removal of unused dependencies from the parent POM, I found
that several entries are required to supply versions for child modules that
declare dependencies without explicit versions.

This PR adds documentation to prevent accidental removal that would break
multi-module builds.

### Verification
- mvn dependency:analyze across multiple modules
- mvn clean verify

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -100,6 +100,14 @@
       <organizationUrl>https://kubernetes.io</organizationUrl>
     </developer>
   </developers>
+  <!--
+  NOTE:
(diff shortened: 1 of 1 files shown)


Gautam-aman:
The CRD Java Models Code Generation job failed during apt-get update due to an Ubuntu mirror sync error (File has unexpected size).
Local builds are clean and there are no codegen scripts in the repo to run locally.
This appears to be CI infrastructure related. Happy to rebase or rerun if needed.

PR #4507, opened 2026-01-09
Fix EndpointSlice deserialization when endpoints is null

Gautam-aman:
Allow V1EndpointSlice to deserialize correctly when the Kubernetes API returns "endpoints": null.
The model validation is relaxed to accept null while still validating array contents when present.

Duplicate and unused imports were also removed during cleanup.


brendandburns:
Unfortunately, this code is generated, so we can't take edits to this file, if you want to make changes it needs to be in the upstream open API code generator:

https://github.com/OpenAPITools/openapi-generator

brendandburns:
~Also, specifically for this, I believe that this is because the Kubernetes API spec definition does not say that null values are allowed here.~

Ah, I see that the spec was revised. We should pick this up the next time that we regenerate.

PR #4627, opened 2026-03-01
Cert manager fix

adithyanms-debug:
This PR fixes the `update.sh` script in the `client-java-contrib/cert-manager` 
module which was broken due to two issues found while attempting to regenerate 
the cert-manager CRD model classes locally.

### 1. Broken Docker Registry
The script was pointing to `docker.pkg.github.com` which is the old GitHub 
Packages registry that requires authentication even for public images.

This caused the following error when running `update.sh`:

    Unable to find image 'docker.pkg.github.com/kubernetes-client/java/crd-model-gen:v2.0.0' locally
    docker: Error response from daemon: Head "https://docker.pkg.github.com/v2/
    kubernetes-client/java/crd-model-gen/manifests/v2.0.0": no basic auth credentials.

GitHub migrated public container images to `ghcr.io` which does not require 
authentication for public images.

### 2. Outdated GitHub Org in CRD URL
The CRD download URL was still using `jetstack/cert-manager` which is the old 
org name. The cert-manager project moved to the `cert-manager/cert-manager` org 
in v1.8. While GitHub currently redirects the old URL, relying on redirects is 
fragile and should not be depended upon.


adithyanms-debug:
I am new to open source. Forgive me if i am wrong

ViliusS:
This PR is an AI slop.

brendandburns:
Can you revert all of the unrelated changes to `mvnw`, then I think this is good to merge.

adithyanms-debug:
@brendandburns  do i need to revert the mnvw in the root directory. I think i didn't make any change over there, and sorry if the code is a slope. Will make it better next time : )

brendandburns:
If you look here:
https://github.com/kubernetes-client/java/pull/4627/changes

There are lots of mvnw related changes, I suspect it is just a difference between your environment and ours.

Thanks!

adithyanms-debug:
Reverted the mvnw changes

brendandburns:
There's also a bunch of file mode changes `100755 -> 100655` which means that the scripts aren't executable, can you revert those changes also?

Thanks

adithyanms-debug:
Made the changes. Please let me know if there are any error

Thanks!!

brendandburns:
image

I still see the file mode changes.

adithyanms-debug:
Oh the file mode need to be 100755 -> 100655. Can you guide me on how to do it. I tried it but no result

Thanks

brendandburns:
I think that `chmod ...` + `git commit` should fix things.

adithyanms-debug:
I made the changes using chmod and git commit. Thanks for your guidance and sorry for responding late

adithyanms-debug:
@brendandburns can you check the push I made :)

brendandburns:
I still see:
image

PR #4741, opened 2026-05-06
fix: refresh EKS authentication token per request

### LINKED ISSUE #4731: EKSAuthentication: no token refresh — single ApiClient becomes unusable after ~15 minutes
**Library Version**: 24.0.0
**Java Version**: 21

**Description**

EKSAuthentication.provide(ApiClient) bakes the bearer token via
client.setApiKey() at ApiClient build time. The token is a SigV4-signed
STS GetCallerIdentity URL whose X-Amz-Expires is capped at 900 seconds
by aws-iam-authenticator, so any single ApiClient instance becomes
unusable ~15 minutes after construction.

In a deployment where a single ApiClient is intended to live for the
process lifetime, we currently use a custom Authentication implementation
— SigV4 presign + an OkHttp Interceptor that injects a fresh
Authorization header per request, backed by a token cache that refreshes
shortly before expiry.

The library already has RefreshAuthentication, which implements the
OkHttp Interceptor pattern for periodic token refresh. The same approach
would apply naturally to EKSAuthentication.

**Question**

Is the current static-bearer behavior intentional (e.g., assuming
short-lived kubectl-style usage)? Or should EKSAuthentication adopt the
interceptor pattern internally?

If the latter, I'd be happy to send a PR.

hwayoungjun:
Fixes #4731

## Description

`EKSAuthentication` currently generates an EKS bearer token when `provide()` is called and stores it on the `ApiClient` as a static API key.

However, EKS authentication tokens are short-lived. Once the generated token expires, subsequent requests continue to use the same expired token and fail with an authentication error.

This change updates `EKSAuthentication` to install an OkHttp interceptor instead. The interceptor adds the bearer token to each request, reuses the token before expiry, and generates a new token after expiry.

## Changes

- Add an OkHttp interceptor to `EKSAuthentication`
- Cache the generated EKS token until its configured expiry
- Refresh the token after expiry and use the refreshed token on subsequent requests
- Keep the maximum token expiry capped at 900 seconds
- Add tests for token injection, reuse before expiry, refresh after expiry, and expiry capping

## Testing

- `./mvnw -pl util -am -Dtest=EKSAuthenticationTest -Dsurefire.failIfNoSpecifiedTests=false test`
- `./mvnw -pl util spotless:check`
- `git diff --check`

I also ran a manual smoke test against an EKS cluster with a 60 second token expiry:

- request at +0s succeeded
- request at +35s succeeded
- request at +70s succeeded

### DIFF
--- a/util/src/main/java/io/kubernetes/client/util/credentials/EKSAuthentication.java
+++ b/util/src/main/java/io/kubernetes/client/util/credentials/EKSAuthentication.java
@@ -12,7 +12,6 @@ limitations under the License.
 */
 package io.kubernetes.client.util.credentials;
 
-import io.kubernetes.client.openapi.ApiClient;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
@@ -29,12 +28,13 @@ import java.time.Duration;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
 import java.util.Base64;
+import java.util.function.Supplier;
 
 /**
  * EKS cluster authentication which generates a bearer token from AWS AK/SK. It doesn't require an "aws"
  * command line tool in the $PATH.
  */
-public class EKSAuthentication implements Authentication {
+public class EKSAuthentication extends RefreshAuthentication {
 
     private static final Logger log = LoggerFactory.getLogger(EKSAuthentication.class);
 
@@ -50,33 +50,45 @@ public class EKSAuthentication implements Authentication {
     }
 
     public EKSAuthentication(AwsCredentialsProvider provider, String region, String clusterName, int expirySeconds) {
-        this.provider = provider;
-        this.region = region;
-        this.clusterName = clusterName;
-        if (expirySeconds > MAX_EXPIRY_SECONDS) {
-            expirySeconds = MAX_EXPIRY_SECONDS;
-        }
-        this.expirySeconds = expirySeconds;
-        this.stsEndpoint = URI.create("https://sts." + this.region + ".amazonaws.com");
+        this(new EksTokenSupplier(provider, region, clusterName, Math.min(expirySeconds, MAX_EXPIRY_SECONDS)));
+    }
+
+    private EKSAuthentication(EksTokenSupplier tokenSupplier) {
+        super(tokenSupplier, Duration.of(tokenSupplier.expirySeconds, ChronoUnit.SECONDS));
+        setExpiry(Instant.now().plus(tokenSupplier.expirySeconds, ChronoUnit.SECONDS));
     }
 
     private static final int MAX_EXPIRY_SECONDS = 60 * 15;
-    private final AwsCredentialsProvider provider;
-    private final String region;
-    private final String clusterName;
-    private final URI stsEndpoint;
 
-    private final int expirySeconds;
+    private static class EksTokenSupplier implements Supplier<String> {
+        private final AwsCredentialsProvider provider;
+        private final String region;
+        private final String clusterName;
+        private final URI stsEndpoint;
+        private final int expirySeconds;
 
-    @Override
-    public void provide(ApiClient client) {
-        SdkHttpRequest httpRequest = generateStsRequest();
-        String presignedUrl = requestToPresignedUrl(httpRequest);
+        EksTokenSupplier(AwsCredentialsProvider provider, String region, String clusterName, int expirySeconds) {
+            this.provider = provider;
+            this.region = region;
+            this.clusterName = clusterName;
+            this.expirySeconds = expirySeconds;
+            this.stsEndpoint = URI.create("https://sts." + this.region + ".amazonaws.com");
+        }
+
+        @Override
+        public String get() {
+            return generateToken(provider, region, clusterName, stsEndpoint, expirySeconds);
+        }
+    }
+
+    private static String generateToken(
+            AwsCredentialsProvider provider, String region, String clusterName, URI stsEndpoint, int expirySeconds) {
+        SdkHttpRequest httpRequest = generateStsRequest(stsEndpoint, clusterName);
+        String presignedUrl = requestToPresignedUrl(httpRequest, provider, region);
         String encodedUrl = presignedUrlToEncodedUrl(presignedUrl);
         String token = "k8s-aws-v1." + encodedUrl;
-        client.setApiKeyPrefix("Bearer");
-        client.setApiKey(token);
         log.info("Generated BEARER token for ApiClient, expiring at {}", Instant.now().plus(expirySeconds, ChronoUnit.SECONDS));
+        return token;
     }
 
     private static String presignedUrlToEncodedUrl(String presignedUrl) {
@@ -85,7 +97,7 @@ public class EKSAuthentication implements Authentication {
                 .encodeToString(SdkHttpUtils.urlEncodeIgnoreSlashes(presignedUrl).getBytes(StandardCharsets.UTF_8));
     }
 
-    private SdkHttpRequest generateStsRequest() {
+    private static SdkHttpRequest generateStsRequest(URI stsEndpoint, String clusterName) {
         return SdkHttpRequest.builder()
                 .uri(stsEndpoint)
                 .putRawQueryParameter("Version", "2011-06-15")
@@ -95,10 +107,11 @@ public class EKSAuthentication implements Authentication {
                 .build();
     }
 
-    private String requestToPresignedUrl(SdkHttpRequest httpRequest) {
+    private static String requestToPresignedUrl(
+            SdkHttpRequest httpRequest, AwsCredentialsProvider provider, String region) {
         AwsV4HttpSigner signer = AwsV4HttpSigner.create();
         SignedRequest signedRequest =
-                signer.sign(r -> r.identity(this.provider.resolveCredentials())
+                signer.sign(r -> r.identity(provider.resolveCredentials())
                         .request(httpRequest)
                         .putProperty(AwsV4HttpSigner.SERVICE_SIGNING_NAME, "sts")
                         .putProperty(AwsV4HttpSigner.REGION_NAME, region)
--- a/util/src/test/java/io/kubernetes/client/util/credentials/EKSAuthenticationTest.java
+++ b/util/src/test/java/io/kubernetes/client/util/credentials/EKSAuthenticationTest.java
@@ -13,6 +13,7 @@ limitations under the License.
 package io.kubernetes.client.util.credentials;
 
 import io.kubernetes.client.openapi.ApiClient;
+import okhttp3.OkHttpClient;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.mockito.Mock;
@@ -20,7 +21,7 @@ import org.mockito.junit.jupiter.MockitoExtension;
 import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
 import software.amazon.awssdk.auth.credentials.AwsSessionCredentials;
 
-import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
@@ -33,6 +34,9 @@ class EKSAuthenticationTest {
     @Mock
     private ApiClient apiClient;
 
+    @Mock
+    private OkHttpClient httpClient;
+
     private String region = "us-west-2";
 
     private String clusterName = "test-2";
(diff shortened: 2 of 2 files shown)


brendandburns:
Rather than re-implementing much of the refresh functionality, please inherit from the RefreshAuthentication similar to what the TokenFileAuthentication class does.

hwayoungjun:
@brendandburns Thanks for the review.

I updated `EKSAuthentication` to extend `RefreshAuthentication`, similar to `TokenFileAuthentication`, and removed the duplicated interceptor/token refresh logic from `EKSAuthentication`.

util/src/main/java/io/kubernetes/client/util/credentials/EKSAuthentication.java
```diff
@@ -50,60 +49,64 @@ public EKSAuthentication(AwsCredentialsProvider provider, String region, String
...
+    private static String generateToken(
+            AwsCredentialsProvider provider, String region, String clusterName, int expirySeconds, Clock clock) {
+        SdkHttpRequest httpRequest = generateStsRequest(region, clusterName);
+        String presignedUrl = requestToPresignedUrl(httpRequest, provider, region, expirySeconds);
         String encodedUrl = presignedUrlToEncodedUrl(presignedUrl);
-        String token = "k8s-aws-v1." + encodedUrl;
-        client.setApiKeyPrefix("Bearer");
-        client.setApiKey(token);
-        log.info("Generated BEARER token for ApiClient, expiring at {}", Instant.now().plus(expirySeconds, ChronoUnit.SECONDS));
+        log.info(
```
brendandburns: I feel like this should maybe be debug level, not info?

brendandburns: Ah, but I see that it was info previously, so you don't have to change this.

util/src/main/java/io/kubernetes/client/util/credentials/EKSAuthentication.java
```diff
@@ -50,60 +49,64 @@ public EKSAuthentication(AwsCredentialsProvider provider, String region, String
...
+                "Generated BEARER token for ApiClient, expiring at {}",
+                Instant.now(clock).plus(expirySeconds, ChronoUnit.SECONDS));
+        return "k8s-aws-v1." + encodedUrl;
     }
 
     private static String presignedUrlToEncodedUrl(String presignedUrl) {
         return Base64.getUrlEncoder()
                 .withoutPadding()
-                .encodeToString(SdkHttpUtils.urlEncodeIgnoreSlashes(presignedUrl).getBytes(StandardCharsets.UTF_8));
```
brendandburns, line 85: Why are we removing the urlEncodeIgnoreSlashes call?

util/src/main/java/io/kubernetes/client/util/credentials/EKSAuthentication.java
```diff
@@ -50,60 +49,64 @@ public EKSAuthentication(AwsCredentialsProvider provider, String region, String
...
 
     public EKSAuthentication(AwsCredentialsProvider provider, String region, String clusterName, int expirySeconds) {
-        this.provider = provider;
-        this.region = region;
-        this.clusterName = clusterName;
-        if (expirySeconds > MAX_EXPIRY_SECONDS) {
-            expirySeconds = MAX_EXPIRY_SECONDS;
-        }
-        this.expirySeconds = expirySeconds;
-        this.stsEndpoint = URI.create("https://sts." + this.region + ".amazonaws.com");
```
brendandburns, line 60: Let's keep the STS endpoint as a member variable vs. constructing it every time.

util/src/main/java/io/kubernetes/client/util/credentials/EKSAuthentication.java
```diff
@@ -50,60 +49,64 @@ public EKSAuthentication(AwsCredentialsProvider provider, String region, String
...
 
     private static String presignedUrlToEncodedUrl(String presignedUrl) {
         return Base64.getUrlEncoder()
                 .withoutPadding()
-                .encodeToString(SdkHttpUtils.urlEncodeIgnoreSlashes(presignedUrl).getBytes(StandardCharsets.UTF_8));
+                .encodeToString(presignedUrl.getBytes(StandardCharsets.UTF_8));
     }
 
-    private SdkHttpRequest generateStsRequest() {
+    private static SdkHttpRequest generateStsRequest(String region, String clusterName) {
```
brendandburns: Does this need to be static? If not, let's keep it as is.

util/src/main/java/io/kubernetes/client/util/credentials/EKSAuthentication.java
```diff
@@ -50,60 +49,64 @@ public EKSAuthentication(AwsCredentialsProvider provider, String region, String
...
+                .uri(URI.create("https://sts." + region + ".amazonaws.com/"))
                 .putRawQueryParameter("Version", "2011-06-15")
                 .putRawQueryParameter("Action", "GetCallerIdentity")
                 .method(SdkHttpMethod.GET)
                 .putHeader("x-k8s-aws-id", clusterName)
                 .build();
     }
 
-    private String requestToPresignedUrl(SdkHttpRequest httpRequest) {
+    private static String requestToPresignedUrl(
```
brendandburns, line 110: As above is there a reason why this needs to be static?

util/src/main/java/io/kubernetes/client/util/credentials/EKSAuthentication.java
hwayoungjun: @brendandburns Thanks for catching this.

I checked the AWS SDK path locally. The signer adds the `X-Amz-*` values as raw query parameters, and `SdkHttpRequest.getUri()` builds the final URI through `encodedQueryParameters()`, so the returned presigned URL already contains percent-encoded query values such as `X-Amz-Credential=...%2F...`.

That was why I initially removed the extra `urlEncodeIgnoreSlashes` step. It looked like another encoding pass on top of an already encoded presigned URL.

However, I should have called this out explicitly in the PR, and this is outside the refresh-related scope of the change.

Since the existing implementation included this extra encoding step, I agree it is safer to preserve the previous token payload behavior. I also should have considered compatibility with other AWS SDK versions and the existing behavior more carefully.

I will restore `urlEncodeIgnoreSlashes`.

util/src/main/java/io/kubernetes/client/util/credentials/EKSAuthentication.java
hwayoungjun: @brendandburns The static methods were mainly introduced because the token supplier has to be passed into the `RefreshAuthentication` constructor, and the existing instance methods could not be passed directly there.

The way I can avoid those static methods is to introduce a small supplier implementation that owns the EKS token generation state. It would roughly look like this:

```java
public class EKSAuthentication extends RefreshAuthentication {
  EKSAuthentication(
      AwsCredentialsProvider provider,
      String region,
      String clusterName,
      int expirySeconds,
      Clock clock) {
    super(
        new EksTokenSupplier(provider, region, clusterName, cappedExpirySeconds(expirySeconds), clock),
        Duration.of(cappedExpirySeconds(expirySeconds), ChronoUnit.SECONDS),
        clock);
  }

  private static class EksTokenSupplier implements Supplier<String> {
    private final AwsCredentialsProvider provider;
    private final String region;
    private final String clusterName;
    private final URI stsEndpoint;
    private final int expirySeconds;
    private final Clock clock;

    @Override
    public String get() {
      return generateToken();
    }

    private String generateToken() {
      // EKS token generation logic.
    }
  }
}

File: util/src/main/java/io/kubernetes/client/util/credentials/EKSAuthentication.java (hunk 5 above)
Comment (hwayoungjun) [reply util/src/main/java/io/kubernetes/client/util/credentials/EKSAuthentication.java:110]: @brendandburns Same reason as above. This was made static as part of passing the token supplier into `RefreshAuthentication`.

With the small supplier implementation, this method can be kept as an instance method there as well.

File: util/src/main/java/io/kubernetes/client/util/credentials/EKSAuthentication.java (hunk 3 above)
Comment (hwayoungjun) [reply util/src/main/java/io/kubernetes/client/util/credentials/EKSAuthentication.java:60]: @brendandburns This was also a side effect of using static methods instead of a separate supplier class. With the small supplier implementation outlined below, `stsEndpoint` could be kept as a field and would not need to be reconstructed each time. I can update it that way if that direction sounds right.

File: util/src/main/java/io/kubernetes/client/util/credentials/EKSAuthentication.java (hunk 4 above)
Comment (brendandburns) [reply util/src/main/java/io/kubernetes/client/util/credentials/EKSAuthentication.java]: Let's do this or alternately you can use an anonymous class (`new Supplier<String> { ... }`)

File: util/src/main/java/io/kubernetes/client/util/credentials/EKSAuthentication.java (hunk 2 above)
Comment (hwayoungjun) [reply util/src/main/java/io/kubernetes/client/util/credentials/EKSAuthentication.java:85]: @brendandburns 

I looked into this more deeply with an actual EKS API request, and I think the issue is a bit broader than the refresh change itself.

## What I found

It looks like the existing `EKSAuthentication` path may not have been producing a token payload that EKS accepts.

In my own usage, I had been using a custom `Authentication` implementation instead of `EKSAuthentication`, so I did not catch this earlier. While verifying this PR against a real EKS cluster, I found that the current `EKSAuthentication` token payload is rejected with `401 Unauthorized`.

## Token payload difference

The expected decoded token payload is the presigned STS URL itself:

```text
https://sts.ap-northeast-2.amazonaws.com/?Version=2011-06-15&Action=GetCallerIdentity&X-Amz-Credential=...%2F...
```

However, with the extra `SdkHttpUtils.urlEncodeIgnoreSlashes(presignedUrl)` step, the decoded token payload becomes a URL-encoded string:

```text
https%3A//sts.ap-northeast-2.amazonaws.com%3FVersion%3D2011-06-15%26Action%3DGetCallerIdentity%26X-Amz-Credential%3D...%252F...
```

That means the URL structure itself is encoded, and already encoded query values such as `%2F` become double-encoded as `%252F`.

## STS path

I also changed the STS request construction to keep the endpoint URI pathless and set the request path explicitly:

```java
.uri(stsEndpoint)
.encodedPath("/")
```

This produces the working presigned URL form:

```text
https://sts...amazonaws.com/?...
```

## Reproduction branch

I created a separate branch to make this easier to verify outside this PR:

https://github.com/hwayoungjun/java/tree/repro/eks-authentication-401

That branch compares:

- the current `EKSAuthentication`
- `FixedEKSAuthentication`, copied from this PR's implementation

The smoke test result against a real EKS cluster was:

```text
Current master decoded payload prefix:
https%3A//sts.ap-northeast-2.amazonaws.com%3F...
Kubernetes API request failed with HTTP 401

Fixed decoded payload prefix:
https://sts.ap-northeast-2.amazonaws.com/?...
Kubernetes API request succeeded, namespaces=13
```

Run command:

```text
./mvnw -pl util -Dtest=EKSAuthenticationSmokeTest -Dsurefire.failIfNoSpecifiedTests=false test
```

## Conclusion

From the real EKS verification, it looks like the existing `EKSAuthentication` token payload is not accepted by EKS today.

Because of that, I committed the token payload fix here by removing the extra whole-URL encoding and making the STS root path explicit with `encodedPath("/")`.

That said, I realize this is separate from the refresh behavior itself. Would you prefer that I split the token payload fix into a separate issue/PR, and keep this PR focused only on refresh support?

brendandburns: Yeah, let's separate this out as a different PR. I don't have access to EKS so we'll need to figure out if there is someone else who can test.

util/src/main/java/io/kubernetes/client/util/credentials/EKSAuthentication.java
```diff
@@ -50,61 +49,82 @@ public EKSAuthentication(AwsCredentialsProvider provider, String region, String
...
+                    .uri(stsEndpoint)
+                    .encodedPath("/")
+                    .putRawQueryParameter("Version", "2011-06-15")
+                    .putRawQueryParameter("Action", "GetCallerIdentity")
+                    .method(SdkHttpMethod.GET)
+                    .putHeader("x-k8s-aws-id", clusterName)
+                    .build();
+        }
+
+        private String requestToPresignedUrl(SdkHttpRequest httpRequest) {
```
brendandburns: Please move this back up to the original line number to minimize the diff in the PR.

util/src/main/java/io/kubernetes/client/util/credentials/EKSAuthentication.java
```diff
@@ -50,61 +49,82 @@ public EKSAuthentication(AwsCredentialsProvider provider, String region, String
...
+            return "k8s-aws-v1." + encodedUrl;
+        }
+
+        private static String presignedUrlToEncodedUrl(String presignedUrl) {
+            return Base64.getUrlEncoder()
+                    .withoutPadding()
+                    .encodeToString(presignedUrl.getBytes(StandardCharsets.UTF_8));
+        }
+
+        private SdkHttpRequest generateStsRequest() {
```
brendandburns: Move this back to the original location to minimize diffs.

brendandburns (changes requested):
Some changes to minimize diffs for readability and a request to separate the token fix from this refresh PR.

Thanks!

util/src/main/java/io/kubernetes/client/util/credentials/EKSAuthentication.java
hwayoungjun: @brendandburns 

I have access to a real EKS cluster and used it throughout to validate the changes in #4779. 
So if there's anything I can help with on the verification side, just let me know.

hwayoungjun:
> Some changes to minimize diffs for readability and a request to separate the token fix from this refresh PR.
> 
> Thanks!

@brendandburns Thanks for the review. I split the EKS token payload fix into a separate PR:

- #4779

That PR contains only the token payload change (`urlEncodeIgnoreSlashes` removal and explicit STS `/` path).

I also reworked this PR in b37814b2e633cc05af0ba567f3a7c1d7696a7386 so it stays focused on `RefreshAuthentication` reuse and token refresh/caching behavior.

brendandburns (approved):

brendandburns:
close/open hack/workaround to trigger CI/CD

PR #4746, opened 2026-05-07
feat: add opt-in toggle for managedFields inclusion in deserialization

### LINKED ISSUE #3995: managedFields not populated since 22.0.0
**Describe the bug**
We are using some managed fields to provide information on `Deployment` objects on a custom dashboard. For this, we are getting a `V1DeploymentList`, like so:

```java
    public V1DeploymentList getKubernetesDeploymentList(String context) throws ApiException {


        KubeConfig kubeConfig = this.kubernetesConfiguration.getKubeConfig();

        AppsV1Api appsApi = new AppsV1Api(this.getApiClient(context));

        return appsApi.listNamespacedDeployment(kubeConfig.getNamespace())
                .labelSelector("app.kubernetes.io/name=our-base-chart")
                .execute();
    }
```

From this, we iterate over the `V1Deployment`s, converting them to our representation (class `Deployment`):

```java
    public List<Deployment> getDeployments(String cluster) {

        try {
            V1DeploymentList deploymentList = this.getKubernetesDeploymentList(cluster);
            List<Deployment> deployments = new ArrayList<>();
            if (deploymentList != null) {
                for (V1Deployment item : deploymentList.getItems()) {
                    deployments.add(DeploymentMapper.mapToDeployment(item));
                }
            }

            return deployments;
        } catch (ApiException e) {
            log.error("Couldn't retrieve deployments.", e);
        }

        return Collections.emptyList();
    }
```

To populate our object, we use:
```java
    public static Deployment mapToDeployment(V1Deployment v1deployment) {

        Deployment deployment = new Deployment();

        V1DeploymentSpec spec = Objects.requireNonNull(v1deployment.getSpec());
        V1ObjectMeta metadata = Objects.requireNonNull(v1deployment.getMetadata());
        List<V1ManagedFieldsEntry> managedFields = Objects.requireNonNull(metadata.getManagedFields());
        log.debug("Managed Fields count: {}", managedFields.size());
        String image = Objects.requireNonNull(spec.getTemplate().getSpec()).getContainers().get(0).getImage();

        deployment.setName(metadata.getName());
        deployment.setStatus(getDeploymentStatus(v1deployment));
        deployment.setVersion(getVersionFromImage(image));
        List<V1ManagedFieldsEntry> fleetFields = managedFields.stream()
                .filter(field -> Objects.equals(field.getManager(), "fleetagent"))
                .toList();
        if (!fleetFields.isEmpty()) {
            LocalDateTime lastUpdated = Objects.requireNonNull(fleetFields.getFirst().getTime()).atZoneSameInstant(ZoneId.systemDefault()).toLocalDateTime();
            deployment.setLastUpdatedAt(lastUpdated.atZone(ZoneId.systemDefault()).toLocalDateTime());
        }
        deployment.setReplicas(spec.getReplicas());

        return deployment;
    }
```

Now, with client version `21.0.2` this works against Kubernetes 1.28.15 and 1.30.10. But with client version 22.0.0 onwards, this does not work anymore (against both Kubernetes versions), as the managed fields are empty, according to the log:

`Managed Fields count: 3`
vs
`Managed Fields count: 0`

As I couldn't find a changelog indicating this change in behavior, I'd assume this is unintentional. If it isn't, I'd find it very helpful if you could point me to the right location in the docs.

So far I have verified that the HTTP response from the API still contains the fields, and those are still present, when the JSON parsing occurs in https://github.com/kubernetes-client/java/blob/release-23/kubernetes/src/main/java/io/kubernetes/client/openapi/JSON.java#L791

**Client Version**
`22.0.0`, but still the case in `23.0.0`. It worked in `21.0.2`.

**Kubernetes Version**
`1.30.10`, but also `v1.28.15`

**Java Version**
Temurin Java 21

**To Reproduce**
1. Get V1DeploymentList for specific namespace
2. Try to get values from getMetadata().getManagedFields()

**Expected behavior**
I expect the managed fields to be populated just like in earlier releases.

**Server:**
 - OS: Ubuntu Linux
 - Platform: Rancher

YuqiGuo105:
## Summary

Since 22.0.0, `V1MetadataExclusionStrategy` has silently excluded `managedFields` from Gson (de)serialization by default to reduce noise. This broke users who rely on `managedFields` data and there was no supported way to re-enable them.

Closes #3995

---

## Workflow: Before this fix

`managedFields` were silently dropped on every API response with no recourse:

```java
// managedFields always empty, even when the server returns them
V1DeploymentList list = appsApi.listNamespacedDeployment(ns).execute();
List<V1ManagedFieldsEntry> fields =
    list.getItems().get(0).getMetadata().getManagedFields(); // []
```

The only workaround was reflection-based hacking into Gson's private `factories` field.

## Workflow: After this fix

Call `JSON.setIncludeManagedFields(true)` **once at startup** to opt in. Everything else is unchanged:

```java
// At application startup (once):
JSON.setIncludeManagedFields(true);

// Then use the API as normal — managedFields will be populated:
V1DeploymentList list = appsApi.listNamespacedDeployment(ns).execute();
List<V1ManagedFieldsEntry> fields =
    list.getItems().get(0).getMetadata().getManagedFields(); // [fleetagent, kubectl-rollout, ...]
```

To restore the default (exclude `managedFields`):

```java
JSON.setIncludeManagedFields(false);
```

The **default remains `false`** (managedFields excluded) so all existing users are unaffected.

---

## Changes

| File | Change |
|------|--------|
| `V1MetadataExclusionStrategy.java` | Add `boolean includeManagedFields` constructor param; when `true`, `shouldSkipField()` returns `false` immediately (no exclusion) |
| `JSON.java` | Add static `includeManagedFields` flag (default `false`); extract 700-line `static {}` into `initGson()` helper; add `public static setIncludeManagedFields(boolean)` that flips the flag and reinitialises the shared Gson instance |
| `JSONTest.java` | Three new tests: default exclusion, opt-in inclusion, and restoration of default after toggling |
| `scripts/patches/json.diff` | Updated patch so the changes survive future OpenAPI code regeneration |

### DIFF
--- a/kubernetes/src/main/java/io/kubernetes/client/openapi/JSON.java
+++ b/kubernetes/src/main/java/io/kubernetes/client/openapi/JSON.java
@@ -57,6 +57,8 @@ import java.util.HashMap;
 public class JSON {
     private static Gson gson;
     private static boolean isLenientOnJson = false;
+    private static final boolean includeManagedFields =
+            Boolean.getBoolean("kubernetes.client.includeManagedFields");
 
     private static final DateTimeFormatter RFC3339MICRO_FORMATTER =
         new DateTimeFormatterBuilder()
@@ -82,7 +84,10 @@ public class JSON {
                 fireBuilder
                         .registerPreProcessor(V1Status.class, new V1StatusPreProcessor())
                         .createGsonBuilder();
-        return builder.setExclusionStrategies(new V1MetadataExclusionStrategy());
+        if (!includeManagedFields) {
+            builder.setExclusionStrategies(new V1MetadataExclusionStrategy());
+        }
+        return builder;
     }
 
     private static String getDiscriminatorValue(JsonElement readElement, String discriminatorField) {
--- a/kubernetes/src/test/java/io/kubernetes/client/openapi/JSONTest.java
+++ b/kubernetes/src/test/java/io/kubernetes/client/openapi/JSONTest.java
@@ -108,4 +108,15 @@ class JSONTest {
     JsonReader jsonReader = new JsonReader(new StringReader("{\"foo\":\"bar\"}"));
     new V1ListMeta.CustomTypeAdapterFactory().create(gson, TypeToken.get(V1ListMeta.class)).read(jsonReader);
   }
+
+  @Test
+  void managedFieldsExcludedByDefault() {
+    String jsonWithManagedFields =
+        "{\"managedFields\":[{\"manager\":\"kubectl\",\"operation\":\"Apply\"}],\"name\":\"test\"}";
+    V1ObjectMeta meta = JSON.getGson().fromJson(jsonWithManagedFields, V1ObjectMeta.class);
+    // managedFields is excluded from deserialization; the field retains its default empty value
+    assertThat(meta.getManagedFields()).satisfiesAnyOf(
+        list -> assertThat(list).isNull(),
+        list -> assertThat(list).isEmpty());
+  }
 }
--- a/scripts/patches/json.diff
+++ b/scripts/patches/json.diff
@@ -1,11 +1,11 @@
-From eeb5069c974d3146495f8a9dc52653f029d3cc2d Mon Sep 17 00:00:00 2001
+﻿From eeb5069c974d3146495f8a9dc52653f029d3cc2d Mon Sep 17 00:00:00 2001
 From: Min Jin <minkimzz@amazon.com>
 Date: Tue, 4 Feb 2025 11:59:25 -0800
 Subject: [PATCH] manually apply JSON patch
 
 ---
- .../io/kubernetes/client/openapi/JSON.java    | 36 ++++++++++++++++---
- 1 file changed, 32 insertions(+), 4 deletions(-)
+ .../io/kubernetes/client/openapi/JSON.java    | 39 +++++++++++++++++---
+ 1 file changed, 35 insertions(+), 4 deletions(-)
 
 diff --git a/kubernetes/src/main/java/io/kubernetes/client/openapi/JSON.java b/kubernetes/src/main/java/io/kubernetes/client/openapi/JSON.java
 index dda3ec708..fe902b293 100644
@@ -31,10 +31,12 @@ index dda3ec708..fe902b293 100644
  import java.util.Date;
  import java.util.Locale;
  import java.util.Map;
-@@ -50,9 +56,20 @@ import java.util.TimeZone;
+@@ -50,9 +56,23 @@ import java.util.TimeZone;
  public class JSON {
      private static Gson gson;
      private static boolean isLenientOnJson = false;
++    private static final boolean includeManagedFields =
++            Boolean.getBoolean("kubernetes.client.includeManagedFields");
 +
 +    private static final DateTimeFormatter RFC3339MICRO_FORMATTER =
 +        new DateTimeFormatterBuilder()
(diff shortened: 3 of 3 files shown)


brendandburns:
This file is auto generated from Kubernetes swagger, I think the fact that the field is missing means that the field must be missing from the swagger. We should fix that in the upstream swagger.

YuqiGuo105:
> This file is auto generated from Kubernetes swagger, I think the fact that the field is missing means that the field must be missing from the swagger. We should fix that in the upstream swagger.

Thanks for reviewing! I want to clarify the root cause — this is not a swagger / code generation issue.

The [managedFields](vscode-file://vscode-app/c:/Users/Yuqi%20Guo/AppData/Local/Programs/Microsoft%20VS%20Code/8b640eef5a/resources/app/out/vs/code/electron-browser/workbench/workbench.html) field is present in both the upstream swagger and the generated [V1ObjectMeta.java](vscode-file://vscode-app/c:/Users/Yuqi%20Guo/AppData/Local/Programs/Microsoft%20VS%20Code/8b640eef5a/resources/app/out/vs/code/electron-browser/workbench/workbench.html) (line 105). The server also returns the field in the HTTP response body (confirmed by the reporter in the linked issue).

The data is being silently dropped during Gson deserialization by [V1MetadataExclusionStrategy](vscode-file://vscode-app/c:/Users/Yuqi%20Guo/AppData/Local/Programs/Microsoft%20VS%20Code/8b640eef5a/resources/app/out/vs/code/electron-browser/workbench/workbench.html) — a hand-written (non-generated) class that tells Gson to skip the [managedFields](vscode-file://vscode-app/c:/Users/Yuqi%20Guo/AppData/Local/Programs/Microsoft%20VS%20Code/8b640eef5a/resources/app/out/vs/code/electron-browser/workbench/workbench.html) field on [V1ObjectMeta](vscode-file://vscode-app/c:/Users/Yuqi%20Guo/AppData/Local/Programs/Microsoft%20VS%20Code/8b640eef5a/resources/app/out/vs/code/electron-browser/workbench/workbench.html). This strategy is injected into the generated [JSON.java](vscode-file://vscode-app/c:/Users/Yuqi%20Guo/AppData/Local/Programs/Microsoft%20VS%20Code/8b640eef5a/resources/app/out/vs/code/electron-browser/workbench/workbench.html) via [json.diff](vscode-file://vscode-app/c:/Users/Yuqi%20Guo/AppData/Local/Programs/Microsoft%20VS%20Code/8b640eef5a/resources/app/out/vs/code/electron-browser/workbench/workbench.html) after each code generation run.

This was an intentional change introduced in 22.0.0 with no public opt-out API, which is what broke the users in the linked issue. This PR adds a simple toggle (JSON.setIncludeManagedFields(true)) while keeping the default behavior unchanged.

kubernetes/src/main/java/io/kubernetes/client/openapi/JSON.java
```diff
@@ -839,6 +840,15 @@ private static Class getClassByDiscriminator(Map classByDiscriminatorValue, Stri
         gson = gsonBuilder.create();
     }
 
+    static {
```
brendandburns: I don't think this part of the change is necessary (the method refactor etc) please remove.

kubernetes/src/main/java/io/kubernetes/client/gson/V1MetadataExclusionStrategy.java
```diff
@@ -17,7 +17,20 @@
 import io.kubernetes.client.openapi.models.V1ObjectMeta;
 
 public class V1MetadataExclusionStrategy implements com.google.gson.ExclusionStrategy {
+    private final boolean includeManagedFields;
```
brendandburns: Rather than modifying this class, let's just not add the exclusion strategy if includeManagedFields is true in the JSON class below.

PR #4751, opened 2026-05-12
cert-manager: build crd-model-gen image locally in update.sh

YuqiGuo105:
## Background

Issue #3489 has two distinct problems mixed together. The original report (V1 cert-manager models missing from the published jar) was already addressed on `master` long ago: `client-java-contrib/cert-manager/src/main/java/io/cert/manager/models/` now contains 109 V1-only types, with no `V1alpha*`/`V1beta*` classes left. So that part doesn't need a code change.

This PR addresses the **other** problem raised in the comments on #3489: contributors who try to follow the README and run `./update.sh` are blocked at the first step because docker pulls fail with `Access is denied`. The default image referenced by `update.sh`,

```
docker.pkg.github.com/kubernetes-client/java/crd-model-gen:v2.0.0
```

lives on a registry that requires GitHub authentication, so anonymous pulls do not work. This was reported by @vaidikcode and @AyushSinha-6409 in the issue thread.

## Changes

- **`client-java-contrib/cert-manager/update.sh`**: detect the missing image at runtime and build it locally from the `Dockerfile` already shipped in `client-java-contrib/`, defaulting to `local/crd-model-gen:latest`. `IMAGE_NAME` and `IMAGE_TAG` can still be overridden to use a prebuilt image.
- **`client-java-contrib/cert-manager/README.md`**: refresh the dependency snippet and compatibility table so they match the current artifact version (`27.0.0-SNAPSHOT`) and the v1.13.4 CRD URL already used by `update.sh`. No regeneration of model classes is included; the existing V1 classes on `master` are unchanged.

## Verification

Running `./update.sh` from a clean checkout (with no prebuilt image present) now builds `local/crd-model-gen:latest` from the in-tree `Dockerfile` and proceeds with the generation pipeline without contacting the private registry.

Refs: kubernetes-client/java#3489

### DIFF
--- a/client-java-contrib/cert-manager/README.md
+++ b/client-java-contrib/cert-manager/README.md
@@ -8,14 +8,14 @@ To use this library, include the following maven dependency
 <dependency>
     <groupId>io.kubernetes</groupId>
     <artifactId>client-java-cert-manager-models</artifactId>
-    <version>0.16.1-SNAPSHOT</version>
+    <version>27.0.0-SNAPSHOT</version>
 </dependency>
 ```
 Please refer to the [CertManagerExample](../../examples/src/main/java/io/kubernetes/client/examples/CertManagerExample.java), which demonstrates how to create a self signed issuer using the model class and Kubernetes Java client generic API.
 ## Compatibility
 Artifact Version|Cert-Manager Release Version|CRD Source
 ----------------|----------------------------|----------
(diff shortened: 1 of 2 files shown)


brendandburns:
I don't think this change includes what you are thinking, it doesn't remove an alpha files.

brendandburns:
Works for me.

/lgtm
/approve

PR #4802, opened 2026-06-08
fix(util): prevent non-tar copy path traversal on release-25

brendandburns:
Harden Copy non-tar directory copy against path traversal from pod-controlled ls -F output by sanitizing entry names and enforcing destination path containment. Add regression test for traversal and positive test for safe entries.


brendandburns:
@copilot please fix this PR by updating all of the github actions to use SHA.

PR #4875, opened 2026-08-06
Update issue templates

sxn-rs5:


brendandburns:
I'm going to close this because it is a text only change and I don't think it adds much value.

PR #4893, opened 2026-09-01
Add ListOptions watch parity and fluent builder

brendandburns:
Splits the generic `ListOptions` changes out of #4892.

## Summary
- add watch-related parity fields to `ListOptions` and `GenericKubernetesApi`
- add a fluent `ListOptions` builder
- return an independent copy from `build()` so later builder mutations do not affect previously built options
- add unit and end-to-end coverage

## Validation


yue9944882 (commented):
/lgtm
/approve

PR #4894, opened 2026-09-01
Add predicate-aware informer event handlers

brendandburns:
Splits predicate-aware informer event-handler changes out of #4892.

## Summary
- add predicate overloads for informer event handlers
- add `FilteringResourceEventHandler`
- handle transitions into and out of a predicate match
- harden filtering against null or malformed objects
- add unit and end-to-end coverage

## Validation

### DIFF
--- a/util/src/main/java/io/kubernetes/client/informer/FilteringResourceEventHandler.java
+++ b/util/src/main/java/io/kubernetes/client/informer/FilteringResourceEventHandler.java
@@ -1,0 +1,80 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


util/src/main/java/io/kubernetes/client/informer/FilteringResourceEventHandler.java
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  private boolean matches(ApiType obj) {
+    if (obj == null) {
+      return false;
+    }
+    try {
+      return filter.test(obj);
+    } catch (RuntimeException e) {
+      return false;
```
brendandburns, line 77: Add a log here so that we don't silently drop events.

yue9944882 (commented):
/lgtm
/approve

PR #4895, opened 2026-09-01
Add exponential backoff for informer watch reconnects

brendandburns:
Splits informer watch reconnect backoff changes out of #4892.

## Summary
- add exponential backoff for transient watch connection failures
- reset backoff at the appropriate run/watch boundaries
- stop retrying promptly when interrupted
- tighten and stabilize backoff tests
- add end-to-end coverage for recovery from initial connection failures

## Validation


yue9944882 (commented):
/lgtm
/approve

PR #4900, opened 2026-09-04
fix: avoid NPE in DeploymentHelper.equalIgnoreHash on templates with no labels

sureshmelvinsigera:
`DeploymentHelper.equalIgnoreHash` NPEs when a pod template has no labels (or no metadata at all), both are valid per the k8s API but weren't null-checked. This breaks `getAllReplicaSets`, which is used for deployment rollout status.

Added null checks before removing the pod-template-hash label, and a fallback for when metadata itself is null. Added a test that repros the old NPE and confirms it's fixed.


brendandburns:
Thank you for the PR. Before further review, please remove all whitespace changes, the easiest way to do this is probably to run `mvn spotless:apply`

Thanks

sureshmelvinsigera:
Fixed - IntelliJ had reformatted some unrelated lines on save. Reset the file and reapplied just the two intended edits. Diff is clean now.

brendandburns:
/lgtm
/approve

Thanks for fixing the whitespace.

PR #4909, opened 2026-09-16
Fix generic type inference in ServerSideApply example

sarveshkaushal:
### Summary
Fix the `ServerSideApply` example by explicitly specifying the resource and list generic types. This prevents Java from inferring their upper-bound types and resolves compilation errors for `V1ConfigMap` class.

### Testing
Ran `mvn install` after the fix and it worked fine on Java 11 and Java 25. 

### Failure
This was the failure when running `mvn install` on the project root:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.16.0:compile (default-compile) on project client-java-examples-release-latest: Compilation failure: Compilation failure:
[ERROR] /Users/sarveshkaushal/git/java/examples/examples-release-latest/src/main/java/io/kubernetes/client/examples/ServerSideApplyExample.java:[69,38] incompatible types: java.lang.Class<io.kubernetes.client.openapi.models.V1ConfigMap> cannot be converted to java.lang.Class<io.kubernetes.client.common.KubernetesObject>
[ERROR] /Users/sarveshkaushal/git/java/examples/examples-release-latest/src/main/java/io/kubernetes/client/examples/ServerSideApplyExample.java:[81,38] incompatible types: java.lang.Class<io.kubernetes.client.openapi.models.V1ConfigMap> cannot be converted to java.lang.Class<io.kubernetes.client.common.KubernetesObject>
```


brendandburns:
I'm not opposed to this PR, but I believe that we build these examples in our CI/CD and CI/CD is passing. Can you take a look and see if we do in fact build the examples, and then also figure out if there is a difference between your environment and our build environment?

sarveshkaushal:
Looks like examples module is not included in the CI/CD build logic. That's why it is not failing in the CI/CD pipeline. See this here. https://github.com/kubernetes-client/java/blob/f7ac17b70a7547ee04002bf2beac02ecd2eba4fe/.github/workflows/bazel.yml#L143-L151

When we build locally with maven, it includes all the projects from the root regardless of the Bazel configurations. 

Another related questions - do we want to include the example module in the CI/CD building process? I can ship a follow-up to allow that.

brendandburns:
Yeah, let's re-include examples, I think I broke that in the switch to bazel.

PR #4910, opened 2026-09-17
fix: remove Thread.sleep() from DefaultWorkQueueTest

sarveshkaushal:
### Notes
This change removes the Thread.sleep() dependency on the `DefaultWorkQueueTest` and replaces it with explicit thread synchronization.. This is related to issue #1223 

### Testing
Ran `mvn install` from project root and build completed successfully.


brendandburns:
I think there are a lot of whitespace changes that make this hard to review because it makes the diff much bigger than it needs to be.

Please minimize the diff and then I can do a real review.

sarveshkaushal:
@brendandburns - Thanks for pointing that out. I ran IntelliJ's code formatter on the file, so the formatting in the changed section is formatter-generated rather than intentional whitespace changes. I checked the Git diff as well, and there don't appear to be whitespace-only changes outside the modified test.

brendandburns:
The git diff makes this pretty hard to review. With changes like this that modify the thread synchronization stuff and modify a unit test, minimizing the diff really makes a big difference in reviewability. Can you take a stab at trying to minimize, for example remove the changes to import ordering.

PR #4911, opened 2026-09-17
Remove Thread.sleep() from port forward test class

sarveshkaushal:
### Summary
Replaces the fixed `Thread.sleep()` in `PortForwardTest` with semaphore-based synchronization. 

The test needs to close the WebSocket handler only after `PortForwardResult.init` has created the input stream which it will block while reading the port identifier. A test-specific `WebSocketStreamHandler` subclass signals a semaphore when stream `0` is created, allowing the test to close the handler at the correct point.

A second semaphore waits for `init()` to finish after the handler is closed.

This is related to issue https://github.com/kubernetes-client/java/issues/1223

### Testing
Ran `mvn install` on the root of the project.


brendandburns:
As with the previous one please minimize diff for easier reviewing.

PR #4915, opened 2026-09-18
fix: removing Thread.sleep() from KubernetesReconcilerCreatorTest

sarveshkaushal:
### Notes

Replaces the fixed `Thread.sleep()` in `KubernetesReconcilerCreatorTest` with semaphore based synchronization.

The custom work-queue key function releases a semaphore after adding the generated request to the controller work queue. The test waits for that signal before asserting on the queue.

This is related to issue https://github.com/kubernetes-client/java/issues/1223

### Testing
Ran `mvn install` on the project root.


brendandburns:
As with the other ones, please try to minimize the diff for easier reviewing.

